PR #9, opened 2022-05-11
410 container rework

vdimiczrc:


include/alp/imf.hpp
```diff
@@ -50,107 +50,202 @@
...
-        /**
-         * A composition of two IMFs.
-         * \f$I_{g,n} =[0, n), I_{g,N} =[0, N)\f$
-         * \f$I_{f,n} =[0, n), I_{f,N} =[0, N)\f$
-         * \f$Composed_{f, g} = I_{g,n} \rightarrow I_{f,N}; i \mapsto f( g( i ) )\f$
-         */
-        
-        class Composed: public IMF {
+				size_t map( const size_t i ) const {
+					std::cout << "Calling Strided map" << std::endl;
```
anyzelman, line 95: Should this be in an `#ifdef _DEBUG` ?

include/alp/imf.hpp
```diff
@@ -50,107 +50,202 @@
...
+				Id( const size_t n ) : Strided( n, n, 0, 1 ) {}
+		};
 
-                Composed(std::shared_ptr<IMF> f, std::shared_ptr<IMF> g): IMF( g->n, f->N ), f(f), g(g) { }
+		class Select: public IMF {
+
+			public:
 
-        };
+				std::vector< size_t > select;
```
anyzelman, line 126: If this IMF is selecting a couple of indices based on a given vector of indices, should it not refer to an ALP vector?

include/alp/imf.hpp
```diff
@@ -50,107 +50,202 @@
...
+
+				size_t map( size_t i ) const {
+					std::cout << "Calling Select map; vec size = " << select.size() << std::endl;
```
anyzelman, line 129: ifdef ?

include/alp/imf.hpp
```diff
@@ -50,107 +50,202 @@
...
+
+				Select(size_t N, std::vector< size_t > &select): IMF( select.size(), N ), select( select ) {
+					//if ( *std::max_element( select.cbegin(), select.cend() ) >= N) {
+					//	throw std::runtime_error("IMF Select beyond range.");
+					//}
+				}
+
+				Select(size_t N, std::vector< size_t > &&select): IMF( select.size(), N ), select( select ) {
+					std::cout << "Select move constructor" << std::endl;
+					std::cout << (this->select).size() << std::endl;
```
anyzelman, line 141: ifdef ?

include/alp/imf.hpp
```diff
@@ -50,107 +50,202 @@
...
+
+		template< typename FirstIMF, typename SecondIMF >
+		class Composed: public IMF {
+
+			public:
+				const FirstIMF &g;
+				const SecondIMF &f;
+
+				size_t map( const size_t i ) const {
+						std::cout << "Calling Composed< IMF, IMF>::map()" << std::endl;
```
anyzelman, line 174: ifdef ?

include/alp/imf.hpp
```diff
@@ -50,107 +50,202 @@
...
+						return f.map( g.map( i ) );
```
anyzelman, line 175: Should we print a warning or something whenever this IMF is used, as it is probably not ideal?

include/alp/imf.hpp
```diff
@@ -50,107 +50,202 @@
...
+		Strided ComposedFactory::create( const Strided &f1, const Id &f2 ) {
+			return Strided( f1.n, f2.N, f1.s * f2.s, f1.s * f2.b + f1.b );
+		}
+
+		/** Composition of two Id IMFs is an Id Imf */
+		template<>
+		Id ComposedFactory::create( const Id &f1, const Id &f2 ) {
+			(void)f2;
+			// The first function's domain must be equal to the second function's domain.
+			assert( f1.N == f2.n );
```
anyzelman, line 241: why is one N capital and the other one is a small n?

include/alp/reference/blas2.hpp
```diff
@@ -100,12 +100,12 @@ namespace alp {
 		class Ring,
 		typename IOType, typename IOStructure, typename IOView,
 		typename InputType1, typename InputStructure1, typename InputView1,
-		typename InputType2, typename InputStructure2,  typename InputView2,
+		typename InputType2, typename InputStructure2,  typename InputView2, typename ImfR2, typename ImfC2,
 		typename InputType3, typename InputStructure3, typename InputView3 >
 	RC vxm( Vector< IOType, IOStructure, Density::Dense, IOView, reference > & u,
 		const Vector< InputType3, InputStructure3, Density::Dense, InputView3, reference > & mask,
 		const Vector< InputType1, InputStructure1, Density::Dense, InputView1, reference > & v,
-		const Matrix< InputType2, InputStructure2, Density::Dense, InputView2, reference > & A,
```
anyzelman, line 108: Should vectors also not have IMFs?

include/alp/structures.hpp
```diff
@@ -168,36 +168,55 @@ namespace alp {
...
+		 * @param imf_l            The IMF applied to the rows of the source matrix.
+		 * @param imf_r            The IMF applied to the columns of the source matrix.
+	
+		 * @return \a false if the function can determined that the new view may alter underlying assumptions
+		 * 			associated with the source structure \a SourceStructure; \a true otherwise.
+		 */
+		template< typename SourceStructure, typename TargetStructure >
+		struct isInstantiable {
+			template< typename ImfL, typename ImfR >
+			static bool check( const ImfL &imf_l, const ImfR &imf_r ) {
```
anyzelman, line 188: Was this intended to be a compile-time check a la `static constexpr bool value = ...` ?

include/alp/base/matrix.hpp
```diff
@@ -442,7 +442,8 @@ namespace alp {
 	} // namespace internal
 
 
-	template< typename T, typename Structure, enum Density density, typename View, enum Backend backend >
+	template< typename T, typename Structure, enum Density density, typename View,
+		typename ImfL, typename ImfR, enum Backend backend >
```
anyzelman, line 446: Is ImfL something that affects rows and ImfR something that affects columns? (Maybe add documentation at least at the initial occurrence of these template arg names:)

vdimiczrc: We adopted a new naming ImfR for rows, ImfC for columns. This accidentally remained, I will fix it.

include/alp/imf.hpp
vdimiczrc: N is the cardinality of the co-domain, n is the cardinality of the domain of a mapping function. For the composition, f2(f1(x)), we require that f1.N == f2.n.
I see that the comment in the code is incorrect wrt the domain/co-domain nomenclature. Will fix

include/alp/reference/blas2.hpp
vdimiczrc: They should, will be added later :)

include/alp/imf.hpp
anyzelman: ah got it! Thanks

anyzelman: the (void)f2; I see now btw could go into an `ifdef NDEBUG`, though that's being very pedantic... I think (but am not sure) that some linting tools will otherwise complain about it (when NDEBUG is not defined)

include/alp/smf.hpp
```diff
@@ -27,27 +27,179 @@
...
+		 * expressed as a BivariateQuadratic polynomial depending
+		 * on the types of the IMFs and the SMF.
+		 */
+		template< typename ImfR, typename ImfC, typename Smf >
+		struct Composition {
+			typedef polynomials::BivariateQuadratic< 1, 1, 1, 1, 1, 1, 1 > type;
+		};
+
+		template<>
+		struct Composition< imf::Strided, imf::Strided, Full_t > {
```
hidanielesgit, line 128: I think this could also be generalized. Could we combine the template coefficients of the source polynomials to determine the final composition?

include/alp/structures.hpp
hidanielesgit: @anyzelman I think this is the runtime counterpart. This is the kind of check we specify in `get_view` will be used only if it involves constant time resolution (e.g., using get_view to select a small (possibly variable-size) non-zero square from a triangular matrix) and only enabled in debug mode for more expensive cases.

include/alp/reference/matrix.hpp
```diff
@@ -357,258 +359,288 @@ namespace alp {
...
-		void getInitialized( MatrixBase &, bool ) noexcept;
+		template< typename DerivedMatrix >
+		void getInitialized( MatrixBase< DerivedMatrix > &, bool ) noexcept;
 	} // namespace internal
 
-	template< typename T, typename Structure, enum Density density, typename View >
-	bool getInitialized( Matrix< T, Structure, density, View, reference > & A ) noexcept {
+	template< typename T, typename Structure, enum Density density, typename View, typename ImfL, typename ImfR >
+	bool getInitialized( Matrix< T, Structure, density, View, ImfL, ImfR, reference > & A ) noexcept {
 		return getInitialized( A );
```
hidanielesgit, line 412: Should this go through something like here below?
```
internal::getInitialized( internal::getContainer(...) );
```

include/alp/reference/matrix.hpp
```diff
@@ -357,258 +359,288 @@ namespace alp {
...
+	void setInitialized( Matrix< T, Structure, density, View, ImfL, ImfR, reference > & A, bool initialized ) noexcept {
 		setInitialized( A, initialized );
 	}
 
-	template< typename D, typename Structure, typename View >
-	size_t nrows( const Matrix< D, Structure, Density::Dense, View, reference > & A ) noexcept {
-		return internal::nrows( A );
+	template< typename D, typename Structure, typename View, typename ImfL, typename ImfR >
+	size_t nrows( const Matrix< D, Structure, Density::Dense, View, ImfL, ImfR, reference > & A ) noexcept {
+		//return internal::nrows( static_cast< internal::MatrixBase< ImfL, ImfR > > ( A ) );
```
hidanielesgit, line 422: Should the commented code be the right way to forward the `nrows`?

include/alp/imf.hpp
```diff
@@ -50,107 +50,202 @@
...
+			typedef Strided type;
+		};
+
+		template<>
+		struct composed_type< Strided, Id > {
+			typedef Strided type;
+		};
+
+		template<>
+		struct composed_type< Id, Id > {
```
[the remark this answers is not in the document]
vdimiczrc: We have Id to avoid writing Strided( 0, N ) instead of Id( N ). Since we do not use dynamic polymorphism, we have to specialize to all variants. Do you suggest completely removing Id?

PR #10, opened 2022-05-13
5 update code style guidelines

vdimiczrc:


djelovina:
@anyzelman It is not clear if function calls with ignored return value should be written
```
(void)f(...)
(void) f(...)
( void )f(...)
( void ) f(...)
```
or something else

anyzelman:
> 

This should be

> @anyzelman It is not clear if function calls with ignored return value should be written
> 
> ```
> (void)f(...)
> (void) f(...)
> ( void )f(...)
> ( void ) f(...)
> ```
> 
> or something else

Sorry, very late reply. IT should be `(void) f( ... )`. As for rationale, it's a short-hand of silencing some static code analyser warnings.

PR #11, opened 2022-05-17
Resolve internal issue 411: create a transition path sparseblas target

anyzelman:
This MR introduces a proof-of-concept where ALP/GraphBLAS is used to generate a (subset) of the standard `blas_sparse.h` and `spblas.h` APIs. This is dubbed a "transition path" since it allows the use of software as-is currently implemented based on any of the preceding two standards, without having to re-write (parts of) the user code to follow the ALP syntax -- even though the latter may be simple to do.

The current functions are tested by external unit tests. Before this MR is merged, issues should be opened to:
- [x] unit tests the internal wrappers that this MR introduces
- [x] smoke tests the public APIs this MR introduces
- [x] expand the coverage of the blas_sparse.h and the spblas.h
- [x] implement the SpMM multiply


anyzelman:
Another code review note:
- this MR also fixes a bug in the reference matrix output iterator, where the last increment did not properly bring the iterator in end position.

PR #16, opened 2022-05-19
Add a getter for a reference to an element in matrix

vdimiczrc:
This exposes a const and non-const getter.
Add a unit test for matrix element access.

### DIFF
--- a/include/alp/reference/matrix.hpp
+++ b/include/alp/reference/matrix.hpp
@@ -420,11 +420,11 @@ namespace alp {
 	namespace internal {
 
 		// Forward declaration
-		template< typename T, typename MatrixType >
-		const T &getReference( const MatrixType &, const size_t, const size_t );
+		template< typename MatrixType >
+		const typename MatrixType::value_type &getReference( const MatrixType &, const size_t, const size_t );
 
-		template< typename T, typename MatrixType >
-		T &getReference( MatrixType &, const size_t, const size_t );
+		template< typename MatrixType >
+		typename MatrixType::value_type &getReference( MatrixType &, const size_t, const size_t );
 
 		/**
 		 * Base Matrix class containing attributes common to all Matrix specialization
@@ -483,8 +483,6 @@ namespace alp {
 					Vector< T, reference >,
 					Vector< T, reference > &
 				>::type container_type;
-
-				friend MatrixBase< self_type >;
 
 				/** A container-type view is characterized by its association with a physical container */
 				container_type container;
--- a/include/alp/reference/matrix.hpp
+++ b/include/alp/reference/matrix.hpp
@@ -421,10 +421,10 @@ namespace alp {
 
 		// Forward declaration
 		template< typename MatrixType >
-		const typename MatrixType::value_type &getReference( const MatrixType &, const size_t, const size_t );
+		const typename MatrixType::value_type &access( const MatrixType &, const size_t, const size_t );
 
 		template< typename MatrixType >
-		typename MatrixType::value_type &getReference( MatrixType &, const size_t, const size_t );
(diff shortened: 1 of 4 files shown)


include/alp/reference/matrix.hpp
```diff
@@ -432,6 +439,23 @@ namespace alp {
...
 					return static_cast< const DerivedMatrix * >( this )->_dims();
 				}
+
+				template< typename MatrixType >
+				friend const typename MatrixType::value_type &getReference( const MatrixType &A, const size_t i, const size_t j );
+
+				template< typename MatrixType >
+				friend typename MatrixType::value_type &getReference( MatrixType &A, const size_t i, const size_t j );
+
+				template< typename T >
```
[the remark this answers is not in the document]
vdimiczrc: At this point DerivedMatrix is incomplete type and its typedefs are not yet defined. We could add a static check in the function body to check if T is same as DerivedMatrix::access_type. Accessing DerivedMatrix::access_type within function body is fine because function body is "evaluated" after the class is fully "processed".

include/alp/reference/matrix.hpp
```diff
@@ -432,6 +439,25 @@ namespace alp {
...
 				}
+
+				template< typename MatrixType >
+				friend const typename MatrixType::access_type access( const MatrixType &A, const size_t i, const size_t j );
+
+				template< typename MatrixType >
+				friend typename MatrixType::access_type access( MatrixType &A, const size_t i, const size_t j );
+
+				template< typename AccessType >
+				const AccessType access( const size_t i, const size_t j ) const {
```
hidanielesgit, line 450: Should we go directly for `DerivedMatrix::access_type` here instead of generalizing with `AccessType`?

include/alp/reference/matrix.hpp
```diff
@@ -432,6 +439,25 @@ namespace alp {
...
+					static_assert( std::is_same< AccessType, typename DerivedMatrix::access_type >::value );
+					return static_cast< const DerivedMatrix & >( *this ).access( i, j );
+				}
+
+				template< typename AccessType >
+				AccessType access( const size_t i, const size_t j ) {
```
hidanielesgit, line 456: Should we go directly for `DerivedMatrix::access_type` here instead of generalizing with `AccessType`?

tests/unit/dense_matrix_access.cpp
```diff
@@ -0,0 +1,124 @@
...
+	// set matrix elements using the internal interface
+	setElements( M, 1 );
+	// create transposed view over M
+	auto Mt = alp::get_view< alp::view::transpose >( M );
+	// create a square view over M and treat it as a matrix view with a square structure
+	const size_t block_size = 4;
+	auto Mview = alp::get_view( M, alp::utils::range( 0, block_size ), alp::utils::range( 0, block_size ) );
+	auto Sq_Mref = alp::get_view< alp::structures::Square > ( Mview );
+
+	// verify that accessing corner elements succeeds
```
hidanielesgit, line 56: Should we also try accessing outside logically acceptable areas?  `assert` could be used to allow disabling checks.

include/alp/reference/matrix.hpp
```diff
@@ -565,7 +641,7 @@ namespace alp {
 				typedef std::function< Ret( Args... ) > lambda_function_type;
 				lambda_function_type &lambda;
 
-				T get( const size_t i, const size_t j ) const {
+				access_type access( const size_t i, const size_t j ) const {
```
hidanielesgit, line 644: I think we should keep a coherent interface between derivations of MatrixBase. Should we have `access` obtain a single "storage index" (a pair in the case of functors) as it happens in the container-type Matrix?

vdimiczrc: Fixed in the latest push

PR #17, opened 2022-05-25
Resolve internal issue 434: amf to provide physical container size

vdimiczrc:


include/alp/smf.hpp
```diff
@@ -170,7 +195,7 @@ namespace alp {
 #endif
 					(void)s;
 					(void)P;
-					return smf.f( imf_r.map( i ), imf_c.map( j ) );
+					return map_poly.evaluate( imf_r.map( i ), imf_c.map( j ) );
```
hidanielesgit, line 198: To confirm, this is only meant for dynamic imfs right?

hidanielesgit (approved):

PR #18, opened 2022-05-25
Resolves internal issue #440

anyzelman:
Bugfix: Makes the new `wrapRawVector` work even if the Coordinates class of the output vector does not match the default one of the reference implementation. This is needed for the backends that degelate some functionalities to `reference` or `reference_omp`.


anyzelman:
All targets (including tests) build, all smoke tests pass, internal CI is happy, and code review OK.

PR #22, opened 2022-06-03
448 internal set getinitialized to be invoked via matrixbase

vdimiczrc:


include/alp/reference/matrix.hpp
```diff
@@ -659,11 +675,13 @@ namespace alp {
 			public:
 
 				MatrixFunctor(
+					const bool initialized,
```
hidanielesgit, line 678: Perhaps annotate that this is a temporary solution. A specific issue will be open about linking the status to input containers.

hidanielesgit (approved):

PR #24, opened 2022-06-13
458 container level traits

vdimiczrc:


include/alp/matrix.hpp
```diff
@@ -52,6 +52,17 @@ namespace alp {
 		enum Backend backend = config::default_backend >
 	class Matrix;
 
+	/** Specializations of ALP backend-agnostic type traits */
+	template< typename T, typename Structure, enum Density density, typename View, typename ImfR, typename ImfC, enum Backend backend >
+	struct inspect_structure< Matrix< T, Structure, density, View, ImfR, ImfC, backend > > {
+		typedef Structure type;
+	};
+
+	template< typename T, typename Structure, enum Density density, typename View, typename ImfR, typename ImfC, enum Backend backend >
```
hidanielesgit, line 61: I think traits should also reflect the visibility of what they inspect. For instance `inspect_view` should  probably be internal.

include/alp/reference/matrix.hpp
```diff
@@ -349,12 +349,15 @@ namespace alp {
...
 	/**
-	 * @brief A reference Matrix is an ALP object.
+	 * Identifies a reference internal matrix is an ALP container.
+	 * \internal \todo Should this relation remain?
 	 */
 	template< typename T >
-	struct is_container< internal::Matrix< T, reference > > {
-		static const constexpr bool value = true;
-	};
+	struct is_container< internal::Matrix< T, reference > > : std::true_type {};
```
hidanielesgit, line 356: Should we have two levels of visibility here? internal::is_container usable only internally to differentiate between internal and user-level containers (i.e., Matrix, Vector, Scalar).

include/alp/type_traits.hpp
```diff
@@ -148,6 +185,71 @@ namespace alp {
...
+	struct inspect_structure {};
+
+	/**
+	 * Used to get a View type of the given ALP container
+	 *
+	 * @tparam T The ALP container to inspect.
+	 *
+	 */
+	template< typename Container >
+	struct inspect_view {};
```
hidanielesgit, line 204: internal? Perhaps a user-level `is_view` could be OK. Not sure when useful but at least users are aware of the concept and might want to use such a trait. But I don't think a user should/would be able to process View types as they shouldn't be accessible to her.

include/alp/type_traits.hpp
```diff
@@ -148,6 +185,71 @@ namespace alp {
...
+
+
+	namespace internal {
+
+		template< typename T >
+		struct is_view_over_concrete_container : is_view_over_concrete_container< typename inspect_view< T >::type::applied_to > {
```
hidanielesgit, line 210: What is the difference between this and `is_original`? I think the latter should be enough or am I missing something?

include/alp/type_traits.hpp
```diff
@@ -148,6 +185,71 @@ namespace alp {
...
+	 *
+	 * @tparam T The ALP container to inspect.
+	 *
+	 * The value is true if the provided type corresponds to an ALP container with
+	 * physical container.
+	 * The value is false if the provided type corresponds to an ALP container
+	 * based on a functor object.
+	 */
+	template< typename T >
+	struct is_concrete : internal::is_view_over_concrete_container< T  > {
```
hidanielesgit, line 231: This also seems redundant for similar reasons?

include/alp/vector.hpp
```diff
@@ -52,6 +52,18 @@ namespace alp {
...
 	class Vector;
 
+	/** Specializations of ALP backend-agnostic type traits */
+	template< typename T, typename Structure, enum Density density, typename View, typename Imf, enum Backend backend >
+	struct inspect_structure< Vector< T, Structure, density, View, Imf, backend > > {
+		typedef Structure type;
+	};
+
+	template< typename T, typename Structure, enum Density density, typename View, typename Imf, enum Backend backend >
+	struct inspect_view< Vector< T, Structure, density, View, Imf, backend > > {
```
hidanielesgit, line 62: internal?

include/alp/type_traits.hpp
```diff
@@ -148,6 +185,125 @@ namespace alp {
...
+		 *
+		 * @tparam T The view to inspect.
+		 *
+		 * \note A Matrix is storage-based if it has
+		 *       - an original view over void, or
+		 *       - any type of view over another storage-based matrix.
+		 *
+		 */
+		template< typename View >
+		struct is_storage_based : is_storage_based<
```
hidanielesgit, line 230: is_storage_based -> is_view_over_storage?

include/alp/type_traits.hpp
```diff
@@ -148,6 +185,125 @@ namespace alp {
...
+		 *
+		 * \note A Matrix is functor-based if it has
+		 *       - a functor view over a lambda type, or
+		 *       - any type of view over another functor-based matrix.
+		 *
+		 * @see is_functor_based_helper
+		 *
+		 */
+		template< typename View >
+		struct is_functor_based : is_functor_based_helper< View, typename View::applied_to > {};
```
hidanielesgit, line 281: is_functor_based -> is_view_over_functor?

include/alp/type_traits.hpp
```diff
@@ -148,6 +185,125 @@ namespace alp {
...
+		 * The value is true if the provided view corresponds to an ALP container that
+		 * - allocates memory for container storage, or
+		 * - allocates memory for a functor
+		 * The value is false otherwise, i.e., if the provided view type corresponds
+		 * to a view over another ALP container, and, therefore, does not need to
+		 * allocate memory for storage/functor.
+		 *
+		 */
+		template< typename View >
+		struct allocates_memory : std::integral_constant<
```
hidanielesgit, line 300: allocates_memory -> requires_allocation?

PR #28, opened 2022-07-07
337 parallelize buildmatrixunique for reference omp backend

anyzelman:
This is to merge several changes:
 - feature: OpenMP-parallel (shared-memory parallel) ingestion into grb::Matrix if input iterators are random access
 - feature: added type trait to determine whether a given type is an ALP/GraphBLAS sparse matrix iterator.
 - bugfix: matrix output iterator did not output any nonzeroes for matrices with all nonzeroes on the last row (did not trigger in other conditions)
 - bugfix: exceptions raised during matrix construction could result in uninitialised fields and a segfaulting destructor
 - bugfix: BSP1D artificially rejected building dense matrices (`<` instead of `<=`)
 - testing: more thorough testing of input and output iteration of matrices
 - testing: more thorough testing of matrix ingestion
 - code improvement: matrix iterators overhaul
 - code improvement: add `noexcept` to some internal buffer management routines
 - code improvement: code style fixes, documentation fixes and addtions, as well as code reorganisations chiefly related to iterators, matrix ingestion, and all files touched by these
 - code improvement: various dead code regions removed
 - code improvement: remove warnings due to superfluous `;` at namespace closing (only triggered by some compilers)

Contributed by Denis Jelovina and Alberto Scolari.

Resolves  internal issues 337, 64, 402, 404, and 474.


anyzelman:
Testing will not start before review has completed which, given the size of this MR, may take a while

anyzelman:
> there was a case where matrices could be instantiated with uninitialised fields

the bugfix here seems to address this on the level of an internal constructor which was leaving some fields uninitialised because it is intended to be followed by an internal call to initialize.

- [ ]  Thus need to check that this MR hence does not fix but a symptom, rather than a deeper bug somewhere else

anyzelman:
Update on TODOs before I forget:
 - move all iterator-related utilities in their own subdirectory
 - MatrixVectorIterator seems not updated with NonzeroStorage
 - buildMatrixUnique might be leaking the requirements of NonzeroStorage, not sure about that

anyzelman:
Review completed, starting tests now

anyzelman:
Full test suite (1403 tests) passed with GCC 9.3.1, LPF 1.0, and MPICH 3.3.2.

anyzelman:
Smoke tests (71 tests) and unit tests (1166 tests) passed with GCC 9.4.0, LPF 1.0, and MPICH 3.3.2 (Ubuntu 20.04 LTS, previous one was on Fedora).

More in-depth testing on different machine configurations will be done when merging develop into master as v0.6, which should happen soon. With the tests so far, merge of this MR into develop is approved.

PR #30, opened 2022-07-22
470 implement the change of amf on transposition

vdimiczrc:


include/alp/storage.hpp
```diff
@@ -109,6 +121,38 @@ namespace alp {
 				return Full_type( 0, 0, 0, dim, 1, 0 );
 			}
 
+			/**
+			 * Exposes the polynomial type that is a result of swapping x and y
+			 * coefficients of the provided polynomial type.
+			 */
+			template< typename Polynomial >
+			struct swap_x_y_coefficients {
```
hidanielesgit, line 129: Perhaps redundant given apply_view<transpose> does the same?

hidanielesgit (approved):

PR #35, opened 2022-08-05
472 grb set does not execute in parallel

anyzelman:
This MR contains the following changes:
- bugfix: an earlier MR moved the `grb::set` functions of the reference and reference_omp backends to `reference/io.cpp`, but did not update the parallelisation guards.
- code improvement: rely on `config::OMP::localRange` instead of parallel-for with a chunked static schedule. In several cases related to calls to `grb::set` and vector construction this leads to significant speedups.

This resolves internal issue 472 and contributes to internal issue 479.

Thanks to Alberto Scolari for contributing!


anyzelman:
Review OK, smoke tests OK, GitHub and internal CI OK. Will merge.

PR #36, opened 2022-08-06
479 openmp chunked static parallel for does not do the intended thing

anyzelman:
In the `reference_omp` backend, as part of the previous merge into develop (#35), Alberto detected that increasing the chunk size of statically-scheduled parallel-for loops could increase the performance noticeably.

The intent of the previous chunk size was to prevent false sharing effects by considering a unit of loop iterands considered by OpenMP to be at least of some size `k`, where `k` is the cache line size. The idea was that then no threads will cause more than O(1) false sharing effects, even if the loop touches single-byte elements-- or indeed zero such effects in the case all memory regions are aligned.

However, despite its name, the OpenMP chunked static schedule introduces a run-time component by assigning the chunks to threads in a round-robin fashion-- aside from run-time overhead, this may also cause more page misses than otherwise would occur.

This MR removes all parallel-fors with static chunked schedules and replaces it with code that only achieves the above intended effect-- a loop is chunked into units of `k` iterands, and then a static schedule is applied on a loop of size `ceil(n/k)`, where `n` is the actual loop size. This is done via the pre-existing (in ALP) `config::OMP::localRange` that was designed to achieve exactly the intended behaviour from within OpenMP parallel sections such as in `vxm_generic`.

This MR also replaces several uses of `#pragma omp for` from within such sections by this pattern.

It also adds some more sanity checking to calls to `grb::zip` (to matrix outputs) and fixes an older version of `grb::zip` still present in the Banshee backend. As always, it also includes some code style fixes across the files that this MR touches.

This resolves internal issue #479.

### DIFF
--- a/include/graphblas/banshee/blas3.hpp
+++ b/include/graphblas/banshee/blas3.hpp
@@ -326,11 +326,18 @@ namespace grb {
 
 	namespace internal {
 
-		template< Descriptor descr = descriptors::no_operation, bool matrix_is_void, typename OutputType, typename InputType1, typename InputType2, typename InputType3, typename Coords >
-		RC matrix_zip_generic( Matrix< OutputType, banshee > & A,
-			const Vector< InputType1, banshee, Coords > & x,
-			const Vector< InputType2, banshee, Coords > & y,
-			const Vector< InputType3, banshee, Coords > & z ) {
+		template<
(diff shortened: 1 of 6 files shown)


anyzelman:
Full test suite OK, internal CI OK, GitHub CI OK.

anyzelman:
Final review OK, will merge

PR #39, opened 2022-08-08
462 copy move assigment of compressed storage iterator is missing the nonzero variable

anyzelman:
Anders reported that copy-assignment of matrix output iterators was not properly implemented -- the local nonzero cache was not updated. The same was true for the copy-constructor, move-constructor, and move-assignment. This MR:
1. fixes all those issues, and 
2. hardens the matrixIterator unit test to test all four of {move,copy}x{assignment,construction} for all combination of pattern- and non-pattern matrices it already was testing for.

Additionally,
3. The assignment operators of the pattern-matrix output iterator specialisation did not return a reference to itself, with this MR also fixed;
4. the matrix output iterator itself now also adheres to the ALP matrix iterator API-- i.e., it defines an i, j, and v member function that return the row, column, and value of a nonzero, respectively. It also now and accordingly defines the `RowIndexType`, `ColumnIndexType`, and `ValueType` typedefs;
5. the SynchronizedNonzeroIterator was missing the ALP `ValueType` typedef for pattern matrices, herewith fixed.

This resolves internal issue #462.

### DIFF
--- a/include/graphblas/SynchronizedNonzeroIterator.hpp
+++ b/include/graphblas/SynchronizedNonzeroIterator.hpp
@@ -418,6 +418,9 @@ namespace grb {
 
 			public:
 
+				/** ALP value typedef */
+				typedef void ValueType;
+
 				/** The type of this class for a short-hand. */
 				using SelfType = SynchronizedNonzeroIterator<
 					RowIndexT, ColIndexT, void,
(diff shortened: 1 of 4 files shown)


anyzelman:
Internal CI OK. Smoke tests, unit tests (with LPF) and code review pending.

anyzelman:
Review, unit tests, and smoke tests OK. Will merge

PR #40, opened 2022-08-09
464 dense spmv test does not work for ecology2 mtx

anyzelman:
Anders reported that the `tests/unit/dense_spmv.cpp` failed verification for `ecology2.mtx`. The reason was that the test relies on `grb::utils::equals` to compare numbers vs. an absolute error bound, however, the utility function is written for relative error checking. This MR clarifies the documentation of `grb::utils::equals', and hardens its API somewhat to prevent misuse.

Summarising, this MR:
- implements an absolute error check for `dense_spmv.cpp` instead of relying on `grb::utils::equals`;
- bugfix: BiCGstab also relied on `grb::utils::equals` while it should not;
- bugfix `grb::utils::equals`: the given relative bound was interpreted twice as loose;
- bugfix `grb::utils::equals`: overflow of the normalisation factor was handled questionably;
- testing: unit test added for `grb::utils::equals` to prevent regressions of its implementation;
- code improvement: number of epsilons to `grb::utils::equals` must be integer to help prevent misuse of `grb::utils::equals`;
- tests that did not pass integer epsilons to `grb::utils::equals` are now fixed;
- improved documentation of `grb::utils::equals` to help prevent its misuse.

All unit tests touched were modernised:
- grb::{init,finalize}->grb::Launcher,
- do not use printf/fprintf,
- standardised use of stderr/stdout and test result printing.

Additional loosely related items:
- BiCGstab test for orthogonality (As vs. s) compared the wrong scalar output (was rho, should be temp), now fixed;
- fixed typo in `smoketests.sh` reporting on two related tests.

The MR includes code style fixes through most files it touches.

### DIFF
--- a/include/graphblas/algorithms/bicgstab.hpp
+++ b/include/graphblas/algorithms/bicgstab.hpp
@@ -314,7 +314,7 @@ namespace grb {
 #ifdef _DEBUG
 				std::cout << "\t\t rho = " << rho << "\n";
 #endif
-				if( ret == SUCCESS && utils::equals( rho, zero, 2*n-1 ) ) {
+				if( ret == SUCCESS && rho == zero ) {
 					std::cerr << "Error: BiCGstab detects r at iteration " << iterations <<
 						" is orthogonal to r-hat\n";
 					return FAILED;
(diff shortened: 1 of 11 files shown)


anyzelman:
Review found one minor issue, fixed, and flagging review passed. Waiting for tests and CIs.

anyzelman:
Review OK, smoke & unit tests (with LPF) OK, CIs OK. Will merge.

PR #41, opened 2022-08-09
490 dynamic scheduling for regular loop

anyzelman:
This MR reviews all uses of the dynamic schedule and replaces it with a static schedule (through `grb::config::OMP::localRange`) in case the loop is determined to be regular. In all irregular loop cases, documentation has been added with rationale as per the schedule choice, as well as for cases where the choice is not clear-cut.

This MR also modifies the following unrelated behaviour:
- eWise{Map,Lambda} no longer serialise lambda invocations when `_DEBUG` is defined. Rationale: if this is desired, the user can add this to the lambda itself-- by default, it is desirable that lambdas run concurrently in order to diagnose any data race issues, for example.

As always, this MR includes some code style fixes throughout all files touched.

### DIFF
--- a/include/graphblas/reference/blas1.hpp
+++ b/include/graphblas/reference/blas1.hpp
@@ -413,8 +413,6 @@ namespace grb {
 						(void) foldr< descr >( scalar, x[ index ], op );
 					}
 				}
-			}
-
 			} else {
 #ifdef _H_GRB_REFERENCE_OMP_BLAS1
 				#pragma omp parallel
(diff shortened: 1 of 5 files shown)


anyzelman:
Review OK. GitHub CI OK. Smoke tests (with LPF) OK. Unit tests and internal CI running.

anyzelman:
Internal CI OK, unit tests (with LPF) still pending

anyzelman:
Unit tests OK, will merge.

PR #45, opened 2022-08-10
422 clear setelement and usage of pinnedvector leads to segementation fault

anyzelman:
Aristeidis and Anders reported an issue where combinations of `clear`, `setElement`, and usage of the `PinnedVector` could segfault for the `reference` and `reference_omp` backends. This MR first and foremost introduces a unit test for the `PinnedVector` that operates on vectors of `double`s, vectors of `std::pair`s, and tests it using both the SEQUENTIAL and PARALLEL I/O modes.

This added unit test found or confirmed the following issues, which this MR also fixes:
- `grb::set` (BSP1D) contains code to handle the `use_index` descriptor, but this code should only pass by the compiler if `size_t`s are convertible into the vector element type;
- `grb::PinnedVector` constructor (reference, reference_omp, and BSP1D) should pin the stack, not the bitmask;
- `grb::clear` (BSP1D) did not undo the `became_dense` flag, potentially resulting in a distributed vector that is both cleared and became dense;
- `grb::PinnedVector` constructor (BSP1D) now throws expections instead of silently ignoring errors during construction.

Also minor performance enhancements, code style fixes, better debug tracing, code documentation improvements, and a fixed typo.

This MR closes internal issue #422.

### DIFF
--- a/include/graphblas/bsp1d/exec.hpp
+++ b/include/graphblas/bsp1d/exec.hpp
@@ -37,8 +37,9 @@
 #include "init.hpp"
 
 #ifndef _GRB_NO_STDIO
-#include <iostream> //for std::cerr
+ #include <iostream> //for std::cerr
 #endif
+
 
 /** Global internal singleton to track whether MPI was initialized. */
 extern bool _grb_mpi_initialized;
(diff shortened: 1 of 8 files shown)


anyzelman:
Review OK. GitHub and internal CI OK. Smoke tests (with LPF) OK. Waiting for unit tests before merge.

anyzelman:
All unit tests OK, will merge

PR #50, opened 2022-08-15
Resolves internal issue #322: number of sockets detection on some machines fails

anyzelman:
This MR prints a warning when the auto-detection of the number of sockets for performance tests fails, and selects a nonzero default instead (i.e., 1). The previous behaviour silently selected 0 sockets on detection failure instead, which would result in performance tests for the hybrid backend to not run.


anyzelman:
Review OK, waiting CIs.

anyzelman:
CI OK, merging

PR #52, opened 2022-08-15
Resolve internal issue #137: re-enable overlap detection between output vector and output mask

anyzelman:
In sparse matrix--vector multiplication, there was disabled code detecting whether the output and output mask were the same vector. When considering the scattering SpMV multiplication, which consists of a loop of updates to the output vector, in such an overlapping case the intermediate scatters would overwrite the mask, potentially altering it during the multiplication. Therefore surely this overlap must be illegal (as it also is defined in the code documentation as well as the initial ALP/GraphBLAS paper), and the overlap check must be re-enabled.

This MR makes indeed re-enables the check, and revises the documentation of `grb::mxv` and `grb::vxm` in `base/blas2.hpp` to bring it up to date with the current state, and in particular with regards to the overlap detection and the `descriptors::safe_overlap` descriptor.

It additionally revises the code in `base/blas2.hpp` to incorporate `grb::Phase` arguments, and takes care to only document the full versions of `grb::mxv` and `grb::vxm`, and not the short-hand versions with mask arguments omitted. It also adds some missing SFINAE checks in the file, and moves the (unchanged) mxv and vxm performance semantics to the backends that define them.

### DIFF
--- a/include/graphblas/base/blas2.hpp
+++ b/include/graphblas/base/blas2.hpp
@@ -146,6 +146,8 @@ namespace grb {
 	 *                     \a u, \a A, and \a v will be cast to the domains of the
 	 *                     additive and multiplicative operators of \a semiring.
 	 *
+	 * The vector \a v may not be the same as \a u.
+	 *
 	 * Instead of passing a \a semiring, users may opt to provide an additive
 	 * commutative monoid and a binary multiplicative operator instead. In this
 	 * case, \a A may not be a pattern matrix (that is, it must not be of type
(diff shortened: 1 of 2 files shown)


anyzelman:
All tests OK. Remaining TODOs:
- [ ] review code change
- [ ] update documentation for mxv and vxm

anyzelman:
Review OK, and all TODOs resolved. Waiting CIs OK and run of all smoke and unit tests (with LPF), and will merge if all OK.

anyzelman:
All CIs and tests OK, will merge

PR #55, opened 2022-08-17
Resolve internal issue #495: fold from sparse vectors into scalars using monoids returns illegal

anyzelman:
This is a regression compared to earlier versions that did accept reducing a sparse vector into a single scalar, using a Monoid structure-- earlier commits reviewed the behaviour under the dense descriptor, but erroneously made fold-into-scalars require dense always. This MR fixes this to only return `ILLEGAL` for sparse input vectors if the dense descriptor is also present.

The MR also found and fixes several bugs:
- casting behaviour was incorrect -- all intermediate results were previously cast to the output scalar type. (Now they are first cast to the monoid output domain.)
- the previously implemented version was not asymptotically optimal in case of sparse vectors and/or masks, now fixed.
- not all trivial cases were caught before dispatching. Prior to this MR this was mostly only a performance bug.

The MR also improves the following items:
- refactor the generic function into a dispatcher, and factored out the dense and sparse variants into separate (internal) functions.
- code previously also allowed sequential dense reductions for associative operators (i.e., non-monoid structures). This is now simplified by assuming monoid structures always, and reduces complexity and performance significantly.
- the unit test for reductions is expanded with testing for sparse vector reductions.

The code simplification described above also removed one previously necessary compiler warning suppression.

### DIFF
--- a/docs/Suppressions.md
+++ b/docs/Suppressions.md
@@ -57,32 +57,7 @@ These are indirectly caused by the following calls:
 
 These are all OK to suppress since the reads are masked.
 
-5. `include/graphblas/reference/blas1.hpp`, fold_from_vector_to_scalar_generic:
-```
-GRB_UTIL_IGNORE_MAYBE_UNINITIALIZED // the below code ensures to set local
-IOType local;                       // whenever our local block is
-GRB_UTIL_RESTORE_WARNINGS           // non-empty
-if( end > 0 ) {
-	if( i < end ) {
-		local = static_cast< IOType >( internal::getRaw( to_fold )[ i ] );
-	} else {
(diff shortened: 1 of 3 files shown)


anyzelman:
All CIs OK. Full test suite (with LPF) pending. Review pending.

anyzelman:
Added one more test case, re-running tests and CIs.

anyzelman:
Review OK

anyzelman:
Tests with LPF revealed a bug. Fixed it and hardened unit test. Review again OK, GitHub CI OK, internal CI and unit tests (with LPF) pending.

anyzelman:
Internal CI OK

anyzelman:
1190 unit tests OK-- will merge

PR #66, opened 2022-09-06
505 dense dot norm2

djelovina:


tests/unit/dense_dot_norm2.cpp
```diff
@@ -22,125 +22,201 @@
...
+			rc = FAILED;
+			return;
+		}
+	}
+
+	{
+		// test 4, init
+		alp::Vector< int > empty_left( 0 ), empty_right( 0 );
+		setInitialized( empty_left, true );
+		setInitialized( empty_right, true );
```
vdimiczrc, line 202: `setInitialized` belongs to `internal` namespace

tests/unit/dense_dot_norm2.cpp
```diff
@@ -22,125 +22,201 @@
...
+			std::cerr << "\t test 3 (dense integer vectors) unexpected value "
+				  << *alpha << ", expected 2 * n = " << ( 2 * n) << ".\n";
+			rc = FAILED;
+			return;
+		}
+	}
+
+	{
+		// test 4, init
+		alp::Vector< int > empty_left( 0 ), empty_right( 0 );
```
vdimiczrc, line 200: I think it is more readable if we declare each variable in a separate line, like:
```
alp::Vector< int > empty_left( 0 );
alp::Vector< int > empty_right( 0 );
```

vdimiczrc (approved):

PR #74, opened 2022-09-16
274 automatic extraction of hyperdags from alp

anyzelman:
Introduces a new backend that intercepts ALP/GraphBLAS calls and, on finalisation of the ALP/GraphBLAS context, writes out the computational structure of executed programs as a HyperDAG. It delegates actual computations to the standard `reference` backend.

This MR is largely authored by Aikaterini Karanasiou, and includes the following:
- bugfix: copy-construction of ALP/GraphBLAS reference and reference_omp matrices could fail;
- specification: all primitives (except getters) now take a phase argument;
- tooling: added a small script to detect common spacing errors in code formatting;
- various code style fixes.


anyzelman:
Several issues to be fixed before code review:

1. ops should not be registered
2. monoids should not be registered
3. semirings should not be registered
4. not all scalar sources are registered using addSource
5. some unit tests show the output hyperdags, while this should be hidden (harden unittests.sh against this)
6. make/build from scratch shows quite a few warnings that need to be resolved

anyzelman:
There's another issue: a lot of the logic is based on pointers, while likely it should be using the GraphBLAS IDs?

anyzelman:
Philosophical question: should getters indeed be captured in a hyperdag? (currently this is done-- unless their are objections, I'm thinking to keep it-- better capture too much rather than too little)

bootstrap.sh
```diff
@@ -146,6 +152,13 @@ or assume default paths (--with-lpf)"
 	--no-reference)
 			reference=no
 			;;
+	--no-hyperdags)
+			hyperdags=no
+			;;
+	--with-hyperdags-using=*)
+			hyperdags=yes
+			hyperdags_uses="${arg#--with-hyperdags-using=}"
```
anyzelman, line 160: should this not be `hyperdags_using=...` ?

anyzelman (commented):
Quick review of build infra changes

anyzelman:
> Philosophical question: should getters indeed be captured in a hyperdag? (currently this is done-- unless their are objections, I'm thinking to keep it-- better capture too much rather than too little)

Related:
- [ ] `grb::set` does not register the source scalar but instead re-registers the container. This seems odd/inconsistent.
- [ ] `grb::setElement` does not register the source index.

anyzelman:
Various I/O functions are not in io.hpp (but rather scattered around blas{0-3}.hpp)

anyzelman:
> There's another issue: a lot of the logic is based on pointers, while likely it should be using the GraphBLAS IDs?

Also the README mentions this as the (last) remaining TODO. Will look at it momentarily.

anyzelman:
Code review complete. Waiting feedback from Katerina and Georg before merge.

anyzelman:
Full test run says performance tests fail (due to never having been enabled)... fixing now.

anyzelman:
Internal CI OK

PR #82, opened 2022-09-22
Resolve internal issue #530: Integrate gemm-like example

vdimiczrc:


tests/unit/alp_gemm.cpp
```diff
@@ -35,26 +56,107 @@ void alp_program( const size_t & unit, alp::RC & rc ) {
...
-		A, 1, 1, 1, 1,
-		B, 2, 1, 2, 4,
-		beta,
-		C, 0, 0, 1, 1,
-		ring );
+	// Initialize containers A, B, C, alpha, beta
+	constexpr double A_value = 2;
+	constexpr double B_value = 3;
+	constexpr double C_value = 4;
+	rc = rc ? rc : alp::set( A, alp::Scalar< double >( A_value ) );
```
hidanielesgit, line 67: Would this (and next line) make it harder for us to assess the correct functioning of transposition?

djelovina (approved):

tests/unit/alp_gemm.cpp
vdimiczrc: Implemented matrix initialization with random values.

hidanielesgit (approved):

PR #83, opened 2022-09-22
Resolve internal issue #494: Implement matrix set via fold

vdimiczrc:


anyzelman:
oh, this is a nice use of right_assign!

djelovina (approved):

include/alp/reference/blas3.hpp
```diff
@@ -1649,84 +1577,10 @@ namespace alp {
...
 		}
 
-		RC rc = internal::set_band< 0 >( C, A );
-
-		if( rc == SUCCESS ) {
-			internal::setInitialized( C, true );
-		}
-
-		return rc;
+		internal::setInitialized( C, true );
```
hidanielesgit, line 1580: Shouldn't this depend on the init status of the A?

include/alp/reference/blas3.hpp
```diff
@@ -1772,13 +1630,8 @@ namespace alp {
...
 		}
 
-		RC rc = internal::set_band< 0 >( C, val );
-
-		if( rc == SUCCESS ) {
-			internal::setInitialized( C, true );
-		}
-
-		return rc;
+		internal::setInitialized( C, true );
```
hidanielesgit, line 1633: Shouldn't this depend on the init status of the A?

include/alp/reference/blas3.hpp
vdimiczrc: Initi status of the A is checked in line 1575 and the function returns if it is not. If we arrive to this point, it means that A is initialized.

include/alp/reference/blas3.hpp
vdimiczrc: Same answer as above.

hidanielesgit (approved):

PR #85, opened 2022-09-27
536 copying vector by assignment multiple times fails asserts

anyzelman:
Bugfix for the following sequence that previously triggered an assertion failure for the `reference_omp` backend:
```
Vector<double> a(n), b(n);
set(a, 1.2);
set(b, 1.5);
a=b;
a=b; //Fails assertion
```

Issue: multiple-registration of the same container.

Cause: the previous implementation of the copy-assignment seems to predate the introduction of IDs. In its implementation, the container `a` is in fact replaced by a totally new container, with a different ID, through the use of `std::swap` and the creation of a new container within the copy-assignment. Additionally, this induces new memory allocations while in principle a dispatch to `grb::set` would have sufficed.

Solution: dispatch to `grb::set` instead. Apart from the performance gain, this also retains the container's original ID and does not trigger the earlier double-registration issue and thus does not trigger the reported assertion failure. Additionally, this MR adds a unit test for double copy-assignment.

This MR also includes
- bugfix: BSP1D `grb::set` was not correctly updated as part of the previous MR (internal issue 533);
- implement previously missing copy-assignment for BSP1D backend;
- improved `_DEBUG` tracing;
- minor code style fixes.

Reported by Anders Hansson -- many thanks!

### DIFF
--- a/include/graphblas/bsp1d/io.hpp
+++ b/include/graphblas/bsp1d/io.hpp
@@ -504,6 +504,13 @@ namespace grb {
 	) noexcept {
 		const size_t n = size( x );
 		const size_t old_nnz = nnz( x );
+
+		// dynamic checks
+		if( (descr & descriptors::dense) && nnz( x ) < n ) {
+			return ILLEGAL;
+		}
+
+		// capacity check
 		if( capacity( x ) < n ) {
 			if( phase == RESIZE ) {
 				return resize( x, n );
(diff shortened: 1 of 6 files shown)


anyzelman:
Review OK. Waiting CIs OK. Also running all unit tests with LPF before merge.

anyzelman:
BSP1D unit test failed. Traced back to previous MR. Tested hotfix in this branch now -- all unit tests with LPF OK.

After hotfix, all internal CIs also OK.

anyzelman:
Code review of hotfix OK. Will merge.

PR #89, opened 2022-09-30
Resolve internal issue #549: Add conjugate operation for ALP matrix

vdimiczrc:


include/alp/reference/blas2.hpp
```diff
@@ -1054,6 +1055,118 @@ namespace alp {
...
+			);
+
+	}
+
+	/** Specialization for square matrices */
+	template<
+		Descriptor descr = descriptors::no_operation,
+		typename DataType, typename Structure, typename View, typename ImfR, typename ImfC,
+		std::enable_if_t<
+			structures::is_in< structures::Square, typename Structure::inferred_structures >::value
```
hidanielesgit, line 1129: There is an `is_a<Structure, Square>` type_trait we can use for this purpose.

tests/unit/dense_conjugate.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+	rc = rc ? rc : alp::foldl( E, A, ring.getAdditiveOperator() );
+	rc = rc ? rc : alp::foldl( E, B, alp::operators::subtract< T >() );
+
+	float fnorm = ring.template getZero< float >();
+	rc = rc ? rc : alp::eWiseLambda(
+		[ &fnorm, &ring ]( const size_t i, const size_t j, T &val ) {
+			(void) i;
+			(void) j;
+			const float valsquare = static_cast< float >( std::real( val * grb::utils::is_complex< T >::conjugate( val ) ) );
```
djelovina, line 89: I would rather use `std::norm( val )` then  `val * conjugate( val )`, because the correct type will be automatically returned (i.e. `double` or `float`).

tests/unit/dense_conjugate.cpp
```diff
@@ -0,0 +1,194 @@
...
+			alp::internal::foldl(
+				fnorm,
+				valsquare,
+				alp::operators::add< float >()
```
djelovina, line 93: add operator should be from the ring

tests/unit/dense_conjugate.cpp
```diff
@@ -0,0 +1,194 @@
...
+		return alp::FAILED;
+	}
+}
+
+template<
+	typename T,
+	typename Structure = typename std::conditional<
+		grb::utils::is_complex< T >::value,
+		alp::structures::Hermitian,
+		alp::structures::Square
```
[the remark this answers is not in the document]
vdimiczrc: Done

tests/unit/dense_conjugate.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+void alp_program( const size_t &n, alp::RC &rc ) {
+
+	rc = alp::SUCCESS;
+
+	rc = rc ? rc : test_conjugate< std::complex< float > >( n );
+	rc = rc ? rc : test_conjugate< float >( n );
+
+	rc = alp::SUCCESS;
```
djelovina, line 147: check `rc` here

tests/unit/dense_conjugate.cpp
```diff
@@ -0,0 +1,194 @@
...
+template< typename T >
+T random_value();
+
+template<>
+float random_value< float >() {
+	return static_cast< float >( rand() ) / RAND_MAX;
+}
+
+template<>
+std::complex< float > random_value< std::complex< float > >() {
```
djelovina, line 40: I would set alias for `folat` at one place (beginning of this file) so it can be easily changed

djelovina (approved):

PR #93, opened 2022-10-03
532 integrate householder with dense alp containers

djelovina:


include/alp/algorithms/householder_tridiag.hpp
```diff
@@ -73,82 +103,125 @@ namespace alp {
...
-									foldl(v [ i ], alpha, minus );
-								}
-							},
-							v );
-
-				Scalar< D > norm_v;
+					[ &alpha, &ring, &divide, &minus ]( const size_t i, D &val ) {
+						if ( i == 0 ) {
+							Scalar< D > norm_v0( std::abs( val ) );
+							internal::foldl( *alpha, val, ring.getMultiplicativeOperator() );
```
hidanielesgit, line 126: Could we use `alp::foldl(alpha, ...)` instead of `internal::foldl` ?
Same holds for the other operations in this block.

include/alp/algorithms/householder_tridiag.hpp
```diff
@@ -73,82 +103,125 @@ namespace alp {
...
+
+#ifdef DEBUG
+				print_matrix( " vvt ", vvt );
+#endif
 
 				// Qk = Qk - vvt ( expanded: I - 2 * vvt )
-				rc = foldl( Qk, vvt, minus );
+				auto Qk_view = get_view( Qk, utils::range( k + 1, n ), utils::range( k + 1, n ) );
+				//conjugate view should replace this check
+				if ( grb::utils::is_complex< D >::value ) {
```
hidanielesgit, line 173: I think since a couple of PRs ago we do have a `conjugate` view :)

include/alp/reference/blas1.hpp
```diff
@@ -2634,15 +2639,15 @@ namespace alp {
 		void >::type * const = NULL
```
hidanielesgit, line 2639: Commenting here since I cannot do this near the signature above.
Are we using this variant variant of `dot` anyware? Shouldn't the one with `alp::Scalar` IO be sufficient?

include/alp/reference/blas1.hpp
```diff
@@ -2923,15 +2930,16 @@ namespace alp {
 		const Vector< InputType, InputStructure, Density::Dense, InputView, InputImfR, InputImfC, backend > &y,
 		const Ring &ring = Ring(),
 		const typename std::enable_if<
-			std::is_floating_point< OutputType >::value,
-		void >::type * const = nullptr
+			std::is_floating_point< OutputType >::value || grb::utils::is_complex< OutputType >::value,
+			void
+		>::type * const = nullptr
 	) {
```
hidanielesgit, line 2936: Similar comment as for `dot`.

tests/smoke/alp_zhetrd.cpp
```diff
@@ -0,0 +1,345 @@
...
+using HermitianOrSymmetricTridiagonal = structures::SymmetricTridiagonal;
+//fully implemented structures
+using HermitianOrSymmetric = structures::Symmetric;
+#endif
+
+constexpr BaseScalarType tol = 1.e-10;
+constexpr size_t RNDSEED = 1;
+
+//temp function untill Hermitian containter is implemented
+//** gnerate symmetric-hermitian matrix in a square container */
```
hidanielesgit, line 55: gnerate -> generate

tests/smoke/alp_zhetrd.cpp
```diff
@@ -0,0 +1,345 @@
...
+	std::vector< T > data( N * N );
+	std::srand( RNDSEED );
+	for( size_t i = 0; i < N; ++i ) {
+		for( size_t j = i; j < N; ++j ) {
+			T val( std::rand(), std::rand() );
+			data[ i * N + j ] = val / std::abs( val );
+			if( j != i ) {
+				data[ j * N + i ] = grb::utils::is_complex< T >::conjugate( data[ i * N + j ] );
+			}
+			if( j == i ) {
```
hidanielesgit, line 75: Not a big deal since temporary but we could just get rid of the ifs and simplify the body

tests/smoke/smoketests.sh
```diff
@@ -538,10 +538,27 @@ for BACKEND in ${BACKENDS[@]}; do
...
+	NTEST_HOUSEHOLDER=100
+	echo ">>>      [x]           [ ]       Tests zhetrd (Householder tridiagonalisaiton) on"
+	echo ">>>                               a real symmetric matrix (${NTEST_HOUSEHOLDER}x${NTEST_HOUSEHOLDER})."
+	bash -c "$runner ${TEST_BIN_DIR}/alp_zhetrd_${BACKEND} ${NTEST_HOUSEHOLDER} &> ${TEST_OUT_DIR}/alp_zhetrd_${BACKEND}.log"
+	head -1 ${TEST_OUT_DIR}/alp_zhetrd_${BACKEND}.log
+	grep 'Test OK' ${TEST_OUT_DIR}/alp_zhetrd_${BACKEND}.log || echo "Test FAILED"
+	echo " "
+
+	NTEST_HOUSEHOLDER_COMPLEX=100
+	echo ">>>      [x]           [ ]       Tests zhetrd (Householder tridiagonalisaiton) on"
```
hidanielesgit, line 551: zhetrd -> dsytrd? Just to clarify the analogous LAPACK call

include/alp/reference/matrix.hpp
```diff
@@ -2251,7 +2251,7 @@ namespace alp {
 
 			template < bool d >
 			struct view_type< view::gather, d > {
-				using type = Matrix< T, structures::General, Density::Dense, view::Gather< self_type >, imf::Strided, imf::Strided, reference >;
+				using type = Matrix< T, structures::Hermitian, Density::Dense, view::Gather< self_type >, imf::Strided, imf::Strided, reference >;
```
hidanielesgit, line 2254: Not sure we can guarantee this. Perhaps best to keep `General` and once a submatrix is gathered view it as Hermitian if this is acceptable in the particular context where this type cast is needed.

include/alp/algorithms/householder_tridiag.hpp
djelovina: Yes, at the expense of using one more Scalar< D >, and unpacking it at the end.

include/alp/algorithms/householder_tridiag.hpp
djelovina: An old comment, I don't need to conjugate here as I took vector v from row rather than from column (it's already conjugated).

include/alp/reference/blas1.hpp
djelovina: We don't use it, it was already in the sparse backed.

include/alp/reference/blas1.hpp
djelovina: same. If you I can remove c++ scalar versions.

tests/smoke/alp_zhetrd.cpp
djelovina: Yes

include/alp/reference/matrix.hpp
djelovina: This will break `L173` in `householder_tridiag.hpp`
`auto Qk_view = get_view( Qk, utils::range( k + 1, n ), utils::range( k + 1, n ) );`
since we want to use `Qk_view` on the matrix with Symmetric/Hermitian structure in the next `foldl`.

PR #96, opened 2022-10-15
533 out of place set does not clear the output vector

anyzelman:
The `grb::set` did not follow out-of-place semantics, while one unit test designed to catch that was bugged. Additionally, the label propagation smoke test relied on the older in-place semantics.

This MR makes sure the existing nonzero structure of the output vector is cleared whenever the dense descriptor is not given. It fixes the corresponding unit test, and replaces the `grb::set( vector, mask, vector )` used by the label propagation algorithm (used for clamping) with the in-place `grb::foldl( vector, mask, vector )` using the `right_assign` operator.

This MR also includes the related changes and fixes listed below.
- Bugfix: introduces the previously missing masked vector-to-vector `grb::foldl` in the BSP1D backend;
- Bugfix: masked vector-to-vector folds with dense vectors but sparse masks would return `ILLEGAL` (the generic implementation could execute it, but did not because it was executing a dynamic check on the dense descriptor using a template argument that was not the exact reverse of being passed that descriptor-- this MR fixes this by lifting checks w.r.t. the dense descriptor out of the generic implementation);
- Bugfix: several variants of `grb::set` did not correctly return `ILLEGAL` when the dense descriptor was used;
- Performance: exchange `grb::set` with `std::swap` at various places in the label propagation algorithm;
- Performance: `grb::set` variants now do not call `assignAll` nor `clear` when a dense structure is assured to remain;
- Testing: unify output from label smoke tests and minor typo fix in `smoketests.sh`;
- Testing: `grb::set` unit test now also tests for correct behaviour when the dense descriptor is used;
- Code improvements: various code style fixes across files touched by this MR;
- Semantics & documentation: dense descriptor also applies to output containers.

Thanks to Aristeidis for finding the bug and implementing the initial fix to `grb::set`!

### DIFF
--- a/include/graphblas/algorithms/label.hpp
+++ b/include/graphblas/algorithms/label.hpp
@@ -275,11 +275,13 @@ namespace grb {
 					std::swap( out, f );
 					return SUCCESS;
 				}
-			} else {
-				std::cerr << "Info: label propagation exiting with " << toString(ret) << "\n";
 			}
 
 			// done
+			if( s == 0 ) {
+				std::cerr << "Warning: label propagation exiting with " << toString(ret)
+					<< "\n";
+			}
 			return ret;
 		}
 
--- a/include/graphblas/algorithms/label.hpp
+++ b/include/graphblas/algorithms/label.hpp
@@ -116,7 +116,8 @@ namespace grb {
 		 *     accelerating the PageRank computation', ACM Press, 2003.
 		 */
 		template< typename IOType >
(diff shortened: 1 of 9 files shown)


anyzelman:
Both CIs OK, all unit and smoke tests (with LPF, as well as some perftests) OK. Pending code review before merge.

anyzelman:
Code review complete. Waiting internal CI OK before merge

PR #98, opened 2022-10-17
569 householder qr decomposition

djelovina:


include/alp/algorithms/householder_qr.hpp
```diff
@@ -0,0 +1,244 @@
...
+
+				// ===== Calculate reflector Qk =====
+				// Q_k = identity( n )
+				typedef typename std::conditional<
+					grb::utils::is_complex< D >::value,
+					structures::Hermitian,
+					structures::Symmetric
+				>::type SymmOrHerm;
+				Matrix<	D, SymmOrHerm, Dense > Qk( n );
+				rc = alp::set( Qk, zero );
```
vdimiczrc, line 170: Here `alp::` is used, but in most other places it is not. Consider either removing it here or, maybe better, adding it to other calls to alp primitives.

include/alp/algorithms/householder_qr.hpp
```diff
@@ -0,0 +1,244 @@
...
+				print_matrix( " << RR( updated ) >> ", RR );
+#endif
+				// ===== End of Update R =====
+
+				// ===== Update Q =====
+				// Q = Q * conjugate(transpose(Qk))
+
+				// Qtmp = Q * conjugate(transpose(Qk))
+				rc = rc ? rc : set( Qtmp, zero );
+				if ( grb::utils::is_complex< D >::value ) {
```
vdimiczrc, line 220: Extra space between `if` and `(`

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+using ScalarType = BaseScalarType;
+#endif
+
+constexpr BaseScalarType tol = 1.e-10;
+constexpr size_t RNDSEED = 1;
+
+//** generate random rectangular matrix data: complex version */
+template<
+	typename T
+>
```
vdimiczrc, line 53: This could maybe be a single line.

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+	const typename std::enable_if<
+		!grb::utils::is_complex< T >::value,
+		void
+	>::type * const = nullptr
+) {
+	std::vector< T > data( N * M );
+	std::srand( RNDSEED );
+	for( size_t i = 0; i < N; ++i ) {
+		for( size_t j = 0; j < M; ++j ) {
+			//data[ k ] = static_cast< T >( i + j*j ); // easily reproducible
```
vdimiczrc, line 89: Do we need this comment?

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+			data[ i * M + j ] = static_cast< T >( std::rand() )  / RAND_MAX;
```
vdimiczrc, line 90: If `T` is an integer type, the assigned value would always be 0.
There is one space too many before `/ RAND_MAX`

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+	class Ring = Semiring< operators::add< T >, operators::mul< T >, identities::zero, identities::one >
+>
+RC check_overlap( alp::Matrix< T, Structure, alp::Density::Dense, ViewType > &Q, const Ring & ring = Ring() ) {
+	// more elegent would be check mxm(Q,conjugate(transpose(Q))) == identity
+	RC rc = SUCCESS;
+	const size_t n = nrows( Q );
+#ifdef DEBUG
+	std::cout << "Overlap matrix for Q:\n";
+#endif
+	for ( size_t i = 0; i < n; ++i ) {
```
vdimiczrc, line 110: extra space after `for`

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+		auto vi = get_view( Q, i, utils::range( 0, n ) );
+		for ( size_t j = 0; j < n; ++j ) {
```
vdimiczrc, line 112: extra space after `for`

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+	size_t N = unit;
+	size_t M = 2 * unit;
+
+	alp::Matrix< ScalarType, Orthogonal > Q( N );
+	alp::Matrix< ScalarType, General > R( N, M );
+	alp::Matrix< ScalarType, General > H( N, M );
+	{
+		auto matrix_data = generate_rectangular_matrix_data< ScalarType >( N, M );
+		rc = rc ? rc : alp::buildMatrix( H, matrix_data.begin(), matrix_data.end() );
+	}
```
vdimiczrc, line 231: Why is this in a separate scope?

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+		std::cout << "Error: mratrix Q is not orthogonal\n";
+	}
+
+	rc = check_solution( H, Q, R );
+	if( rc != SUCCESS ) {
+		std::cout << "Error: solution numerically wrong\n";
+	}
+}
+
+int main( int argc, char ** argv ) {
```
vdimiczrc, line 255: No space between ** and argv.

include/alp/algorithms/householder_qr.hpp
```diff
@@ -0,0 +1,244 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <iostream>
+#include <sstream>
+
+#include <alp.hpp>
+#include <graphblas/utils/iscomplex.hpp> // use from grb
+#include "../tests/utils/print_alp_containers.hpp"
```
vdimiczrc, line 22: If this is included just for debug-time printing, maybe the include could also be guarded with `#ifdef DEBUG`

include/alp/algorithms/householder_qr.hpp
```diff
@@ -0,0 +1,244 @@
...
+			typename GeneralType,
+			typename OrthogonalType,
+			class Ring = Semiring< operators::add< D >, operators::mul< D >, identities::zero, identities::one >,
+			class Minus = operators::subtract< D >,
+			class Divide = operators::divide< D >
+		>
+		RC householder_qr(
+			Matrix< D, GeneralType, Dense > &H,
+			Matrix< D, OrthogonalType, Dense > &Q,
+			Matrix< D, GeneralType, Dense > &R,
```
vdimiczrc, line 60: If this algorithm is called on matrix views, the compiler would throw no matching call error because the default view type does not match the value used in matrix view. Therefore, View, ImfR and ImfC should be template parameters.
Additionally, since the algorithms can be compiled for different backends, the backend should also be a template parameter.

include/alp/algorithms/householder_qr.hpp
```diff
@@ -0,0 +1,244 @@
...
+
+#include <iostream>
+#include <sstream>
+
+#include <alp.hpp>
+#include <graphblas/utils/iscomplex.hpp> // use from grb
+#include "../tests/utils/print_alp_containers.hpp"
+// TEMPDISABLE enables workarounds for non-implemented features
+// should be removed in the final version
+#define TEMPDISABLE
```
vdimiczrc, line 25: Is this still used? If not, it can be removed.

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+#include <iomanip>
+#endif
+
+#include <alp.hpp>
+#include <alp/algorithms/householder_qr.hpp>
+#include <graphblas/utils/iscomplex.hpp> // use from grb
+#include "../utils/print_alp_containers.hpp"
+
+//once TEMPDISABLE is removed the code should be in the final version
+#define TEMPDISABLE
```
vdimiczrc, line 33: Is this necessary?

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+std::vector< T > generate_rectangular_matrix_data(
+	size_t N,
+	size_t M,
+	const typename std::enable_if<
+		grb::utils::is_complex< T >::value,
+		void
+	>::type * const = nullptr
+) {
+	std::vector< T > data( N * M );
+	std::srand( RNDSEED );
```
vdimiczrc, line 63: Due to this line, the function `generate_rectangular_matrix_data` will always produce the same data, no matter how many times it is called in an application. While in this one it is called only once, the side-effect should be noted for future use.
Maybe it could be good to move the call to `srand` outside, so it is called only once (again - not relevant for this particular test, but rather for potential future expansion).

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+	typename D,
+	typename StructureGen,
+	typename StructureOrth,
+	class Minus = operators::subtract< D >,
+	class Ring = Semiring< operators::add< D >, operators::mul< D >, identities::zero, identities::one >
+>
+RC check_solution(
+	alp::Matrix< D, StructureGen, alp::Density::Dense > &H,
+	alp::Matrix< D, StructureOrth, alp::Density::Dense > &Q,
+	alp::Matrix< D, StructureGen, alp::Density::Dense > &R,
```
vdimiczrc, line 157: Similar comment as above: View, ImfR, ImfC and backend should be template parameters.

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+		std::cout << "The Frobenius norm is too large.\n";
+		return FAILED;
+	}
+
+	return rc;
+}
+
+
+
+void alp_program( const size_t & unit, alp::RC & rc ) {
```
vdimiczrc, line 211: no spaces between & and variable

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+	alp::Semiring<
+		alp::operators::add< ScalarType >,
+		alp::operators::mul< ScalarType >,
+		alp::identities::zero,
+		alp::identities::one
+	> ring;
+
+	// dimensions of sqare matrices H, Q and R
+	size_t N = unit;
+	size_t M = 2 * unit;
```
vdimiczrc, line 223: Can these be const?

tests/smoke/alp_zgeqrf.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+//** check if rows/columns or matrix Q are orthogonal */
+template<
+	typename T,
+	typename Structure,
+	typename ViewType,
+	class Ring = Semiring< operators::add< T >, operators::mul< T >, identities::zero, identities::one >
+>
+RC check_overlap( alp::Matrix< T, Structure, alp::Density::Dense, ViewType > &Q, const Ring & ring = Ring() ) {
+	// more elegent would be check mxm(Q,conjugate(transpose(Q))) == identity
```
hidanielesgit, line 104: elegent -> elegant

hidanielesgit (approved):

include/alp/algorithms/householder_qr.hpp
djelovina: I'm not sure if structure::Orthogonal is the most general structure we can have. In principle we can non-full-rank orthogonal matrix here, something we need to discuss. I'll skip the static_assert at the moment if that is OK @hidanielesgit  ?

tests/smoke/alp_zgeqrf.cpp
djelovina: There is no much sense to call this algorithm on an integer type.

tests/smoke/alp_zgeqrf.cpp
djelovina: to call deconstructor for `std::vector matrix_data`    :)

include/alp/algorithms/householder_qr.hpp
```diff
@@ -78,31 +83,31 @@ namespace alp {
...
 			rc = rc ? rc : alp::set( Qdiag, one );
 			if( rc != SUCCESS ) {
-				std::cerr << " set( Q, I ) failed\n";
+				std::cerr << " alp::set( Q, I ) failed\n";
 				return rc;
 			}
 
 			// Out of place specification of the computation
-			Matrix< D, GeneralType, Dense > RR( n, m );
+			Matrix< D, GeneralType, alp::Dense, GenView, GenImfR, GenImfC > RR( n, m );
```
vdimiczrc, line 91: GenView, ImfR and ImfC should be used in input parameters to support passing matrix views to the function. For any internal variable within the function these should not be used. In summary, just remove them here.

include/alp/algorithms/householder_qr.hpp
```diff
@@ -78,31 +83,31 @@ namespace alp {
...
 				return rc;
 			}
 #ifdef DEBUG
 			print_matrix( " << RR >> ", RR );
 #endif
 
-			// a temporary for storing the mxm result
-			Matrix< D, OrthogonalType, Dense > Qtmp( n, n );
+			// a temporary for storing the alp::mxm result
+			Matrix< D, OrthogonalType, alp::Dense, OrthogonalView, OrthogonalImfR, OrthogonalImfC > Qtmp( n, n );
```
vdimiczrc, line 103: Same comment about View, ImfR and ImfC as above.

include/alp/algorithms/householder_qr.hpp
```diff
@@ -113,47 +118,47 @@ namespace alp {
 				// v = v - alpha * e1
 				// v = v / norm ( v )
 
-				auto v_view = get_view( RR, utils::range( k, n ), k );
-				Vector< D, GeneralType, Dense > v( n - k );
-				rc = set( v, v_view );
+				auto v_view = alp::get_view( RR, utils::range( k, n ), k );
+				Vector< D, GeneralType, alp::Dense, GenView, GenImfR, GenImfC > v( n - k );
```
vdimiczrc, line 122: Same comment about View, ImfR and ImfC as above.

include/alp/algorithms/householder_qr.hpp
```diff
@@ -195,17 +200,17 @@ namespace alp {
 				// RR = Qk * RR
 
 				// QkRR = Qk * RR
-				Matrix< D, GeneralType, Dense > QkRR( n, m );
-				rc = rc ? rc : set( QkRR, zero );
-				rc = rc ? rc : mxm( QkRR, Qk, RR, ring );
+				Matrix< D, GeneralType, alp::Dense, GenView, GenImfR, GenImfC > QkRR( n, m );
```
vdimiczrc, line 203: Same comment about View, ImfR and ImfC as above.

vdimiczrc (approved):

PR #99, opened 2022-10-19
Resolve internal issue 564: bug in fold from vector to scalar fullloopsparse

anyzelman:
The O(n) variant of folding a vector into a scalar that should deal with both sparsity of the vector as well as masks (the so-called fullLoopSparse variant), was in fact not checking for the sparsity of the input vector. All unit tests that could detect this issue were (apparently) instead triggering other variants (i.e., the maskDriven, the vectorDriven, or the dense variant)-- even though unit tests did exist for which the fullLoopSparse variant was triggered, those did not expose the bug detected here.

This MR fixes the bug, and adds a test case that reliably triggers the full sparse loop variant, by using a structural and inverted mask while assuring the sparsity pattern of the vector does not match that of the given (inverted) mask, which thus would also reliably trigger the detected bug. The MR also includes a minor code style fix and improved `_DEBUG` tracing.

Thanks to Aristeidis for detecting the bug!

### DIFF
--- a/include/graphblas/reference/blas1.hpp
+++ b/include/graphblas/reference/blas1.hpp
@@ -374,6 +374,9 @@ namespace grb {
 			const Vector< MaskType, reference, Coords > &mask,
 			const Monoid &monoid
 		) {
+#ifdef _DEBUG
+			std::cout << "Entered fold_from_vector_to_scalar_fullLoopSparse\n";
+#endif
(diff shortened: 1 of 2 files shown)


anyzelman:
Code review and GitHub CI OK. Waiting internal CI OK, as well as OK from smoke tests with LPF before merge.

anyzelman:
All OK, merging

PR #100, opened 2022-10-19
556 output verification in tests does not support integers

anhansson:


tests/utils/output_verification.hpp
```diff
@@ -99,35 +101,35 @@ int data_fscanf( const fileType in, std::complex< T > * const out ) {
...
+ * @returns  7 if all of 4, 2, and 1 apply
+ * @returns  8 if the computation of the 2-norm failed
+ * @returns  9 if both 8 and 1 apply
+ * @returns 10 if both 8 and 2 apply
+ * @returns 11 if all of 8, 2, and 1 apply
+ * @returns 12 if both 8 and 4 apply
+ * @returns 13 if all of 8, 4, and 1 apply
+ * @returns 14 if all of 8, 4, and 2 apply
+ * @returns 15 if all of 8, 4, 2, and 1 apply
+ * @returns 16 if 2-norm verification failed
```
anyzelman, line 119: Should all these not also throw errors, if for some others we now also throw?

tests/utils/output_verification.hpp
```diff
@@ -141,36 +143,37 @@ int vector_verification(
 	const constexpr T one = static_cast< T >( 1 );
 
 	// open verification file
-	FILE * const in = fopen( truth_filename, "r" );
+	std::ifstream in;
+	in.open( truth_filename);
```
anyzelman, line 147: missing space before `)`

tests/utils/output_verification.hpp
```diff
@@ -141,36 +143,37 @@ int vector_verification(
...
 	}
 
 	for( size_t i = 0; i < n; i++ ) {
 		const int rc = data_fscanf( in, truth + i );
 		if( rc != 0 ) {
-			std::cerr << "The verification file looks incomplete. " << "Line i = " << i
+			std::stringstream error;
+			error << "The verification file looks incomplete. " << "Line i = " << i
 				<< ", data = " << truth[ i ] << ", rc = " << rc << std::endl;
 			delete [] truth;
```
anyzelman, line 170: Canonically, should this `delete []` not go just after a try-block?

tests/utils/output_verification.hpp
```diff
@@ -191,7 +194,7 @@ int vector_verification(
 	if( raw_output_vector == nullptr || written_to == nullptr ) {
 		std::cerr << "Could not allocate necessary buffers" << std::endl;
 		delete [] truth;
```
anyzelman, line 196: same here

tests/utils/output_verification.hpp
```diff
@@ -71,14 +74,13 @@ int data_fscanf( const fileType in, double * const out ) {
...
-	const int rc = fscanf( in, "%lf%lf", &x, &y );
-	if( rc == 2 ) {
+template< typename T >
+int data_fscanf( std::ifstream& in, std::complex< T > * const out ) {
+	T x, y;
+	if(in >> x >> y){
 		*out = std::complex< T >( x, y );
 		return 0;
-	} else {
+	}else{
```
anyzelman, line 83: the previous spacing was correct

tests/utils/output_verification.hpp
```diff
@@ -30,37 +30,40 @@
...
-template< typename fileType >
-int data_fscanf( const fileType in, double * const out ) {
-	const int rc = fscanf( in, "%lf", out );
-	if( rc == 1 ) {
-		return 0;
-	} else {
-		return 1;
-	}
+template< typename T >
+int data_fscanf( std::ifstream& in, T * const out ) {
```
anyzelman, line 58: should be `std::ifstream &in`

tests/utils/output_verification.hpp
```diff
@@ -30,37 +30,40 @@
...
- * Attempts to read in a double value from a given file into a given memory
+ * Attempts to read in a value from a given file into a given memory
  * location.
+ * 
+ * @tparam T 		The datatype of the value
  *
- * @param[in]  in  The input file
- * @param[out] out Where to store the read value.
+ * @param[in]  in  	The input file
+ * @param[out] out 	Where to store the read value.
```
anyzelman, line 49: if these are tabs (after in and out), then they should be removed or replaced with spaces

anyzelman (changes requested):
Also could you elaborate if this was required by any upcoming MR perhaps? (This MR does not seem to use the new ability of supporting integers)

tests/utils/output_verification.hpp
anhansson: Not sure what you're implying since there's no exception thrown around here

anyzelman:
Oh right that is a very good point – we would need some sort of verification_error exception that could be thrown, and then caught at the caller site... that sounds like overengineering indeed. Then also the other comment that looks to have caused confusion would fall away. Will likely merge this as-is over the weekend then, many thanks!!

发件人: anhansson ***@***.***>
发送时间: Thursday, December 8, 2022 10:44 PM
收件人: Algebraic-Programming/ALP ***@***.***>
抄送: Albert Jan YZELMAN ***@***.***>; Review requested ***@***.***>
主题: Re: [Algebraic-Programming/ALP] 556 output verification in tests does not support integers (PR #100)


@anhansson commented on this pull request.

________________________________

In tests/utils/output_verification.hpp<https://github.com/Algebraic-Programming/ALP/pull/100#discussion_r1043433965>:

> + * @returns  1 if the \a output_vector was not dense

+ * @returns  2 if inf-norm verification failed

+ * @returns  3 if both 1 and 2 apply

+ * @returns  4 if the computation of the inf-norm failed

+ * @returns  5 if both 4 and 1 apply

+ * @returns  6 if both 4 and 2 apply

+ * @returns  7 if all of 4, 2, and 1 apply

+ * @returns  8 if the computation of the 2-norm failed

+ * @returns  9 if both 8 and 1 apply

+ * @returns 10 if both 8 and 2 apply

+ * @returns 11 if all of 8, 2, and 1 apply

+ * @returns 12 if both 8 and 4 apply

+ * @returns 13 if all of 8, 4, and 1 apply

+ * @returns 14 if all of 8, 4, and 2 apply

+ * @returns 15 if all of 8, 4, 2, and 1 apply

+ * @returns 16 if 2-norm verification failed

Not necessarily since these values represent a normal failure of the verification. This is in contrast to bad_alloc and the other errors thrown which are actual exceptions in the execution, which means the verification process could not occur at all.

tests/utils/output_verification.hpp
anyzelman: Just to clarify: if all errors were to be exceptions, then this block should throw. We could then throw before deleting truth, and handle memory reclaiming by catching all errors after allocating truth, deleting truth, and then rethrow. But as per the earlier remark, not everything should have to be an exception so the current pattern is fine=)

anyzelman:
Waiting internal CI and unit + smoke tests with LPF OK.

Waiting additional code review -- not sure how it happened, but some previous remarks that were resolved still appear in the latest code. I will do a pass and fix them, and if the above tests OK, will then merge.

anyzelman (approved):
Review OK

anyzelman:
Waiting smoke tests with LPF, waiting internal CI before merge

anyzelman:
Meanwhile, concept release notes:

In preparation for upcoming merges, this MR enables the output verification helper utility on integer data types; previously, only `double` data types were supported. The MR also removes some C style programming from the C++ code.

anyzelman:
Smoke tests with LPF (101 tests) OK, waiting internal CI

anyzelman:
Internal CI OK; merging

PR #101, opened 2022-10-19
507 copy constructor of pattern matrices does not compile by default

anhansson:
In compressed_storage.hpp there was a function copyFrom() specially made for pattern matrices. This has a few template parameters, one is use_id. In the other copyFrom() functions this has a default value, but for the pattern matrix it does not.

The fix for the compilation error was to add the default value for use_id in function copyFrom() for pattern matrices

### DIFF
--- a/include/graphblas/reference/compressed_storage.hpp
+++ b/include/graphblas/reference/compressed_storage.hpp
@@ -1175,12 +1175,19 @@ namespace grb {
 				/**
 				 * \internal copyFrom specialisation for pattern matrices.
 				 */
-				template< bool use_id = false, typename InputType, typename UnusedType = void >
+				template<
+					bool use_id = false,
+					typename InputType,
+					typename UnusedType = void
+				>
 				void copyFrom(
 					const Compressed_Storage< InputType, IND, SIZE > &other,
 					const size_t nz, const size_t m, const size_t start, size_t end,
 					const UnusedType * __restrict__ = nullptr
 				) {
+					// the use_id template is meaningless in the case of pattern matrices, but
+					// is retained to keep the API the same as with the non-pattern case.
+					(void) use_id;
 #ifdef _DEBUG
 					std::cout << "CompressedStorage::copyFrom (void) called with range "
 						<< start << "--" << end << "\n";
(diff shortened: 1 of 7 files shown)


anyzelman:
RFC: add unit test (or add to an existing unit test) that would have triggered the bug. (Also requires rebase.)

anyzelman:
@anhansson -- also here GitHub now reports a rebase is no longer possible due to conflict. Could you have a look?

Also could you indicate what the bug was / write a concept MR note?

anyzelman:
Concept release notes based on Anders' description:

Issue: copy-construction of pattern matrices did not compile.

Cause: In `compressed_storage.hpp` there was a function `copyFrom` specially made for pattern matrices. This has a few template parameters, one is `use_id`. In the other `copyFrom` functions (i.e., the non-pattern specialisation) this has a default value, but for the pattern matrix it does not.

This MR fixes the original compilation error by adding the default value `false` for `use_id` in `copyFrom` for pattern matrices. This template argument is retained so that callees can always assume the same API, regardless of whether pattern matrices are used or not. Furthermore, a unit test that copies pattern matrices into one another has been added, using both `grb::set` and the copy constructor.

This added test then triggered a bug where matrices with initial requested capacity zero would not have their offset arrays initialised. Later use (even after valid and successful resizing) would then trigger segfaults. This MR also fixes this bug.

A separate metabug concerns the tracking of freed memory, which (if the ALP/GraphBLAS was configured to emit this information) would result in the use of uninitialised memory when reporting on how many bytes were freed for a matrix with zero capacity. This MR also fixes this bug.

The MR also includes code style fixes, also in one unrelated unit test.

anyzelman (approved):
Review OK (after adding a unit test for the issue this MR fixes, and after resolving two other bugs that the new unit test triggered)

anyzelman:
Waiting internal CI OK, waiting unit tests with LPF OK before merging.

anyzelman (changes requested):
Actually -- the commit history on this one looks a bit funky. @anhansson might you have an idea why 419 and 565 appear in the commit history, was there an earlier merge with these two branches perhaps? (Is it OK if we just rebase and delete these commits from the 507 branch?)

PR #102, opened 2022-10-19
486 add prefetching to gather style spmv

anhansson:


anyzelman:
@anhansson Hi Anders sorry, could I bother you to rebase on top of the current develop? (Conflicts have appeared in the mean while)

include/graphblas/reference/blas2.hpp
```diff
@@ -409,13 +412,34 @@ namespace grb {
 				}
 				// check for sparsity at source
 				if( !dense_hint ) {
-					if( !internal::getCoordinates( source_vector ).assigned( source_index ) ) {
+					if( config::PREFETCHING< reference >::enabled() ) {
+						size_t dist = k + 2 * config::PREFETCHING< reference >::distance();
+						if( dist < nz ) {
+							const size_t prefetch_target_assigned = matrix.row_index[ dist ];
+							src_coordinates.prefetch_assigned( prefetch_target_assigned );
```
anyzelman, line 419: @anhansson -- this is the branch that didn't do much correct, and could be disabled? Just to check before I push a change that disables it

anyzelman: Hope you had awesome holidays!! I looked at it further and this is actually not a straight index prefetch-- it's not immediately clear why this branch should not be effective. The distance of course can be off with some machine- and data-dependent factor, even when relative to the data prefetch, but that's an inherent problem with software prefetching. I'll hence merge it as-is, though will likely add some documentation before doing so~

anyzelman (approved):

anyzelman:
Concept release notes:

This MR introduces basic prefetching capabilities to ALP/GraphBLAS. The new capabilities, however, by default are turned off as a thorough test on all matrices and graphs in our dataset did not identify any safe settings on any one of two architectures. Here, "safe" means that no significant performance degradation would occur for any one kernel on any one matrix or graph, out of all kernels and inputs tested. This indicates that prefetching should only be used for specific use cases, preferably while tuning the prefetch distance for that particular use case.

Whether prefetching is enabled and at which distance is controlled by two respective settings in `<root>/include/graphblas/reference/config.hpp`. The README is updated to describe these (and other) config settings.

This MR include code style fixes across all files touched, as well as in particular code style fixes across the various `config.hpp` files while composing the corresponding new README section. Additionally, the references at the bottom of the README file have been updated.

anyzelman:
Review OK, GitHub CI OK; waiting internal CI, waiting smoke tests with LPF

PR #104, opened 2022-10-19
565 k nn and mpv algorithms gives wrong results

anhansson:
Anders detected an off-by-one error in the mpv (matrix powers) kernel / algorithm, which this MR fixes. This affects the k-NN algorithms as well as the smoke and performance tests based on it, which were previously verified against the sequential ALP/GraphBLAS baseline. This MR also hardens these tests by comparing against a ground truth verified by an external framework, in this case Octave.
This MR has been thoroughly tested using the full test suite. Additionally, an issue (GitHub #136) has been raised to consider an mpv implementation which is not unrolled and simply based on `std::swap`-- suspected is that the current implementation predates supporting swap, and, different from an implementation using `grb::set` instead of swap, a major performance loss compared to an unrolled algorithm no longer is expected.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -29,6 +29,8 @@ to Huawei Technologies Co., Ltd. or one of its subsidiaries:
 
  - Auke Booij, Huawei Technologies Switzerland AG; 2021.
 
+ - Anders Hansson, Huawei Technologies Switzerland AG; 2022-2023.
+
 The experimental banshee backend has been developed in collaboration with
 Prof. Luca Benini at ETH Zuerich and his group. In particular this backend
 is with great thanks due to Dan, Paul Scheffler, Fabian Schuiki, and Samuel
--- a/include/graphblas/algorithms/mpv.hpp
+++ b/include/graphblas/algorithms/mpv.hpp
@@ -144,7 +144,7 @@ namespace grb {
 				ret = mxv< descr >( temp, A, u, ring );
 				// check if this was the final multiplication
 				assert( iterate <= k );
-				if( iterate == k || ret != SUCCESS ) {
+				if( iterate + 1 == k || ret != SUCCESS ) {
 					break;
 				}
 				// multiply with output into u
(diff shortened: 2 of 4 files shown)


anyzelman (dismissed):
OK (though I believe the implementation could be shortened to a non-unrolled for-loop-- this implementation probably predated `std::swap`)

anyzelman:
Waiting smoke tests with LPF OK, waiting CIs OK, waiting open issue for simplifying mpv implementation.

anyzelman:
Rebased. Code review OK. New issue for improving mpv created. Waiting smoke tests with LPF OK, waiting CIs OK.

anyzelman:
RFC made (@anhansson) re automated tests against a trusted external baseline

anyzelman (approved):
New ground truth generated by Octave. Review OK.

anyzelman:
Many thanks Anders for the update. Review OK. GitHub CI OK.

Waiting internal CI OK, perform full test suite run with LPF before merging

anyzelman (approved):
com-orkut 4-NN size confirmed using Octave, re-approve MR

anyzelman:
Full test suite run with LPF found a checksum error for com-orkut. Now reconfirmed using Octave -- indeed it was previously missed. Finishing full test suite to make sure this was the only error, and re-running internal CI. Will merge if both OK.

anyzelman:
Internal CI failed and in so doing implicitly detected that this MR was no longer based on the latest develop. My bad. Restarting internal CI check.

anyzelman (approved):
Approve after a non-code change (attribution/NOTICE)

anyzelman:
> Internal CI failed and in so doing implicitly detected that this MR was no longer based on the latest develop. My bad. Restarting internal CI check.

All is now OK. Found missing attribution to @anhansson in NOTICE file -- added it, and will now proceed to write merge comments before merging.

PR #107, opened 2022-10-20
106 internal ci picks up assert failure in hyperdags backend

anyzelman:
The internal CI builds unit and smoke tests in the `develop` branch (also) using the debug mode (in addition to the standard release mode), which after a totally unrelated commit failed for unit tests that call `eWiseApply` on matrices, for the hyperdags backend only.

The issue was that the output matrix was not correctly registered as a source, resulting in containers being "lost" in the internal hyperdag representation-- for which `src/hyperdags/hyperdag.cpp` has an assertion check. Indeed, assertion checks in sources are only active when compiled in Debug mode, though it remains unclear why those unit tests did not fail before.


anyzelman:
Review OK. GitHub CI OK. Manual test on `unit/eWiseApply_matrix_debug_hyperdags` with ALP/GraphBLAS compiled in Debug mode OK. Waiting internal CI OK before merge.

anyzelman:
Internal CI OK, will merge.

PR #113, opened 2022-10-25
560 alp reference sort primitive

hidanielesgit:
PR for introducing iterator-based (vector) sort primitive. Its use in existing algorithms will follow with a separate PR.


include/alp/amf-based/vector.hpp
```diff
@@ -58,6 +58,15 @@ namespace alp {
 		void setInitialized( alp::Vector< T, Structure, Density::Dense, View, ImfR, ImfC, backend > & v, bool initialized ) noexcept {
 			setInitialized( static_cast< typename alp::Vector< T, Structure, Density::Dense, View, ImfR, ImfC, backend >::base_type &>( v ), initialized );
 		}
+
+		template< typename T, typename Structure, typename View, typename ImfR, typename ImfC, enum Backend backend >
```
vdimiczrc, line 62: Template specifications should be at most 80 lines long, can you please break this into lines. Also applies to other locations.

include/alp/amf-based/vector.hpp
```diff
@@ -58,6 +58,15 @@ namespace alp {
...
+		typename alp::Vector< T, Structure, Density::Dense, View, ImfR, ImfC, backend >::iterator
+		begin( alp::Vector< T, Structure, Density::Dense, View, ImfR, ImfC, backend > & v ) noexcept;
```
vdimiczrc, line 64: no space between & and variable name.

include/alp/amf-based/vector.hpp
```diff
@@ -110,6 +119,152 @@ namespace alp {
...
+			class VectorIterator: 
+				public std::iterator< std::random_access_iterator_tag, T > {
+				
+				friend class Vector< T, structures::General, Density::Dense, View, ImfR, ImfC, backend >;
+
+				private:
+					
+					typedef typename self_type::storage_index_type index_type;
+
+					self_type * vec;
```
vdimiczrc, line 131: no space after *

include/alp/amf-based/vector.hpp
```diff
@@ -110,6 +119,152 @@ namespace alp {
...
+					index_type position;
+
+					VectorIterator( self_type * vptr ) noexcept :
```
vdimiczrc, line 134: no space after *

include/alp/amf-based/vector.hpp
```diff
@@ -110,6 +119,152 @@ namespace alp {
...
+						++position;
+						return *this;
+					}
+
+					VectorIterator& operator--() {
+						--position;
+						return *this;
+					}
+
+					VectorIterator operator++(int) {
```
vdimiczrc, line 204: missing spaces within parentheses. Also applies below in several ocasions.

include/alp/reference/blas1.hpp
```diff
@@ -2847,28 +2847,85 @@ namespace alp {
...
 	template<
-		typename IndexType, typename IndexStructure, typename IndexView, typename IndexImfR, typename IndexImfC,
+		typename IndexStructure, typename IndexView, typename IndexImfR, typename IndexImfC,
+		typename ValueType, typename ValueStructure, typename ValueView, typename ValueImfR, typename ValueImfC
+	>
+	RC sort(
+		Vector< size_t, IndexStructure, Density::Dense, IndexView, IndexImfR, IndexImfC, reference > &permutation,
+		const Vector< ValueType, ValueStructure, Density::Dense, ValueView, ValueImfR, ValueImfC, reference > &toSort
+	) noexcept {
+
```
vdimiczrc, line 2869: Should this function check if the permutation vector and the data vector are of the same size?

tests/unit/dense_sort.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    auto print_std_vector = [](std::vector< T > const vec) {
+        for (auto val : vec) {
+            std::cout << val << ' ';
+        }
+        std::cout << std::endl;
+    };
+
+	// Check with vector of length n randomly intitialized and shuffled
+	alp::Vector< size_t > perm( n );
```
vdimiczrc, line 42: It seems like the indentation is not consistent within `alp_program`.

tests/unit/dense_sort.cpp
```diff
@@ -0,0 +1,154 @@
...
+		if ( stdv[i] != sorted_v[ i ] ) {
+			std::cerr << "Error: ( std::v[ " << i << " ] = " << stdv[i] << " ) != " << " ( sorted_v[ " << i << " ] = " << sorted_v[ i ] << " )" << std::endl;
+			rc = alp::FAILED;
+		}
+	}
+
+    std::cout << "Sorted alp::Vector:" << std::endl;
+	print_vector("sorted_v", sorted_v);
+
+	if (rc == alp::FAILED ) {
```
vdimiczrc, line 77: Spaces

tests/unit/dense_sort.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+	alp::sort( perm, v );
+
+	std::sort( std::begin( stdv ), std::end( stdv ) );
+
+	auto sorted_v = alp::get_view< alp::structures::General >( v, perm );
+
+	// Check sorted view
+	for ( size_t i = 0; i < n; i++ ) {
+		if ( stdv[i] != sorted_v[ i ] ) {
```
vdimiczrc, line 68: No spaces needed after for and if.

tests/unit/dense_sort.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+	std::iota( std::begin( stdv ), std::end( stdv ), 0. );
+	std::shuffle( std::begin( stdv ), std::end( stdv ), rng );
+
+	alp::buildVector( v, std::begin( stdv ), std::end( stdv ) );
+
+    std::cout << "Original content of the std::vector:" << std::endl;
+	print_std_vector( stdv );
+    std::cout << "Original content of the alp::Vector:" << std::endl;
+	print_vector("v", v);
```
vdimiczrc, line 58: missing spaces

include/alp/amf-based/vector.hpp
```diff
@@ -110,6 +119,152 @@ namespace alp {
 			typedef Matrix< T, structures::General, Density::Dense, View, ImfR, ImfC, backend > base_type;
 
 		private:
+			class VectorIterator: 
+				public std::iterator< std::random_access_iterator_tag, T > {
+				
+				friend class Vector< T, structures::General, Density::Dense, View, ImfR, ImfC, backend >;
```
vdimiczrc, line 125: Does ALP Vector need to be a friend of the iterator?

include/alp/reference/blas1.hpp
```diff
@@ -2847,28 +2847,85 @@ namespace alp {
 	}
 
 	/**
-	 * Sort vectors, function available to user, e.g. to sort eigenvectors
+	 * Compute the permutation vector needed to sort the input vector according to
+	 * the provided \a cmp function.
```
vdimiczrc, line 2851: It seems that this function uses a hard-coded comparison function - it is not provided by the caller.

vdimiczrc (approved):

include/alp/reference/blas1.hpp
```diff
@@ -2847,28 +2847,85 @@ namespace alp {
...
-		typename Compare
+		typename CompareType
 	>
 	RC sort(
-		Vector< IndexType, IndexStructure, Density::Dense, IndexView, IndexImfR, IndexImfC, reference > &permutation,
+		Vector< size_t, IndexStructure, Density::Dense, IndexView, IndexImfR, IndexImfC, reference > &permutation,
 		const Vector< ValueType, ValueStructure, Density::Dense, ValueView, ValueImfR, ValueImfC, reference > &toSort,
-		Compare cmp
-		//PHASE &phase = EXECUTE
+		CompareType cmp
```
djelovina, line 2903: can there be one version of `sort` with default `cmp = less_then` ?

tests/unit/dense_sort.cpp
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+    std::cout << "Sorted alp::Vector in descending order:" << std::endl;
+	print_vector("desc_sorted_v", desc_sorted_v);
+
+
+	rc = alp::SUCCESS;
+}
+
+int main( int argc, char ** argv ) {
```
djelovina, line 107: `**argv`

djelovina (approved):

include/alp/amf-based/vector.hpp
hidanielesgit: Yes to allow access to private constructors

include/alp/reference/blas1.hpp
hidanielesgit: This will be addressed soon by allowing a single variant of the function which always expects a poset/total order.

include/alp/reference/blas1.hpp
hidanielesgit: This will be addressed soon by allowing a single variant of the function which always expects a poset/total order.

hidanielesgit:
Based on previous discussion I've added a first draft of an ALP relation concept including few basic cases directly usable with alp::sort. Possible items of discussion/review: 
- In this draft I've separated the concept of operator from the one of relation, although relations could be seen as operators $SET \times SET \rightarrow bool$ (or operators as ternary relations). On the one hand this allowed to play with properties which are not typical of operators such as reflexivity, etc.On the other hand this creates a bit of redundancy with existing `operators::equal`, and few others already possibly in use in an operator context.
- We could also add relations such complement, union, and intersection which would allow building more complex relations. While checking would be straightforward (they would "negate", "and", and "or", respectively their source relations) the effect of them on the resulting relation's properties requires a bit of thought.

include/alp/base/internalrels.hpp
```diff
@@ -0,0 +1,796 @@
...
+				 * casting is required. This version will be automatically called whenever
+				 * possible.
+				 */
+				static bool check( const D1 & x, const D2 & y ) {
+					return REL::check( &x, &y );
+				}
+			};
+
+			/**
+			 * TODO: Update for Relation
```
anyzelman, line 638: Use `\TODO` or `\todo` so that it ends up in the doxy gen also.

Maybe just delete the old comments?

include/alp/base/internalrels.hpp
```diff
@@ -0,0 +1,796 @@
...
+						typename REL::left_type, 
+						typename REL::right_type 
+					>::value 
+				> * = nullptr
+			>
+			class HomogeneousRelation : public Relation< REL, implementation > {
+			};
+
+		} // namespace internal
+
```
anyzelman, line 790: Are there no type traits for transitive etc.?

I think a single generic type trait would work, meaning a single one for each new algebraic type trait, as follows (psuedocode):

```
template< Relation >
class is_transitive {
    ...value = false;
}

template< Relation, bool = SFINAE< is_relation< Releation > >
class is_transitive {
   ...value = Relation::is_transitive;
};
```

anyzelman (changes requested):

include/alp/base/internalrels.hpp
hidanielesgit: Updated doxygen block.

include/alp/reference/blas1.hpp
```diff
@@ -2218,28 +2221,55 @@ namespace alp {
...
 	>
 	RC sort(
-		Vector< IndexType, IndexStructure, Density::Dense, IndexView, IndexImfR, IndexImfC, reference > &permutation,
+		Vector< size_t, IndexStructure, Density::Dense, IndexView, IndexImfR, IndexImfC, reference > &permutation,
 		const Vector< ValueType, ValueStructure, Density::Dense, ValueView, ValueImfR, ValueImfC, reference > &toSort,
-		Compare cmp
-		//PHASE &phase = EXECUTE
+		const Relation &rel = Relation(),
+		const std::enable_if_t<
+			! alp::is_object< ValueType >::value
```
djelovina, line 2245: I think the current standard is not to use space after `!`

include/alp/base/internalrels.hpp
```diff
@@ -0,0 +1,796 @@
...
+					const D2 b = static_cast< D2 >( y );
+					return REL::check( &a, &b );
+				}
+
+				/**
+				 * This is the high-performance version of check() in the sense that no
+				 * casting is required. This version will be automatically called whenever
+				 * possible.
+				 */
+				static bool check( const D1 & x, const D2 & y ) {
```
djelovina, line 632: no spaces between `&` and `x`, here and on the other places

tests/unit/alp_relations.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+	static_assert( alp::is_strict_partial_order< dbl_lt >::value );
+
+	static_assert( ! alp::is_total_order< dbl_lt >::value );
+
+	static_assert( alp::is_strict_total_order< dbl_lt >::value );
+
+	static_assert( ! alp::is_equivalence_relation< dbl_lt >::value );
+
+	if( ! dbl_lt::check(2.4, 5) ) {
```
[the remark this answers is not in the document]
djelovina: missing spaces in parenthesis, i.e. it should be  `( 2.4, 5 )`, here and on the other places

tests/unit/alp_relations.cpp
```diff
@@ -0,0 +1,345 @@
...
+#endif
+		rc = alp::FAILED;
+		return;
+	}
+
+	rc = alp::SUCCESS;
+
+}
+
+int main( int argc, char ** argv ) {
```
djelovina, line 316: no space after `**`

djelovina (changes requested):
Could you also update (if needed) https://github.com/Algebraic-Programming/ALP/blob/7315f18928c06e4ded3fe73da9f3234c0db1b844/include/alp/algorithms/symm_tridiag_eigensolver.hpp#L487 in this branch?

include/alp/base/internalrels.hpp
hidanielesgit: yes good point - added.

include/alp/reference/blas1.hpp
```diff
@@ -2218,28 +2221,55 @@ namespace alp {
...
 		typename ValueType, typename ValueStructure, typename ValueView, typename ValueImfR, typename ValueImfC,
-		typename Compare
+		typename Relation = relations::lt< ValueType, reference >
 	>
 	RC sort(
-		Vector< IndexType, IndexStructure, Density::Dense, IndexView, IndexImfR, IndexImfC, reference > &permutation,
+		Vector< size_t, IndexStructure, Density::Dense, IndexView, IndexImfR, IndexImfC, reference > &permutation,
 		const Vector< ValueType, ValueStructure, Density::Dense, ValueView, ValueImfR, ValueImfC, reference > &toSort,
-		Compare cmp
-		//PHASE &phase = EXECUTE
```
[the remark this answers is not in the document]
hidanielesgit: Yes Phases are currently not considered in ALP/Dense. But are often commented out as we might want to reason about their role in dense.

include/alp/type_traits.hpp
```diff
@@ -118,6 +118,17 @@ namespace alp {
...
+	 * Used to inspect whether a given type is an ALP relation.
+	 *
+	 * @tparam T The type to inspect.
+	 */
+	template< typename T >
+	struct is_relation {
+		/** Base case: an arbitrary type is not a relation. */
+		static const constexpr bool value = false;
+	};
+
```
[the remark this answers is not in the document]
hidanielesgit: Added generic type traits in internalrels.hpp which rely on default `is_relation`.

tests/unit/dense_sort.cpp
```diff
@@ -0,0 +1,171 @@
...
+	alp::buildVector( v, std::begin( stdv ), std::end( stdv ) );
+
+#ifndef NDEBUG
+	std::cout << "Original content of the std::vector:" << std::endl;
+	print_std_vector( stdv );
+	std::cout << "Original content of the alp::Vector:" << std::endl;
+	print_vector("v", v);
+#endif
+
+	alp::sort( perm, v );
```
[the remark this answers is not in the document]
hidanielesgit: Correct.

include/alp/reference/blas1.hpp
```diff
@@ -2218,28 +2221,55 @@ namespace alp {
...
 	) noexcept {
-		return SUCCESS;
+
+		internal::setInitialized( permutation, internal::getInitialized( toSort ) );
+
+		if( !internal::getInitialized( toSort ) ) {
+			return SUCCESS;
+		}
+
+		RC rc = alp::eWiseLambda(
```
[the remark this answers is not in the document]
hidanielesgit: Thanks for pointing this one out. I'll take a look.

include/alp/algorithms/symm_tridiag_eigensolver.hpp
```diff
@@ -484,24 +484,12 @@ namespace alp {
...
-			std::sort(
-				isort_dtmp.begin(),
-				isort_dtmp.end(),
-				[ &dtmp ]( const size_t &a, const size_t &b ) {
-					return ( dtmp[ a ] < dtmp[ b ] );
-				}
-			);
+			// permutation that sorts dtmp
 			alp::Vector< size_t > permutation_vec( n );
+			rc = rc ? rc : alp::sort(permutation_vec, dtmp, alp::relations::lt< D >() );
```
vdimiczrc, line 489: missing space before permutation_vec

tests/unit/alp_relations.cpp
```diff
@@ -0,0 +1,345 @@
...
+
+	static_assert( alp::is_strict_partial_order< dbl_lt >::value );
+
+	static_assert( !( alp::is_total_order< dbl_lt >::value ) );
+
+	static_assert( alp::is_strict_total_order< dbl_lt >::value );
+
+	static_assert( !( alp::is_equivalence_relation< dbl_lt >::value ) );
+
+	if( !( dbl_lt::check(2.4, 5) ) ) {
```
vdimiczrc, line 46: missing spaces within inner-most parentheses. Also in other similar appearances in this file.

include/alp/type_traits.hpp
```diff
@@ -118,6 +118,17 @@ namespace alp {
 		static const constexpr bool value = false;
 	};
 
+	/**
+	 * Used to inspect whether a given type is an ALP relation.
```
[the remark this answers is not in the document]
hidanielesgit: No problem with this. Should we adapt all surrounding docs as well? Just to keep the wording similar across the header.

djelovina:
> Could you also update (if needed)
> 
> https://github.com/Algebraic-Programming/ALP/blob/7315f18928c06e4ded3fe73da9f3234c0db1b844/include/alp/algorithms/symm_tridiag_eigensolver.hpp#L487
> in this branch?

I think it is ok to do it in this branch.

hidanielesgit:
> > * In this draft I've separated the concept of operator from the one of relation, although relations could be seen as operators
>
>One question here, btw: could a relation be an interpretation of said binary operator, thus potentially preventing duplication? (E.g., the less-than relation could "wrap" around the less-than operator?)

I think this could be a possibility for relations that have similar assumptions on the existence of their respective c++ operators. 
For example, for _lt_ the assumption would shift from 

_Assumes native availability of \a operator> on the given data types_

to something like

_Assumes availability of ALP operator \a less_than forming a strict total order on SET_

This way we could decide the best way to wrap around. Eg, the `lt<SET, impl>::check` method would rely on `less_than<SET, SET, bool, impl>::apply`. 

On the pros side, I think it would be a way to rely on a single apply logic when this exists. Still the dependence on the existence of an operator would not be strictly enforced, allowing if needed/desired the definition of relations not based on any particular operator (eg, I could decide to create a relation, perhaps with no properties at all, where I could literally enumerate a subset of $SET\times SET$).

Perhaps on the cons side, we would always have to allocate a variable for the boolean output `c`? Also differently from `OP::apply` `REL::check` does not impose a `restrict` assumption on the two input pointers since a relation should also be able to check if $aRa$.

Does this make sense?

tests/unit/dense_sort.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "../utils/print_alp_containers.hpp"
+
+using namespace alp;
+
+void alp_program( const size_t &n, alp::RC &rc ) {
+
+	typedef double T;
+
+#ifndef NDEBUG
+	auto print_std_vector = [](std::vector< T > const vec) {
```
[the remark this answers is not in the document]
hidanielesgit: Good point. Noted for tests on future parallel backends.

anyzelman:
> > > * In this draft I've separated the concept of operator from the one of relation, although relations could be seen as operators
> > 
> > 
> > One question here, btw: could a relation be an interpretation of said binary operator, thus potentially preventing duplication? (E.g., the less-than relation could "wrap" around the less-than operator?)
> 
> I think this could be a possibility for relations that have similar assumptions on the existence of their respective c++ operators. For example, for _lt_ the assumption would shift from
> 
> _Assumes native availability of \a operator> on the given data types_
> 
> to something like
> 
> _Assumes availability of ALP operator \a less_than forming a strict total order on SET_
> 
> This way we could decide the best way to wrap around. Eg, the `lt<SET, impl>::check` method would rely on `less_than<SET, SET, bool, impl>::apply`.
> 
> On the pros side, I think it would be a way to rely on a single apply logic when this exists. Still the dependence on the existence of an operator would not be strictly enforced, allowing if needed/desired the definition of relations not based on any particular operator (eg, I could decide to create a relation, perhaps with no properties at all, where I could literally enumerate a subset of SET×SET).
> 
> Perhaps on the cons side, we would always have to allocate a variable for the boolean output `c`? Also differently from `OP::apply`, `REL::check` does not impose a `restrict` assumption on the two input pointers since a relation should also be able to check if `a REL a`.
> 
> Does this make sense?

Yeah I think that makes sense. The other advantage is that all of the ALP operator mechanics get inherited -- auto-vectorisation (if ever applicable), but also operator overloads to specific backends (e.g., Banshee/Snitch).

The allocation of a boolean -- yes and no: a place for it is required, but it need not be dynamically allocated, it goes on the stack? Like `const bool check = false; apply( check, ..., ltOp );` At reasonable compiler optimisation levels it is likely to be equivalent even?

anyzelman (approved):
Ok for me, pending perhaps the dependence of relations on operators issue

hidanielesgit:
> The allocation of a boolean -- yes and no: a place for it is required, but it need not be dynamically allocated, it goes on the stack? Like `const bool check = false; apply( check, ..., ltOp );` At reasonable compiler optimisation levels it is likely to be equivalent even?

Correct we would use stack memory there. And I agree the compiler might optimize things anyway given that they are very short (inline-able) function calls. 

OK I think I'll try wrapping around existing operators (equal/not-equal if I am not mistaken) before merging into 303 and later if we are OK with it we could also add missing operators.

PR #118, opened 2022-10-27
579 dense forward substitution

djelovina:


include/alp/algorithms/forwardsubstitution.hpp
```diff
@@ -0,0 +1,148 @@
...
+#endif
+
+namespace alp {
+
+	namespace algorithms {
+
+		/**
+		 *        Solves linear system Ax=b
+		 *        where A is LowerTriangular matrix, b is given RHS vector
+		 *        and x is the solution.
```
vdimiczrc, line 33: We use one space to separate text from `*`.

tests/smoke/alp_forwardsubstitution.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+/** generate complex lower triangular positive definite matrix data */
+template<
+	typename T,
+	const typename std::enable_if<
+		grb::utils::is_complex< T >::value,
+		void
+	>::type * const = nullptr
+>
+std::vector< T > generate_lpd_matrix( size_t N  ) {
```
vdimiczrc, line 104: This looks like a good candidate for utils (maybe for future work).

include/alp/algorithms/forwardsubstitution.hpp
```diff
@@ -0,0 +1,148 @@
...
+ 				rc = rc ? rc : alp::foldl( alpha, A_ii, ring.getAdditiveMonoid() );
+ 				rc = rc ? rc : alp::foldl( x_i, alpha, divide );
+			}
+
+			return rc;
+		}
+
+		template<
+			typename D = double,
+			typename ViewA,
```
hidanielesgit, line 99: All these arguments don't seem (and correctly so) to be of use in the algorithm. Do we need to expose them? Otherwise type traits should be used to check alp-ness or other properties.

hidanielesgit (approved):

PR #124, opened 2022-11-03
406 Reuse GRB algebraic types as ALP ones

vdimiczrc:


djelovina (approved):

include/alp/base/internalops.hpp
```diff
@@ -31,3144 +31,198 @@
...
-				 * @param[in,out] c Pointer to the left-hand side input data. This also
-				 *                  dubs as the output memory area.
-				 * @param[in]     b Pointer to the right-hand side input data.
-				 */
-				static void foldl( result_type * __restrict__ const c, const right_type * __restrict__ const b ) {
-					(void)b;
-					(void)c;
-				}
-			};
+			using left_assign = grb::operators::internal::left_assign< IN1, IN2, OUT, grb::Backend::reference >;
```
hidanielesgit, line 67: Not sure if it is a good idea to stick to `grb::Backend::reference` at this point. what if we are in a non-reference scenario and the operator is also specialized in grb? This would not go for the right specialization right?
If this comment makes sense then should we have a thin mapping layer `alp::Backend::x` -> `grb::Backend::y` ?

Similar remark for a number of other occurrences in this file and other non-reference headers.

include/alp/ops.hpp
```diff
@@ -34,488 +35,102 @@ namespace alp {
...
 		template< typename D1, typename D2 = D1, typename D3 = D2, enum Backend implementation = config::default_backend >
-		class left_assign : public internal::Operator< internal::left_assign< D1, D2, D3, implementation > > {
-		public:
-			template< typename A, typename B, typename C, enum Backend D >
-			using GenericOperator = left_assign< A, B, C, D >;
-			left_assign() {}
-		};
-
-		/** TODO documentation. */
+		using left_assign = grb::operators::left_assign< D1, D2, D3, grb::Backend::reference >;
```
hidanielesgit, line 40: here and similar situations below - see question about backend in `base/internalops.hpp`.

PR #130, opened 2022-11-20
Resolves internal issue #593: dangling include in reference matrix

anyzelman:
Alberto detected an erroneous include in the reference `matrix.hpp`, and a missing include in the spy algorithm. This MR fixes both issues. Many thanks for both detecting the bug and for providing the fix!


anyzelman:
Review OK. Internal CI OK. GitHub CI OK. Will merge

PR #135, opened 2022-11-29
134 additional internal ci failures triggered for hyperdag in debug mode

anyzelman:
Internal CI detects errors in the hyperDAGs backend for the `eWiseApply` and `zip` unit tests, as well as in the `hpcg` smoke test, only when run in debug mode (without `-DNDEBUG`).

Issue: the debug mode detects disconnected hyperDAGs when they are dumped. This indicates a vertex from which there is no path to some output.

Cause: some variants of `eWiseApply` did not connect to the output container as an input to the `eWiseApply`. This sounds paradoxical since this primitive is out of place, but like for the `clear`, the container must still be registered as a source because its previous contents are invalidated. The same issue existed for `unzip`.

This MR fixes both issues, and fixes the code style of the `tests/unit/zip.cpp` unit test.

### DIFF
--- a/include/graphblas/hyperdags/blas1.hpp
+++ b/include/graphblas/hyperdags/blas1.hpp
@@ -166,7 +166,11 @@ namespace grb {
 		if( phase != EXECUTE ) { return ret; }
 		if( size( internal::getVector(in) ) == 0 ) { return ret; }
 		std::array< const void *, 0 > sourcesP{};
-		std::array< uintptr_t, 1 > sourcesC{ getID( internal::getVector(in) ) };
+		std::array< uintptr_t, 3 > sourcesC{
+			getID( internal::getVector(x) ),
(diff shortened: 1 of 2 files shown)


anyzelman:
Review OK. Smoke tests with LPF OK. Waiting CIs OK before merge.

anyzelman:
Smoke tests with LPF in debug mode OK. GitHub CI OK. Waiting internal CI OK.

anyzelman:
It remains unclear why these bugs did not trigger internal CI failures earlier-- this is the second instance that it is happened. One effect is that not all eWiseApply variants were tested two commits ago, but this is not the case for unzip, and neither did the expanded unit testing initially trigger the CI failure.

anyzelman:
Internal CI OK, will merge

PR #138, opened 2022-12-07
599 dense lu factorisation blocked

djelovina:


include/alp/structures.hpp
```diff
@@ -666,6 +674,14 @@ namespace alp {
 
 		};
 
+		template<>
+		struct isInstantiable< UpperTrapezoidal, UpperTrapezoidal > {
+			template< typename ImfR, typename ImfC >
+			static bool check( const ImfR &imf_r, const ImfC &imf_c ) {
+				return imf_r.map( 0 ) <= imf_c.map( 0 );
```
hidanielesgit, line 681: I think we should make sure that `imf_r.map( 0 ) <= imf_c.map( n - 1 )` ?

tests/utils/print_alp_containers.hpp
```diff
@@ -54,7 +54,7 @@ void print_matrix( std::string name, const MatrixType &A ) {
 					std::cout <<  ",";
 				}
 			} else {
-				std::cout << std::setprecision( 0 ) << "\t" << ".";
+				std::cout << std::setprecision( 0 ) << "\t" << "0, ";
```
hidanielesgit, line 57: Is this needed?

hidanielesgit (changes requested):

tests/utils/print_alp_containers.hpp
djelovina: Since print_matrix is normally used in debug, it seems to be more convenient to have exact values when you copy matrix to eg. python. If there is a good reason I can leave it as `.`.

include/alp/structures.hpp
djelovina: Is `n` here `imf_c.n`? What would this test? The current condition is checking that the diagonal is taken above the original diagonal.

tests/utils/print_alp_containers.hpp
hidanielesgit: good point. Maybe we could overload with a value of type `T` to print as zero (in which case it uses a numpy-like format) and if not provided a dot.

include/alp/structures.hpp
hidanielesgit: If not mistaken it would make sure you are not selecting a complete zero region.

tests/smoke/alp_zgetrf.cpp
```diff
@@ -170,29 +174,87 @@ RC check_lu_solution(
...
+	auto L_lowerTrapez = get_view( L, utils::range( 1, m ), utils::range( 0, kk ) );
+	rc = rc ? rc : alp::set( L_lowerTrapez, LU_Ltrapez );
+
+	rc = rc ? rc : check_lu_solution( H, L, U, p, ring );
 
-/** no pivoting version */
+	return rc;
+}
+
+/** check if LU decomposition is correct H==LU, in-place without pivoting version */
```
hidanielesgit, line 235: out-of-place?

hidanielesgit (approved):

PR #145, opened 2022-12-21
Fixes bug in HyperDAG backend detected by internal CI

anyzelman:
The internal CI detected the zip unit test failing for the hyperdags backend, when built in debug release mode.

The cause was that the zip (vector x vector -> matrix), for both pattern and non-pattern matrices did not register the output matrix as a source in the hyperDAG to be generated, thus causing vertices to be "lost". The hyperDAG backend includes an (expensive) sanity check against this, but only when compiled in debug release. This MR fixes the implementation of both matrix zip variants.


anyzelman:
Review OK. Manual check OK. GitHub CI OK. Waiting internal CI OK before merge.

anyzelman:
Internal CI OK-- merging

PR #147, opened 2022-12-22
507 copy constructor of pattern matrices does not compile

anyzelman:
Issue: copy-construction of pattern matrices did not compile.

Cause: In `compressed_storage.hpp` there was a function `copyFrom` specially made for pattern matrices. This has a few template parameters, one is `use_id`. In the other `copyFrom` functions (i.e., the non-pattern specialisation) this has a default value, but for the pattern matrix it does not.

This MR fixes the original compilation error by adding the default value `false` for `use_id` in `copyFrom` for pattern matrices. This template argument is retained so that callees can always assume the same API, regardless of whether pattern matrices are used or not. Furthermore, a unit test that copies pattern matrices into one another has been added, using both `grb::set` and the copy constructor.

This added test then triggered a bug where matrices with initial requested capacity zero would not have their offset arrays initialised. Later use (even after valid and successful resizing) would then trigger segfaults. This MR also fixes this bug.

A separate metabug concerns the tracking of freed memory, which (if the ALP/GraphBLAS was configured to emit this information) would result in the use of uninitialised memory when reporting on how many bytes were freed for a matrix with zero capacity. This MR also fixes this bug.

The MR also includes code style fixes, also in one unrelated unit test.

### DIFF
--- a/include/graphblas/reference/compressed_storage.hpp
+++ b/include/graphblas/reference/compressed_storage.hpp
@@ -1175,12 +1175,19 @@ namespace grb {
 				/**
 				 * \internal copyFrom specialisation for pattern matrices.
 				 */
-				template< bool use_id = false, typename InputType, typename UnusedType = void >
+				template<
+					bool use_id = false,
+					typename InputType,
+					typename UnusedType = void
+				>
 				void copyFrom(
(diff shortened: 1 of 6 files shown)


anyzelman:
This is a rebased version of #101

Issue #95, opened 2022-10-14
The CG algorithm exits with SUCCESS when max_iterations are reached...


anyzelman:
...even if it has not converged.

Expected behaviour: return FAILED instead.

To be confirmed.

anyzelman:
Fixed

PR #148, opened 2022-12-27
Closes GitHub issue #95: CG returns SUCCESS when not converged

anyzelman:
The CG when not converging previously returned SUCCESS, while FAILED was expected.

This MR fixes the detected issue, and also defines more clearly what a user may expect when different error codes are returned under different conditions.

No test was added due to algorithms at present only being smoke tested-- there are quite a few boundary conditions that proper unit tests should otherwise take into account.

### DIFF
--- a/include/graphblas/algorithms/conjugate_gradient.hpp
+++ b/include/graphblas/algorithms/conjugate_gradient.hpp
@@ -85,7 +85,7 @@ namespace grb {
 		 * Additional outputs (besides \a x):
 		 *
 		 * @param[out]    iterations     The number of iterations the algorithm has
-		 *                               performed.
+		 *                               started.
 		 * @param[out]    residual       The residual corresponding to output \a x.
 		 *
 		 * The CG algorithm requires three workspace buffers with capacity \f$ n \f$:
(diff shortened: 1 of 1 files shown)


anyzelman:
Review OK. GitHub CI OK. Waiting internal CI OK and smoke tests + LPF OK.

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -85,7 +85,7 @@ namespace grb {
 		 * Additional outputs (besides \a x):
 		 *
 		 * @param[out]    iterations     The number of iterations the algorithm has
-		 *                               performed.
+		 *                               started.
```
djelovina, line 88: Just a comment, if it is `out`, then is is number of iterations _performed_, is that right?

anyzelman: If SUCCESS were returned yes -- otherwise, it is possible that an iteration is stopped mid-way, hence the change

anyzelman:
Internal CI OK, smoke tests + LPF OK, will merge

PR #158, opened 2023-01-30
Speed up several cases of unmasked grb::set when the dense descriptor is given

anyzelman:
Previously, `grb::set` always accounted for sparsity in both source and destination. This MR exploits when `grb::descriptors::dense` is passed to `grb::set`, and makes sure to never read the input sparsity structure nor to write to the output sparsity structure. This is of course only valid in the unmasked case of `grb::set`.

Furthermore, this MR exploits the use of the dense descriptor if both the destination and source vectors are pattern vectors. Having such vectors that have a dense pattern, again in the unmasked case, makes the `grb::set` a no-op. The same applies when the destination vector is a pattern vector while the dense descriptor was given.

Finally, this MR moves forward an assertion check that if there is only one pattern vector, then this must be the destination vector -- it cannot be that the source is a pattern vector while the destination is not. This check previously was hidden in an if-else statement and wasn't very clear. To be complete-- this also pre-exists as a static-assert -- the moved check is one based on asserts instead, which guards against regressions and against invalid use by other backends.

As always, also a minor code style fix: rely on `grb::nnz` instead of internal coordinate functionality, since `grb::nnz` is now defined as a (cheap) getter function.

### DIFF
--- a/include/graphblas/reference/io.hpp
+++ b/include/graphblas/reference/io.hpp
@@ -760,7 +760,7 @@ namespace grb {
 						index = internal::getCoordinates( x ).asyncCopy(
 							internal::getCoordinates( y ), i );
 					} else {
-						index = start;
+						index = i;
 					}
 					if( !out_is_void && !in_is_void ) {
 						dst[ index ] = internal::setIndexOrValue< descr, OutputType >(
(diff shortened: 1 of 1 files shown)


anyzelman:
Review OK. Wait GitHub CI, internal CI, and smoke + unit tests with LPF

anyzelman:
All tests OK, will merge

PR #164, opened 2023-02-08
Implementation of the nonblocking backend

aristeidis-mastoras:
@anyzelman Could you please review the branch for the nonblocking backend and merge into develop?


anyzelman:
Just to document the review strategy:
- only cursory code style checks on files in `include/graphblas/nonblocking`;
- in-depth review of all other files.

Testing requirement aside from CI OKs are to pass the *full* test suite with LPF.

CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ endif()
 option( WITH_REFERENCE_BACKEND "With Reference backend" ON )
 option( WITH_OMP_BACKEND "With OMP backend" ON )
 option( WITH_HYPERDAGS_BACKEND "With Hyperdags backend" ON )
+option( WITH_NONBLOCKING_BACKEND "With Nonblocking backend" YES )
```
anyzelman, line 55: Change YES to ON to keep consistency with surrounding options?

bootstrap.sh
```diff
@@ -96,6 +97,7 @@ the location where LPF is installed"
 reference=yes
 hyperdags=yes
 hyperdags_using=reference
+nonblocking=yes
```
anyzelman, line 100: The nonblocking flag is never checked for in the remainder of this script. I believe it should be appended to the if-statement in line 264.

anyzelman: The script should also override `WITH_NONBLOCKING_BACKEND` to `OFF` when `nonblocking` equals `no` by passing the corresponding flag to `cmake` later on.

include/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ set( HEADERS_REGEX ".+\.(hpp|h|hxx|hh|h\\+\\+)$" )
 # to avoid flaky acrobatics with regex or glob expressions, copy main files directly
 install( FILES "graphblas.hpp" DESTINATION "${INCLUDE_INSTALL_DIR}" )
 set( root_files
-	"graphblas.hpp" "graphblas/backends.hpp" "graphblas/benchmark.hpp"
```
anyzelman, line 45: Why is `graphblas.hpp` removed from here?

include/graphblas/bsp1d/config.hpp
```diff
@@ -132,6 +132,12 @@ namespace grb {
 					return IMPLEMENTATION< _GRB_BSP1D_BACKEND >::coordinatesBackend();
 				}
 
+				/**
+				 * The selected backend may perform nonblocking execution depending on the underlying backend.
+				 */
+				static constexpr bool isNonblockingExecution() {
+					return IMPLEMENTATION< _GRB_BSP1D_BACKEND >::isNonblockingExecution();
+				}
```
anyzelman, line 140: Another empty line after this one is standard

include/graphblas/properties.hpp
```diff
@@ -30,6 +30,9 @@
 #ifdef _GRB_WITH_REFERENCE
 #include <graphblas/reference/properties.hpp>
 #endif
+#ifdef _GRB_WITH_NONBLOCKING
+#include "graphblas/nonblocking/properties.hpp"
```
anyzelman, line 34: space is missing (and elsewhere in this file also)

include/graphblas/reference/vector.hpp
```diff
@@ -231,6 +231,7 @@ namespace grb {
 		friend uintptr_t getID( const Vector< InputType, reference, Coords > & );
 
 		friend class PinnedVector< D, reference >;
+		friend class PinnedVector< D, nonblocking >;
```
anyzelman, line 235: Remove this new line? (Or add one before the nonblocking friend-- code style has no clear preference here, so the only thing's to be consistent)

include/graphblas/utils/alloc.hpp
```diff
@@ -56,6 +56,9 @@ namespace grb {
 #ifdef _GRB_WITH_HYPERDAGS
  #include "graphblas/hyperdags/alloc.hpp"
 #endif
+#ifdef _GRB_WITH_NONBLOCKING
+#include "graphblas/nonblocking/alloc.hpp"
```
anyzelman, line 60: Missing space

src/graphblas/CMakeLists.txt
```diff
@@ -74,6 +76,92 @@ set( backend_reference_srcs
...
+		target_link_libraries( backend_shmem_shared PRIVATE ${backend_nonblocking_headers} )
+	endif()
+	install( TARGETS backend_shmem_shared
+		EXPORT GraphBLASTargets
+		LIBRARY DESTINATION "${SHMEM_BACKEND_INSTALL_DIR}"
+	)
+
+	add_dependencies( libs backend_shmem_static )
+	add_dependencies( libs backend_shmem_shared )
+endif()
```
anyzelman, line 162: What happened here, why were these moved from another file?

anyzelman: I see now why the macro has moved, but should lines 126-162 not stay in `reference/CMakeLists.txt` ?

src/graphblas/reference/CMakeLists.txt
```diff
@@ -22,84 +22,13 @@
...
-	target_link_libraries( backend_shmem_shared PRIVATE ${backend_shmem_base_headers} )
-
-	install( TARGETS backend_shmem_shared
-		EXPORT GraphBLASTargets
-		LIBRARY DESTINATION "${SHMEM_BACKEND_INSTALL_DIR}"
-	)
-
-	add_dependencies( libs backend_shmem_static )
-	add_dependencies( libs backend_shmem_shared )
-endif()
```
anyzelman, line 100: See also the above comment -- should this part actually not stay in this file, with only the macro moving up to `src/graphblas/CMakeLists.txt` ?

tests/smoke/smoketests.sh
```diff
@@ -139,6 +139,7 @@ for BACKEND in ${BACKENDS[@]}; do
 				echo "Test DISABLED; GitHub runner does not have enough memory for this test"
 			else
 				echo ">>>      [x]           [ ]       Tests HPCG on a small matrix"
+				echo "Functional test executable: ${TEST_BIN_DIR}/hpcg_${BACKEND}"
```
anyzelman, line 142: I don't believe this line should be here, does line 144 actually not already have this effect? (Two options for fixing: either remove this line, or remove line 144-- slight preference for the latter actually for me)

tests/unit/unittests.sh
```diff
@@ -215,25 +215,27 @@ for MODE in debug ndebug; do
 				grep 'Test OK' ${TEST_OUT_DIR}/ewiseapply_large_${MODE}_${BACKEND}_${P}_${T} || echo "Test FAILED"
 				echo " "
 
-				echo ">>>      [x]           [ ]       Testing grb::eWiseApply using + on matrices"
-				$runner ${TEST_BIN_DIR}/eWiseApply_matrix_${MODE}_${BACKEND} &> ${TEST_OUT_DIR}/eWiseApply_matrix_${MODE}_${BACKEND}_${P}_${T}
-				head -1 ${TEST_OUT_DIR}/eWiseApply_matrix_${MODE}_${BACKEND}_${P}_${T}
-				grep 'Test OK' ${TEST_OUT_DIR}/eWiseApply_matrix_${MODE}_${BACKEND}_${P}_${T} || echo "Test FAILED"
-				echo " "
+				if [ "$BACKEND" != "nonblocking" ]; then
```
anyzelman, line 218: There are multiple such if-statements. Please group all tests that are currently not supported by the nonblocking backend together so that there is only one such if-statement, or, if it is not possible due to other such clauses, a minimum number of such if-statements.

tests/unit/unittests.sh
```diff
@@ -215,25 +215,27 @@ for MODE in debug ndebug; do
...
-				head -1 ${TEST_OUT_DIR}/copyVoidMatrices_${MODE}_${BACKEND}_${P}_${T}
-				grep 'Test OK' ${TEST_OUT_DIR}/copyVoidMatrices_${MODE}_${BACKEND}_${P}_${T} || echo "Test FAILED"
-				echo " "
+					echo ">>>      [x]           [ ]       Testing copy-constructor of square pattern matrices"
+					echo "                                 of size 1003."
+					$runner ${TEST_BIN_DIR}/copyVoidMatrices_${MODE}_${BACKEND} 1003 &> ${TEST_OUT_DIR}/copyVoidMatrices_${MODE}_${BACKEND}_${P}_${T}
+					head -1 ${TEST_OUT_DIR}/copyVoidMatrices_${MODE}_${BACKEND}_${P}_${T}
+					grep 'Test OK' ${TEST_OUT_DIR}/copyVoidMatrices_${MODE}_${BACKEND}_${P}_${T} || echo "Test FAILED"
+					echo " "
+				fi
```
anyzelman, line 238: There should be an else statement in which the script prints that several tests have been disabled for / do not apply to the nonblocking backend.

anyzelman: (This is why there's ideally only one such if-statement-- see also elsewhere in this script an example where a single if-statement just disables the remainder of the tests by issuing `continue`-- e.g., line 639)

anyzelman (changes requested):
First review pass

include/CMakeLists.txt
aristeidis-mastoras: The graphblas.hpp file is copied to the installation directory with the command in line 43. By including this file in line 44, it is copied in the subdirectory as well, which is incorrect.

src/graphblas/CMakeLists.txt
aristeidis-mastoras: The code that was moved from `src/graphblas/reference/CMakeLists.txt` to  `src/graphblas/CMakeLists.txt` is about the binary that is created for shared memory backends. Every backend should rely on the same binary, i.e., the one for shared-memory execution, just with different compilation flags. The code that was moved creates the shared-memory binary and is not specific to the reference backend. After moving this code, the `CMakeLists.txt` of every backend, e.g., reference, reference_omp, nonblocking, looks exactly the same. I think the main reason for confusion is that the name of some variables in the cmake files is misleading, since it includes the word "reference" instead of "shmem". Renaming these variables in the current branch is nontrivial, because of the hyperdags backend that actually creates a dedicated binary. There is already an internal issue to clean this up. Therefore, I would suggest that the code in this branch should remain as is, and "renaming the variables" (which will involve additional work in multiple backends) will be fixed in a separate issue.

include/CMakeLists.txt
anyzelman: Got it, thanks!!

include/graphblas/nonblocking/analytic_model.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+				/**
+				 * The number of total tiles that result from the selected tile size.
+				 */
+				size_t num_tiles;
+
+
+			public:
+
+				AnalyticModel();
```
anyzelman, line 80: Should this constructor not be disabled?

include/graphblas/nonblocking/analytic_model.hpp
```diff
@@ -0,0 +1,98 @@
...
+				AnalyticModel(
+					const size_t data_type_size,
+					const size_t vector_size,
+					const size_t accessed_vectors
+				);
+				size_t getVectorsSize() const;
+				size_t getNumThreads() const;
+				size_t getTileSize() const;
+				size_t getNumTiles() const;
```
anyzelman, line 89: Could I ask you to document these functions @aristeidis-mastoras ?

include/graphblas/nonblocking/analytic_model.hpp
```diff
@@ -0,0 +1,98 @@
...
+				void computePerformanceParameters();
```
anyzelman, line 90: Can this not be a private member function that gets called by the first constructor directly above here?

anyzelman (changes requested):
Hi Aristeidis, just some minor changes. I believe after this the following three steps are everything required before merge:
1. code style fixes to `src/nonblocking/*`;
2. enable all tests by adding an option that goes to blocking implementation of currently unsupported primitives, and enabling that option during tests; and
3. setting the pipeline data structures to a configurable fixed initial capacity (and print a warning if the data structures have been resized during lazy evaluation).

I'll do those three steps if OK for you, but could you check the issues flagged in this review? Many thanks!

include/graphblas/nonblocking/analytic_model.hpp
aristeidis-mastoras: it's currently used by the coordinates class.

include/graphblas/nonblocking/analytic_model.hpp
aristeidis-mastoras: Done.

tests/unit/set.cpp
```diff
@@ -250,27 +315,35 @@ static grb::RC dense_tests(
...
+		for( const auto &pair : dst ) {
+			if( pair.second != 0 ) {
+				std::cerr << "\t got ( " << pair.first << ", " << pair.second << " ), "
+					<< "expected zero values\n";
+				ret = FAILED;
+			}
+		}
+		if( ret != SUCCESS ) { return ret; }
 	}
-	if( ret != SUCCESS ) { return ret; }
```
anyzelman, line 270: Should this statement not remain outside of the else-if? (Consider e.g. if the nonblocking set dst back to 0 failed?)

tests/unit/set.cpp
```diff
@@ -171,27 +225,35 @@ static grb::RC dense_tests(
...
+		for( const auto &pair : dst ) {
+			if( pair.second != 0.0 ) {
+				std::cerr << "\t got ( " << pair.first << ", " << pair.second << " ), "
+					<< "expected 0\n";
+				ret = FAILED;
+			}
+		}
+		if( ret != SUCCESS ) { return ret; }
 	}
-	if( ret != SUCCESS ) { return ret; }
```
anyzelman, line 191: Should the "if ret not success" not be outside the if-else block? (Consider e.g. if the nonblocking set dst to 0 fails)

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
+	ret = ret ? ret : grb::wait( dst );
 	if( ret != ILLEGAL ) {
 		std::cerr << " expected ILLEGAL, got " << toString( ret ) << "\n";
 		return FAILED;
 	}
-	if( nnz( dst ) != 0 ) {
-		std::cerr << " expected 0, got " << nnz( dst ) << "\n";
-		return FAILED;
+	if( nonblocking_execution ) {
+		ret = ret ? ret : grb::clear( dst );
```
anyzelman, line 196: what is grb::clear here does not return success? This is currently not tested for

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
```
anyzelman, line 180: What if grb::clear here does not return success? The test now would not catch such erroneous behaviour is it not?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
```
anyzelman, line 164: What if grb::clear here does not return success? The test now would not catch such erroneous behaviour is it not?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
+	} else {
+		if( nnz( dst ) != 0 ) {
+			std::cerr << " expected 0, got " << nnz( dst ) << "\n";
+			return FAILED;
+		}
 	}
 
 	std::cerr << "\b 8:";
 	ret = grb::set( src, 3.14 );
+	ret = ret ? ret : grb::wait( src );
```
anyzelman, line 156: Isn't this one superfluous?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
+	ret = ret ? ret : grb::wait( dst );
 	if( ret != ILLEGAL ) {
 		std::cerr << " expected ILLEGAL, got " << toString( ret ) << "\n";
 		return FAILED;
 	}
-	if( nnz( dst ) != 0 ) {
-		std::cerr << " expected 0, got " << nnz( dst ) << "\n";
-		return FAILED;
+	if( nonblocking_execution ) {
+		ret = ret ? ret : grb::clear( dst );
```
anyzelman, line 146: What if grb::clear here does not return success? The test now would not catch such erroneous behaviour is it not?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
```
anyzelman, line 130: What if grb::clear here does not return success? The test now would not catch such erroneous behaviour is it not?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
```
anyzelman, line 114: What if grb::clear here does not return success? The test now would not catch such erroneous behaviour is it not?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
 	ret = grb::set< descriptors::dense >( dst, full_mask, 1.0 );
+	ret = ret ? ret : grb::wait( dst );
 	if( ret != ILLEGAL ) {
 		std::cerr << " expected ILLEGAL, got " << toString( ret ) << "\n";
 		return FAILED;
 	}
-	if( nnz( dst ) != 0 ) {
-		std::cerr << " expected 0, got " << nnz( dst ) << "\n";
-		return FAILED;
+	if( nonblocking_execution ) {
```
anyzelman, line 97: What if grb::clear here does not return success? The test now would not catch such erroneous behaviour is it not?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
+		ret = ret ? ret : grb::clear( dst );
```
anyzelman, line 82: What if grb::clear here does not return success? The test now would not catch such erroneous behaviour is it not?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
```
anyzelman, line 66: What if grb::clear here does not return success? The test now would not catch such erroneous behaviour is it not?

tests/unit/set.cpp
```diff
@@ -51,107 +57,155 @@ static grb::RC dense_tests(
...
+	} else {
+		if( nnz( dst ) != 0 ) {
+			std::cerr << " expected 0, got " << nnz( dst ) << "\n";
+			return FAILED;
+		}
 	}
 
 	std::cerr << "\b 11:";
 	ret = grb::set( dst, 0 );
+	ret = ret ? ret : grb::wait( dst );
```
anyzelman, line 206: Isn't this one superfluous?

tests/unit/set.cpp
```diff
@@ -171,27 +225,35 @@ static grb::RC dense_tests(
 
 	std::cerr << "\b 12:";
 	ret = grb::set( dst, 0 );
+	ret = ret ? ret : grb::wait( dst );
```
anyzelman, line 228: Isn't this one superfluous?

tests/unit/set.cpp
```diff
@@ -250,27 +315,35 @@ static grb::RC dense_tests(
 
 	std::cerr << "\b 16:";
 	ret = grb::set( dst, 0 );
+	ret = ret ? ret : grb::wait( dst );
```
anyzelman, line 318: Isn't this one superfluous?

tests/unit/eWiseMul.cpp
```diff
@@ -139,6 +145,10 @@ void grb_program( const size_t &n, RC &rc ) {
 		rc = SUCCESS;
 	}
 	if( rc != SUCCESS ) { return; }
+	if( nonblocking_execution ) {
+		rc = rc ? rc : clear( out );
```
anyzelman, line 149: The output error code of grb::clear is never expected. The spec says it should return SUCCESS; any other behaviour should probably cause this test to fail?

tests/unit/eWiseMul.cpp
```diff
@@ -150,6 +160,10 @@ void grb_program( const size_t &n, RC &rc ) {
 		rc = SUCCESS;
 	}
 	if( rc != SUCCESS ) { return; }
+	if( nonblocking_execution ) {
+		rc = rc ? rc : set( left, 1 );
```
anyzelman, line 164: The output error code of grb::set is never expected. Should it fail, the test should exit earlier or otherwise a misleading error will follow from the below subtest?

tests/unit/eWiseMul.cpp
```diff
@@ -161,6 +175,10 @@ void grb_program( const size_t &n, RC &rc ) {
...
```
anyzelman, line 179: The output error code of grb::set is never expected. Should it fail, the test should exit earlier or otherwise a misleading error will follow from the below subtest?

tests/unit/eWiseMul.cpp
```diff
@@ -173,6 +191,10 @@ void grb_program( const size_t &n, RC &rc ) {
 		rc = SUCCESS;
 	}
 	if( rc != SUCCESS ) { return; }
+	if( nonblocking_execution ) {
+		rc = rc ? rc : set( right, 2 );
```
anyzelman, line 195: The output error code of grb::set is never expected. Should it fail, the test should exit earlier or otherwise a misleading error will follow from the below subtest?

tests/unit/eWiseMul.cpp
```diff
@@ -184,6 +206,10 @@ void grb_program( const size_t &n, RC &rc ) {
 		rc = SUCCESS;
 	}
 	if( rc != SUCCESS ) { return; }
+	if( nonblocking_execution ) {
+		rc = rc ? rc : set( left, 1 );
```
anyzelman, line 210: The output error code of grb::set is never expected. Should it fail, the test should exit earlier or otherwise a misleading error will follow from the below subtest?

tests/unit/eWiseMul.cpp
```diff
@@ -195,6 +221,9 @@ void grb_program( const size_t &n, RC &rc ) {
...
```
anyzelman, line 225: The output error code of grb::set is never expected. Should it fail, the test should exit earlier or otherwise a misleading error will follow from the below subtest?

anyzelman (changes requested):
@aristeidis-mastoras please have a look at the unit test modifications -- they look incomplete to me. Many thanks!

anyzelman (approved):

anyzelman:
Concept merge notes:

This MR re-implements and releases the nonblocking ALP/GraphBLAS backend presented by Aristeidis Mastoras et al. [1,2] on top of the current `develop` branch.

Additionally, it includes the following improvements:
- reworks the build infrastructure to be a bit more modular in terms of packaging new backends into existing libraries;
- as a happy side-effect, removes an earlier CMake work-around that due to the above change no longer is required;
- revised specification of `eWiseLambda`, in particular, specifies additional constraints on the use of `eWiseLambda` (**only** calling `operator[]` on captured ALP vectors is allowed, any other use of the captured ALP vectors is not);
- introduces the `isBlockingExecution` and `isNonblockingExecution` `static constexpr const` properties to `grb::Properties<>`;
- introduces a test of `grb::Properties` and makes sure base properties class is now empty, forcing backends to specialise it;
- hardens some tests against the possibility of nonblocking execution;
- hardens the GitHub CI by printing the full log even if test script itself fails;
- adds missing file headers to the k-core decomposition algorithm header; and
- improves code documentation and code style issues throughout the code base.

The nonblocking backend exposed the following bugs, which this MR fixes:
- fix missing containers to call to `eWiseLambda` in `pregel.hpp` and `bicgstab.hpp`;
- the `hyperdags` backend erroneously did not provide a specialisation to `grb::Properties`;
- the `sparse_vxm` unit test erroneously did not pass one of its captured vectors as an argument to `eWiseLambda`; and
- the `vmx` and `vmxa` unit tests depended on internal functions.

[1] [Design and implementation for nonblocking execution in GraphBLAS: tradeoffs and performance](https://dl.acm.org/doi/10.1145/3561652)
by Aristeidis Mastoras, Sotiris Anagnostidis, and A. N. Yzelman, ACM Transactions on Architecture and Code Optimization 20, 1, Article 6 (March 2023).

[2] [Nonblocking execution in GraphBLAS](https://ieeexplore.ieee.org/document/9835271)
by Aristeidis Mastoras, Sotiris Anagnostidis, and A. N. Yzelman in 2022 IEEE International Parallel and Distributed Processing Symposium Workshops (IPDPSW), pp. 230-233, IEEE (2022).

anyzelman:
Waiting for internal CI OK before merge.

GitHub CI and smoke tests + LPF are OK. On the penultimate version, furthermore, the full test suite with LPF (2390 tests) passed.

anyzelman (approved):

anyzelman:
All test OK, will merge

PR #167, opened 2023-03-03
492 hpcg on steroids

alberto-scolari:
Merging modifications done to HPCG to improve

- performance
- robustness
- clarity of code
- logging ability


anyzelman:
Rebased and waiting to see if GitHub CI still fails

anyzelman:
@alberto-scolari please check the GitHub CI output -- smoke tests are not passing.

Also here I suggest to target v0.8 -- assuming there are only code style issues, given that this is a rather massive MR, it still would take quite some time to review while I'd prefer to release v0.7 asap.

It would also help if you make an initial code style pass yourself after fixing the smoke tests -- I think you're getting well familiar with the standard now, and e.g. there are some obvious too-long lines that should be easy to fix.

Many thanks!!

include/graphblas/algorithms/hpcg/average_coarsener.hpp
```diff
@@ -0,0 +1,412 @@
...
+						grb::utils::multigrid::ArrayVectorStorage< DIMS, CoordType >
+					> LinearSystemType;
+
+				typedef typename LinearSystemType::Iterator LinearSystemIterType;
+
+				typedef AverageGeneratorIterator< DIMS, CoordType, ValueType > SelfType;
+
+				typedef std::array< CoordType, DIMS > ArrayType;
+
+				class ValueGenerator {
```
anyzelman, line 97: While `_ValueGenerator` is common in some code bases where the underscore indicates a private member , it is not used anywhere in ALP -- instead, there is the `internal` namespace, or `hpcg` namespace (I know it is a nested struct in this case, but a convention for underscoring nested structs seems unnecessary to me -- the whole file is only about ~300 lines so the dev should see it)

include/graphblas/algorithms/hpcg/average_coarsener.hpp
```diff
@@ -0,0 +1,412 @@
...
+					friend SelfType;
+
+
+					private:
+
+						RowIndexType _i;
+
+						ColumnIndexType _j;
+
+						ValueType _value;
```
anyzelman, line 108: These private members were used by the below code, while the private field declaration was put at the bottom of the class. The convention is to declare it first and then use it, so I inverted the two blocks

include/graphblas/algorithms/hpcg/average_coarsener.hpp
```diff
@@ -0,0 +1,412 @@
...
+
+				typedef ValueGenerator value_type;
+
+				typedef const value_type pointer;
+
+				typedef const value_type & reference;
+
+				typedef typename LinearSystemIterType::difference_type difference_type;
+
+				AverageGeneratorIterator( const SelfType &o ) = default;
```
anyzelman, line 152: Note the lack of a space between `&` and `o`

include/graphblas/algorithms/hpcg/average_coarsener.hpp
```diff
@@ -0,0 +1,412 @@
...
+				/**
+				 * Array of steps, i.e. how much each column coordinate (finer system) must
+				 * be incremented when incrementing the row coordinates; it is the ratio
+				 * between #finer_sizes and row_generator#physical_sizes
+				 */
+				grb::utils::multigrid::ArrayVectorStorage< DIMS, CoordType > steps;
+			};
+
+		} // namespace internal
+
```
anyzelman, line 406: note the empty lines after each closing bracket

include/graphblas/algorithms/hpcg/average_coarsener.hpp
```diff
@@ -0,0 +1,412 @@
...
+	} // namespace algorithms
+
+} // namespace grb
+
+#endif // _H_GRB_ALGORITHMS_AVERAGE_COARSENER
+
```
anyzelman, line 412: note the empty line at the end of the file

anyzelman (changes requested):
Hi @alberto-scolari -- I did a code style review of the first file github served me. I already made the changes referred to in the comments here, but since they were many, I thought it good to flag what I fixed.

It looks like the remainder files have more of such issues -- we can do two things: 1) I can continue reviewing each file whenever I have time and fix the issues, or 2) would you like to take a pass over the remainder files first?

PR #174, opened 2023-04-05
V0.7.0 rc2

anyzelman:
Second release candidate for the v0.7.0 release


anyzelman:
Compiler errors with GCC 12.0.0, rc2 is rejected

PR #181, opened 2023-05-10
627 remove an alone and forgotten debug print

byjtew:
Not blaming anyone, just found it here


anyzelman (approved):
How this went undetected for so long I don't know, but many thanks for fixing!!

anyzelman:
Review OK, GitHub CI OK, paranoia doesn't want me to skip internal CI so waiting for that before merging.

Concept squash commit message:

Bugfix: stray printf in logical_or::apply removed. Thanks to Benjamin for finding this one and fixing it!

PR #183, opened 2023-05-12
629 missing type definitions for multiple backends

byjtew:
### Main idea:
Declare the Matrix internal types __value\_type__ and __const\_iterator__ for all backends, as it's already done for the _base_ and _reference_ ones. This would allow the users to consider _base_ as the general interface for a non-specialized usage of grb.

Defined in the [reference backend](https://github.com/Algebraic-Programming/ALP/blob/master/include/graphblas/reference/matrix.hpp#L1864-1872):
```c++
/** @see Matrix::value_type */
typedef D value_type;

/** The iterator type over matrices of this type. */
typedef typename internal::Compressed_Storage<
	D, RowIndexType, NonzeroIndexType
>::template ConstIterator<
	internal::Distribution< reference >
> const_iterator;
```

### DIFF
--- a/include/graphblas/bsp1d/matrix.hpp
+++ b/include/graphblas/bsp1d/matrix.hpp
@@ -318,6 +318,16 @@ namespace grb {
 
 		public:
 
+			/** @see Matrix::value_type */
+			typedef D value_type;
+
+			/** The iterator type over matrices of this type. */
+			typedef typename internal::Compressed_Storage<
+				D, RowIndexType, NonzeroIndexType
+			>::template ConstIterator<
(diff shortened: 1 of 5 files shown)


byjtew:
> Hi @byjtew , thanks for this one! I made a couple small changes, more notably perhaps being an automated test that checks for the presence of these typedefs in any backend's Matrix class definition. Are the last few commits ok for you, any issues I might have missed?

It's all good for me. I love the simplification, brilliant!

anyzelman (approved):

anyzelman:
Review OK, CIs OK, manual tests OK; will merge.

PR #184, opened 2023-05-15
632 Matrix eWiseApply implemented but not declared in base

byjtew:
The two variants were already implemented in every backend, but not declared in _base_:
- `eWiseApply([out] Matrix, [in] Matrix, [in] Matrix, [in] Monoid, ...)`
- `eWiseApply([out] Matrix, [in] Matrix, [in] Matrix, [in] Operator, ...)`

In order to have a user-friendly interface for all the grb APIs, it would be interesting to declare these variants in _base_ headers.
It would then appear [here](http://albert-jan.yzelman.net/alp/user/group__BLAS3.html).

```c++
RC eWiseApply(
	Matrix< OutputType, backend > &C,
	const Matrix< InputType1, backend > &A,
	const Matrix< InputType2, backend > &B,
	const Monoid &mulmono,
	const Phase phase = EXECUTE,
	const typename std::enable_if< !grb::is_object< OutputType >::value &&
		!grb::is_object< InputType1 >::value &&
		!grb::is_object< InputType2 >::value &&
		grb::is_monoid< Monoid >::value,
	void >::type * const = nullptr
)
```

```c++
RC eWiseApply(
	Matrix< OutputType, backend > &C,
	const Matrix< InputType1, backend > &A,
	const Matrix< InputType2, backend > &B,
	const Operator &op,
	const Phase phase = EXECUTE,
	const typename std::enable_if< !grb::is_object< OutputType >::value &&
		!grb::is_object< InputType1 >::value &&
		!grb::is_object< InputType2 >::value &&
		grb::is_operator< Operator >::value,
	void >::type * const = nullptr
)
```

### DIFF
--- a/include/graphblas/hyperdags/blas3.hpp
+++ b/include/graphblas/hyperdags/blas3.hpp
@@ -40,11 +40,10 @@ namespace grb {
 	template<
 		Descriptor descr = descriptors::no_operation,
 		typename OutputType, typename InputType1, typename InputType2,
-		typename RIT, typename CIT, typename NIT,
 		class MulMonoid
 	>
 	RC eWiseApply(
-		Matrix< OutputType, hyperdags, RIT, CIT, NIT > &C,
+		Matrix< OutputType, hyperdags > &C,
 		const Matrix< InputType1, hyperdags > &A,
 		const Matrix< InputType2, hyperdags > &B,
 		const MulMonoid &mulmono,
@@ -83,13 +82,12 @@ namespace grb {
 	template<
 		Descriptor descr = grb::descriptors::no_operation,
 		typename OutputType, typename InputType1, typename InputType2,
-		typename RIT, typename CIT, typename NIT,
 		class Operator
 	>
 	RC eWiseApply(
-		Matrix< OutputType, hyperdags, RIT, CIT, NIT > &C,
-		const Matrix< InputType1, hyperdags, RIT, CIT, NIT > &A,
-		const Matrix< InputType2, hyperdags, RIT, CIT, NIT > &B,
+		Matrix< OutputType, hyperdags > &C,
+		const Matrix< InputType1, hyperdags > &A,
+		const Matrix< InputType2, hyperdags > &B,
 		const Operator &mulOp,
 		const Phase phase = EXECUTE,
 		const typename std::enable_if<
--- a/include/graphblas/hyperdags/blas3.hpp
+++ b/include/graphblas/hyperdags/blas3.hpp
@@ -39,13 +39,16 @@ namespace grb {
 
 	template<
 		Descriptor descr = descriptors::no_operation,
+		class MulMonoid,
 		typename OutputType, typename InputType1, typename InputType2,
-		class MulMonoid
+		typename RIT, typename CIT, typename NIT,
+		typename RIT1, typename CIT1, typename NIT1,
+		typename RIT2, typename CIT2, typename NIT2
 	>
 	RC eWiseApply(
-		Matrix< OutputType, hyperdags > &C,
-		const Matrix< InputType1, hyperdags > &A,
-		const Matrix< InputType2, hyperdags > &B,
+		Matrix< OutputType, hyperdags, RIT, CIT, NIT > &C,
+		const Matrix< InputType1, hyperdags, RIT1, CIT1, NIT1 > &A,
+		const Matrix< InputType2, hyperdags, RIT2, CIT2, NIT2 > &B,
 		const MulMonoid &mulmono,
 		const Phase phase = EXECUTE,
 		const typename std::enable_if<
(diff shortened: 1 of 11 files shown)


include/graphblas/hyperdags/blas3.hpp
```diff
@@ -40,11 +40,10 @@ namespace grb {
 	template<
 		Descriptor descr = descriptors::no_operation,
 		typename OutputType, typename InputType1, typename InputType2,
-		typename RIT, typename CIT, typename NIT,
 		class MulMonoid
 	>
 	RC eWiseApply(
-		Matrix< OutputType, hyperdags, RIT, CIT, NIT > &C,
+		Matrix< OutputType, hyperdags > &C,
 		const Matrix< InputType1, hyperdags > &A,
```
anyzelman, line 47: These look like programming errors -- is it the case there was/is no unit test for this eWiseApply variant? (Is there an issue for that?)

anyzelman:
Hi @byjtew -- I made some minor code style changes, please check if OK for you

include/graphblas/hyperdags/blas3.hpp
byjtew: I can extend the _eWiseApply_matrix_ test for the Monoid variant if you want @anyzelman

anyzelman: If these are not programming errors (i.e., the code compiles when called, which I surmise in one case does work), then removing these template args means that they are assumed to revert back to their defaults, correct? In that case, it would mean passing a matrix instantiated with non-default RIT, CIT, NIT would not compile, isn't that correct?

But in such a case the removed code also does not seem to be correct, because it only applies it on C, and not on e.g. A. And potentially it is also incorrect elsewhere in the HyperDAGs backend. The thing that we then apparently do not test for, is the use of grb::Matrix with non-standard row, column, or nonzero indices.

If these all look like correct statements to you, then the best thing to do is probably to open up an issue about reviewing the Matrix signatures throughout the code base, and make sure that when they're public facing, it's templated in RIT, CIT, and NIT everywhere and for all arguments separately, and to happy path test them on all related primitives? (It's a lot of work, and imo not very high priority)

anyzelman: And a separate issue for eWiseApply_matrix test for the Monoid variant, that would be higher priority indeed=)

byjtew: > If these are not programming errors (i.e., the code compiles when called, which I surmise in one case does work), then removing these template args means that they are assumed to revert back to their defaults, correct? In that case, it would mean passing a matrix instantiated with non-default RIT, CIT, NIT would not compile, isn't that correct?

You are right, I misunderstood. 
I assumed that the template parameters would be deduced which I realise now does not make sense.
Will fix it.

byjtew: Fixed [here](https://github.com/Algebraic-Programming/ALP/pull/184/commits/6f15aa45dbcaf4f4d2f29e1ac677f2b200cd6649)

anyzelman: Thanks! I fixed similar errors throughout the code base under this MR just now, see latest commit. Will test the changes via CI as well as manually.

(It still does not include the tests)

byjtew: Ah perfect! You've been faster than me

byjtew: > And a separate issue for eWiseApply_matrix test for the Monoid variant, that would be higher priority indeed=)

I just realised that I caught this a few weeks ago: https://github.com/Algebraic-Programming/ALP/issues/203

anyzelman: Oh great! Should I help out with that one?

anyzelman: Meanwhile flagging this thread as resolved (tests will be added as part of #203 and #205).

anyzelman:
CIs and manual tests are OK. Rebasing and then setting merge process in motion.

anyzelman (approved):

include/graphblas/hyperdags/blas3.hpp
byjtew: > Oh great! Should I help out with that one?

Well, it's a missing implementation so if I have the time I'll do it. In the mean, I let it as an open issue to anyone who wants to try it.

anyzelman:
Code review OK. Waiting CIs OK before merge.

Concept release notes:

Several eWiseApply primitives for matrix inputs and outputs were implemented within all backends, but lacked a base specification. This MR adds a base specification of the operator *and* monoid based matrix eWiseApply to `include/graphblas/base/blas3.hpp`. In providing such, it found and fixes the following items:

- matrix eWiseApply did not compile for non-standard matrix index types;
- some other primitives did not compile for non-standard matrix types (e.g., `zip`); while
- yet other primitives assumed uniform matrix index types for all matrix arguments (e.g., `mxm`).

Different backends had several primitives for which this occurred; see this MR details for information on which primitives were affected. This MR included a code review that fixes all such occurrences in *all* backends except banshee--see note [1]. This MR does *not* resolve a previously flagged issue about the testing of matrix-based eWiseApply variants-- see note [2].

Additionally, this MR fixes several issues where passing the algebraic structure as a template argument would not work due to erroneously ordered template arguments. For example, both bottom lines in the following snippet should be equivalent:

```
typedef Semriring< operators::add< double >, operators::mul< double >, identities::zero, identities::one > MyRing;
MyRing ring;
grb::mxv< descriptors::no_operation >( y, A, x, ring );
grb::mxv< descriptors::no_operation, MyRing >( y, A, x );
```

The latter variant requires the algebraic structure types to follow the descriptor template argument. This MR fixes several occurrences where this was not the case. (Though both variants are supported, the former is the suggested usage.)

Finally, this MR applies various minor code style fixes throughout all files it touches.

[1] In the regular usage of ALP/GraphBLAS, the issues regarding custom matrix index types that this MR solves, do not trigger compile-time nor run-time (functional) bugs. The use or compilation of the SparseBLAS and SpBLAS transition APIs (the latter of which does employ different matrix index types) likewise does not trigger related compilation or run-time issues. This MR however notably does *not* include extensive testing for primitives with multiple matrix arguments that use differing index types-- see issue #205.

[2] In resolving this MR it was detected that unit testing only was done for the monoid-based matrix eWiseApply, not for the operator-based variant. This MR solves the issue that for new backends, code calling an unimplemented operator-based variant now does correctly return `UNSUPPORTED` (and triggering an assertion failure when in debug mode). However, it does not enable the automatic detection of such a missing implementation during standard unit testing-- this is to be resolved in issue #203 .

anyzelman:
All CIs OK, will merge

PR #188, opened 2023-05-23
633 Provide grb::foldl/foldr( [in,out] T&, [in] Matrix<D>, monoid )

### LINKED ISSUE #187: Provide grb::foldl/foldr( Matrix<D, reference>, T )
### This issue concerns the implementation for the reference+omp backend.

The __foldl__ and __foldr__ primitives, which are currently not implemented in any backend for Matrix _reduction_, appear to be crucial methods that need to be implemented. 
I suggest that we keep the same API signature as the current blas1::foldl and blas1::foldr

_Current documentation for __foldr__ (foldl is implicit):_
```c++
/**
 * Reduces, or \em folds, a matrix into a scalar.
 *
 * Reduction takes place according a monoid \f$ (\oplus,1) \f$, where
 * \f$ \oplus:\ D_1 \times D_2 \to D_3 \f$ with associated identities
 * \f$ 1_k in D_k \f$. Usually, \f$ D_k \subseteq D_3, 1 \leq k < 3 \f$,
 * though other more exotic structures may be envisioned (and used).
 *
 * Let \f$ x_0 = 1 \f$ and let
 * \f$ x_{i+1} = \begin{cases}
 *   x_i \oplus y_i\text{ if }y_i\text{ is nonzero and }m_i\text{ evaluates true}
 *   x_i\text{ otherwise}
 * \end{cases},\f$
 * for all \f$ i \in \{ 0, 1, \ldots, n-1 \} \f$.
 *
 * \note Per this definition, the folding happens in a right-to-left direction.
 *       If another direction is wanted, which may have use in cases where
 *       \f$ D_1 \f$ differs from \f$ D_2 \f$, then either a monoid with those
 *       operator domains switched may be supplied, or #grb::foldr may be used
 *       instead.
 *
 * After a successfull call, \a x will be equal to \f$ x_n \f$.
 *
 * Note that the operator \f$ \oplus \f$ must be associative since it is part
 * of a monoid. This algebraic property is exploited when parallelising the
 * requested operation. The identity is required when parallelising over
 * multiple user processes.
 *
 * \warning In so doing, the order of the evaluation of the reduction operation
 *          should not be expected to be a serial, right-to-left, evaluation of
 *          the computation chain.
 *
 * @tparam descr     The descriptor to be used (descriptors::no_operation if
 *                   left unspecified).
 * @tparam Monoid    The monoid to use for reduction.
 * @tparam InputType The type of the elements in the supplied ALP/GraphBLAS
 *                   matrix \a y.
 * @tparam IOType    The type of the output scalar \a x.
 * @tparam MaskType  The type of the elements in the supplied ALP/GraphBLAS
 *                   matrix \a mask.
 *
 * @param[out]   x   The result of the reduction.
 * @param[in]    A   Any ALP/GraphBLAS matrix.
 * @param[in]  mask  Any ALP/GraphBLAS matrix.
 * @param[in] monoid The monoid under which to perform this reduction.
 *
 * @return grb::SUCCESS  When the call completed successfully.
 * @return grb::MISMATCH If a \a mask was not empty and does not have size
 *                       equal to \a y.
 * @return grb::ILLEGAL  If the provided input matrix \a y was not dense, while
 *                       #grb::descriptors::dense was given.
 *
 * @see grb::foldl provides similar in-place functionality.
 * @see grb::eWiseApply provides out-of-place semantics.
 *
 * \parblock
 * \par Valid descriptors
 * grb::descriptors::no_operation, grb::descriptors::no_casting,
 * grb::descriptors::dense, grb::descriptors::invert_mask,
 * grb::descriptors::structural, grb::descriptors::structural_complement
 *
 * \note Invalid descriptors will be ignored.
 *
 * If grb::descriptors::no_casting is given, then 1) the first domain of
 * \a monoid must match \a IOType, 2) the second domain of \a op must match
 * \a InputType, 3) the third domain must match \a IOType, and 4) the element type
 * of \a mask must be <tt>bool</tt>. If one of these is not true, the code
 * shall not compile.
 * \endparblock
 *
 * \par Performance semantics
 * Each backend must define performance semantics for this primitive.
 *
 * @see perfSemantics
 */
template<
	class Monoid,
	typename InputType, typename IOType, typename MaskType,
	Backend backend
>
RC foldr(
	IOType &x,
	const Matrix< InputType, backend > &A,
	const Matrix< MaskType, backend > &mask,
	const Monoid &monoid = Monoid(),
	const typename std::enable_if< !grb::is_object< IOType >::value &&
		!grb::is_object< InputType >::value &&
		!grb::is_object< MaskType >::value &&
		grb::is_monoid< Monoid >::value, void
	>::type * const = nullptr
) {}
```

### LINKED ISSUE #219: hyperdags output as a scalar
In order to complete #188 (matrix-to-scalar reduction), we need the *hyperdags* backend to handle scalar as outputs.

byjtew:
Closes #187


tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+	 *  * Expected result: 2*n
+	 */
+	{
+		rc = foldLR_test(
+			"2", "A simple reduction(+) with the same types for the nzs and the reduction result.", I, mask, (NzType)n, (NzType)( 2 * n ), Monoid< operators::add< NzType >, identities::zero >() );
+		if( rc )
+			return;
+	}
+
+	/**     Test case 3:
```
anyzelman, line 289: enter after `/**`

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+	 *  A simple additive reduction with different types for the nzs and the reduction result (int <- int + NzType).
```
anyzelman, line 290: line likely too long

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+	 *  * Initial value is 0
+	 *  * Expected result: n
+	 */
+	{
+		rc = foldl_test( "3", "A simple reduction(+) with different types for the nzs and the reduction result (int <- int + NzType).", I, mask, (int)0, (int)n,
```
anyzelman, line 295: line too long plus casting in this code base should be done using `static_cast`

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+			Monoid< operators::add< int, NzType, int >, identities::zero >() );
+		if( rc )
+			return;
+		rc = foldr_test( "3", "A simple reduction(+) with different types for the nzs and the reduction result (int <- NzType + int).", I, mask, (int)0, (int)n,
```
anyzelman, line 299: same here

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+	{
+		rc = foldl_test( "4", "A simple reduction(+) with different types for the nzs and the reduction result (int <- int + NzType).", I, mask, (int)n, (int)( 2 * n ),
+			Monoid< operators::add< int, NzType, int >, identities::zero >() );
+		if( rc )
+			return;
+		rc = foldr_test( "4", "A simple reduction(+) with different types for the nzs and the reduction result (int <- NzType + int).", I, mask, (int)n, (int)( 2 * n ),
+			Monoid< operators::add< NzType, int, int >, identities::zero >() );
+		if( rc )
+			return;
+	}
```
anyzelman, line 319: similar comments

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+		std::vector< size_t > rows( grb::nrows( I ) * grb::ncols( I ) ), cols( grb::nrows( I ) * grb::ncols( I ) );
+		for( size_t x = 0; x < grb::nrows( I ); x++ ) {
+			std::fill( rows.begin() + x * grb::ncols( I ), rows.begin() + ( x + 1 ) * grb::ncols( I ), x );
+			std::iota( cols.begin() + x * grb::ncols( I ), cols.begin() + ( x + 1 ) * grb::ncols( I ), 0 );
+		}
+		buildMatrixUnique( dense_mask, rows.data(), cols.data(), rows.size(), SEQUENTIAL );
+		rc = foldLR_test<descriptors::add_identity>( "16", "Reduction with a dense void mask, with the descriptors::add_identity.", I, dense_mask, (NzType)0, (NzType)n + std::min( nrows( I ), ncols( I ) ), Monoid< operators::add< NzType >, identities::zero >(), false, false );
+		if( rc )
+			return;
+	}
```
anyzelman, line 520: similar comments

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+	if( argc > 2 ) {
+		printUsage = true;
+	}
+	if( argc == 2 ) {
+		n = std::atol( argv[ 1 ] );
+	}
+	if( printUsage ) {
+		std::cerr << "Usage: " << argv[ 0 ] << " [n]\n";
+		std::cerr << "  -n (optional, default is 10): an even integer, the test "
+				  << "size.\n";
```
anyzelman, line 538: one tab too much?

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+	grb::Launcher< AUTOMATIC > launcher;
+	grb::RC rc = RC::SUCCESS;
+
+	if( ! rc ) { // Identity square-matrix
+		Matrix< NzType > I( n, n );
+		std::vector< size_t > I_rows( n ), I_cols( n );
+		std::vector< NzType > I_vals( n, 1.f );
+		std::iota( I_rows.begin(), I_rows.end(), 0 );
+		std::iota( I_cols.begin(), I_cols.end(), 0 );
+		buildMatrixUnique( I, I_rows.data(), I_cols.data(), I_vals.data(), I_vals.size(), SEQUENTIAL );
```
anyzelman, line 552: return code ignored, line may be too long?

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+		Matrix< void > mask( n, n );
+		buildMatrixUnique( mask, I_rows.data(), I_cols.data(), I_rows.size(), SEQUENTIAL );
```
anyzelman, line 554: return code ignored, line may be too long?

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+	}
+
+	if( ! rc ) { // Build a square-matrix with n 1s on the first row
+		Matrix< NzType > I( n, n );
+		std::vector< size_t > I_rows( n, 0 ), I_cols( n );
+		std::vector< NzType > I_vals( n, 1.f );
+		std::iota( I_cols.begin(), I_cols.end(), 0 );
+		buildMatrixUnique( I, I_rows.data(), I_cols.data(), I_vals.data(), I_vals.size(), SEQUENTIAL );
+		Matrix< void > mask( n, n );
+		buildMatrixUnique( mask, I_rows.data(), I_cols.data(), I_rows.size(), SEQUENTIAL );
```
anyzelman, line 571: similar comments (return code, line length)

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+		Matrix< void > mask( n, 1 );
+		buildMatrixUnique( mask, I_rows.data(), I_cols.data(), I_rows.size(), SEQUENTIAL );
+		std::cout << "-- Running test 06: [n = " << n << " rows, 1 column] matrix, filled with 1s" << std::endl;
+		input< NzType, void > input(I, mask);
+		if( launcher.exec( &grb_program, input, rc, true ) != SUCCESS ) {
+			std::cerr << "Launching test 06 FAILED\n";
+			return 255;
+		}
+		std::cout << std::endl << std::flush;
+	}
```
anyzelman, line 648: similar comments (space after `!`, return codes ignored, line lengths)

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+
+	if( rc != SUCCESS ) {
+		std::cout << "Test FAILED (" << grb::toString( rc ) << ")" << std::endl;
+		return rc;
+	} else {
+		std::cout << "Test OK" << std::endl;
+		return 0;
+	}
+}
```
anyzelman, line 657: missing empty line at the end

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+		if( launcher.exec( &grb_program, input, rc, true ) != SUCCESS ) {
+			std::cerr << "Launching test 06 FAILED\n";
+			return 255;
+		}
+		std::cout << std::endl << std::flush;
+	}
+
+	if( rc != SUCCESS ) {
+		std::cout << "Test FAILED (" << grb::toString( rc ) << ")" << std::endl;
+		return rc;
```
anyzelman, line 652: the tests so far typically return an error code that corresponds to the failed subtest, which helps debugging without needing to involve gdb to see what failed where. I believe this test can effect that by having the helper subtest functions return an int error code that is unique to the `return` statement. Typically we use increments of `10` (rather than successive error numbers) to allow for the situation where a subtest may be inserted at some later stage

include/graphblas/reference/blas3.hpp
```diff
@@ -57,6 +57,25 @@
...
+			#define _DEBUG_PRINT( msg ) \
+				_GRB_DEBUG_CRITICAL_SECTION \
+				{ \
+					std::cout << "[T" << omp_get_thread_num() << "] - " << msg << std::flush; \
+				}
+		#else
+			#define _DEBUG_PRINT( msg ) std::cout << msg << std::flush;
+		#endif
+	#endif
+#endif
```
anyzelman, line 77: This is a tricky thing to add -- if it's here, maybe it should be used everywhere in reference and beyond

If adding it everywhere, I'm not that keen on inserting many critical sections as part of debugging.

If this is to become a logger mechanism, it clashes with one that Alberto was also working on

If this is instead only meant for this file only, would it not be better to just use the current `#ifdef _DEBUG` everywhere? (Is it that critical(;) to have the thread number and critical sections for the functions in this file?)

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+				_DEBUG_PRINT( "The input matrix is empty, nothing to compute\n" );
+				return rc;
+			}
+
+			if ( descr & descriptors::force_row_major && descr & descriptors::transpose_left ) {
+				_DEBUG_PRINT( "Masked fold with force_row_major and transpose_left is not supported\n" );
+				return RC::ILLEGAL;
+			}
+			if ( descr & descriptors::force_row_major && descr & descriptors::transpose_matrix ) {
+				_DEBUG_PRINT( "Masked fold with force_row_major and transpose_matrix is not supported\n" );
```
anyzelman, line 964: maybe line length too long?

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+				return RC::ILLEGAL;
```
anyzelman, line 965: is it that it cannot be supported or that there currently is no implementation? (to me it seems like transpose or no can still both be supported by CRS only?)

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+			}
+
+			const auto &A_raw = (descr & grb::descriptors::transpose_matrix || descr & grb::descriptors::transpose_left ) ?
```
anyzelman, line 968: line length

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+				internal::getCCS( A ) : internal::getCRS( A );
+			const size_t A_nnz = nnz( A );
+
+			const auto& op = monoid.getOperator();
```
anyzelman, line 972: `auto& op` -> `auto &op`

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+			RC local_rc = rc;
+			auto local_x = monoid.template getIdentity< typename Monoid::D3 >();
+
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+	#pragma omp parallel default(none) shared(A_raw, x, rc, std::cout) firstprivate(local_x, local_rc, A_nnz, op)
```
anyzelman, line 977: tabs here can be aligned with the `{`. Also, line length. The code so far typically does not use `default(none)` for OMP regions -- is it totally necessary here?

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+
+				for( size_t idx = start; idx < end; ++idx ) {
+					// Get A value
+					const InputType a_val = A_raw.values[ idx ];
+					_DEBUG_PRINT( "A.values[ " + std::to_string( idx ) + " ] = " + std::to_string( a_val ) + "\n" );
+
+					// Compute the fold for this coordinate
+					auto local_x_before = local_x;
+					local_rc = local_rc ? local_rc : grb::apply< descr >( local_x, local_x_before, a_val, op );
+					_DEBUG_PRINT( "Computing: local_x = op(" + std::to_string( a_val ) + ", " + std::to_string( local_x_before ) + ") = " + std::to_string( local_x ) + "\n" );
```
anyzelman, line 994: line length?

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+				}
+
+
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+	#pragma omp critical
+#endif
+				{ // Reduction with the global result (critical section if OpenMP)
+					auto x_before = x;
+					local_rc = local_rc ? local_rc : grb::apply< descr >( x, x_before, local_x, op );
+					_DEBUG_PRINT( "Computing x: op(" + std::to_string( local_x ) + ", " + std::to_string( x_before ) + ") = " + std::to_string( x ) + "\n" );
```
anyzelman, line 1004: line length

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+				_DEBUG_PRINT( "Masked fold with force_row_major and transpose_matrix is not supported\n" );
+				return RC::ILLEGAL;
+			}
+
+			const auto &A_raw = (descr & grb::descriptors::transpose_matrix || descr & grb::descriptors::transpose_left ) ?
+				internal::getCCS( A ) : internal::getCRS( A );
+			const size_t m = (descr & grb::descriptors::transpose_matrix || descr & grb::descriptors::transpose_left ) ?
+				ncols( A ) : nrows( A );
+			const size_t n = (descr & grb::descriptors::transpose_matrix || descr & grb::descriptors::transpose_left ) ?
+				nrows( A ) : ncols( A );
```
anyzelman, line 1045: line lengths?

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+
+			const auto& op = monoid.getOperator();
+			const auto& identity = monoid.template getIdentity< typename Monoid::D3 >();
```
anyzelman, line 1048: `auto& x`->`auto &x

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +937,301 @@ namespace grb {
...
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+	#pragma omp critical
+#endif
+				{ // Reduction with the global result (critical section if OpenMP)
+					auto x_before = x;
+					local_rc = local_rc ? local_rc : grb::apply< descr >( x, x_before, local_x, op );
+					_DEBUG_PRINT( "Computing x: op(" + std::to_string( local_x ) + ", " + std::to_string( x_before ) + ") = " + std::to_string( x ) + "\n" );
+					rc = rc ? rc : local_rc;
+				}
+			}
```
anyzelman, line 1229: similar comments here -- I can review again in more details once you made an initial pass?

anyzelman (changes requested):
Hi Benjamin, thanks for this one!!

Couple of comments, mostly on code style, throughout the MR as indicated in this report.

One overall message on testing: unit tests typically should also test that specific error codes are returned when they are expected -- e.g., `MISMATCH` in case of the masked primitives. I believe this currently is not tested for (in fact it sometimes seems that error codes are lost/ignored?).

anyzelman (changes requested):
Hi Benjamin, thanks for this one!!

Couple of comments, mostly on code style, throughout the MR as indicated in this report.

One overall message on testing: unit tests typically should also test that specific error codes are returned when they are expected -- e.g., `MISMATCH` in case of the masked primitives. I believe this currently is not tested for (in fact it sometimes seems that error codes are lost/ignored?).

byjtew:
> Hi Benjamin, thanks for this one!!
> 
> Couple of comments, mostly on code style, throughout the MR as indicated in this report.
> 
> One overall message on testing: unit tests typically should also test that specific error codes are returned when they are expected -- e.g., `MISMATCH` in case of the masked primitives. I believe this currently is not tested for (in fact it sometimes seems that error codes are lost/ignored?).

That's clearly an oversight. 
I'll fix all the styles+tests then will re-ask for your review.

Thank you!

include/graphblas/blas0.hpp
```diff
@@ -34,6 +34,7 @@
 #include "graphblas/descriptors.hpp"
 #include "graphblas/rc.hpp"
 #include "graphblas/type_traits.hpp"
+#include "graphblas/identities.hpp"
```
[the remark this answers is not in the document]
byjtew: It **was** needed, will be removed

include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,239 @@ namespace grb {
...
+	/**
+	 * Reduces, or \em folds, a matrix into a scalar. 
+	 * Right-to-left unmasked variant.
+	 * 
+	 * Please see the masked grb::foldr variant for a full description.
+	 * 
+	 * \parblock
+	 * 
+	 * \par Valid descriptors specific to this variant
+	 * - descriptors::transpose_matrix: A^T will be considered instead
```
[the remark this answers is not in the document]
byjtew: Valid descriptors specific **to this variant**

include/graphblas/hyperdags/blas3.hpp
```diff
@@ -332,6 +332,172 @@ namespace grb {
...
+		if( ret != SUCCESS ) { return ret; }
+		if( nrows( A ) == 0 || ncols( A ) == 0 ) { return ret; }
+		std::array< const void *, 0 > sourcesP{};
+		std::array< uintptr_t, 2 > sourcesC{
+			getID( internal::getMatrix(A) ),
+			getID( internal::getMatrix(mask) )
+		};
+		std::array< uintptr_t, 0 > destinations{};
+		// NOTE scalar output is ignored
+		// std::array< uintptr_t, 1 > destinations{ &x };
```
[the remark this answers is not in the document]
byjtew: It is the exact same code in **dot**, except if you meant input

byjtew:
Waiting for CI before new review

include/graphblas/blas0.hpp
```diff
@@ -604,6 +605,37 @@ namespace grb {
...
+				template < Descriptor descr = descriptors::no_operation, typename MaskStruct >
+				MaskHasValue( const MaskStruct& mask_raw, const size_t k ) {
+						bool hasValue = (bool) mask_raw.values[ k ];
+						if (descr & grb::descriptors::invert_mask) {
+							hasValue = !hasValue;
+						}
+						value = hasValue;
+					}
+
+				bool value;
```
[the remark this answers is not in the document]
byjtew: Actually this was not really required so I removed it

include/graphblas/nonblocking/blas3.hpp
```diff
@@ -571,6 +571,118 @@ namespace grb {
...
+	) {
+#ifdef _DEBUG
+		std::cout << "In grb::foldr( nonblocking, matrix, mask, monoid )\n";
+#endif
+		// nonblocking execution is not supported
+		// first, execute any computation that is not completed
+		internal::le.execution();
+
+		// second, delegate to the reference backend
+		return foldr< descr, Monoid >( x, internal::getRefMatrix( A ), internal::getRefMatrix( mask ), monoid );
```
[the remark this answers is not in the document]
byjtew: Will check that

tests/unit/fold_matrix_to_scalar.cpp
```diff
@@ -0,0 +1,657 @@
...
+constexpr bool SKIP_UNMASKED = false;
+constexpr bool SKIP_MASKED = false;
+// Benchmarking
+constexpr bool PRINT_TIMERS = false;
+constexpr size_t ITERATIONS = 1;
+
+// #define _DEBUG
+
+template< class Iterator >
+void printSparseMatrixIterator( size_t rows, size_t cols, Iterator begin, Iterator end, const std::string & name = "", std::ostream & os = std::cout ) {
```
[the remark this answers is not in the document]
byjtew: I actually don't even know if I should let this function, you decide

include/graphblas/reference/blas3.hpp
byjtew: Will remove, we'll adapt using Alberto's work

include/graphblas/base/blas3.hpp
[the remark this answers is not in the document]
anyzelman: I guess we interpret that differently;)

The way I interpret it, is that what follows, is a list of descriptors that can be given to the current variant. This list is only given for variants that allows descriptors that are not in the base variant OR in the case that it does not allow for descriptors that are in the base variant.

The rationale for providing a full list is that it would be too cumbersome for readers of a reference document to only define what is added to or missing from the descriptors of the base variant.

anyzelman:
> Waiting for CI before new review

lemme know when I can get back in on this one:)

include/graphblas/hyperdags/blas3.hpp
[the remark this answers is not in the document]
anyzelman: Ah you're right -- i saw the output `z` being registered but that's as input indeed (it's inplace). For the purpose of this MR then let's keep it like this, and worry as part of #219 if we should enable scalars as output and if yes, do it there. (I will append in that MR also perhaps whether iterators should be considered output.)

Anyway, for this MR, I suggest to disregard my comments regarding this=)

tests/unit/fold_matrix_to_scalar.cpp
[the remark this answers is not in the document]
anyzelman: Deleting it is also ok for me, maybe even preferred (ideally a separate MR contributes utilities like these)

include/graphblas/base/blas3.hpp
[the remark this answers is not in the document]
byjtew: I agree, do you have any suggestion regarding the documentation before I copy it to all the variants ?

byjtew:
> > Waiting for CI before new review
> 
> lemme know when I can get back in on this one:)

You can start already

include/graphblas/base/blas3.hpp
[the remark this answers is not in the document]
anyzelman: OK. So I would suggest listing all descriptors that are valid for this specific variant (and similarly elsewhere), including descriptors::no_operation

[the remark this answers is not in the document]
byjtew: Sure, I meant suggestions for the rest of the documentation

tests/unit/fold_matrix_to_scalar.cpp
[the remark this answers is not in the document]
byjtew: I'll try to do a separate MR for utilities like this one, that could be pretty useful, I was also thinking about some matrix/vectors factory for testing purposes, like *create_identity*, *create_dense*

include/graphblas/reference/blas3.hpp
```diff
@@ -1247,134 +1237,91 @@ namespace grb {
...
-			firstprivate(local_x, local_rc, m, n, op, identity)
+			firstprivate(local_x, local_rc, m, n, op, identity, mask_identity)
 #endif
 			{
 				size_t start_row = 0;
 				size_t end_row = m;
 #ifdef _H_GRB_REFERENCE_OMP_BLAS3
 				config::OMP::localRange( start_row, end_row, 0, m );
 #endif
+				bool local_visited_columns[ n ] = { false };
```
anyzelman, line 1260: Instead of this one, which performs dynamic allocation on the fly, the suggested approach is to use a pre-allocated thread-local SPA. In the (sequential) reference backend, this is done via `internal::getMatrixBuffers`, as exemplified in `reference/blas3.hpp`. Those buffers are then used to create a temporary `internal::Coordinates` class, to play the role of `local_visited_columns`.

The difference with this code is, however, that we want to run in parallel and that `local_visited_columns` is a thread-local array. This leads to a scalability issue: if there are `T` threads each with an array of size `n`, then `T*n` may well be much larger than `nz`, where `nz` is the number of nonzeroes the primitive is to process-- meaning, calling this primitive would require a much larger work space than the operands of the primitive would require. This does not scale, and therefore in this parallelisation strategy, `T` must be limited to something that is scalable (e.g., `T=floor(nz/n)`).

ALP does guarantee at least one buffer (through `getMatrixBuffers`). It also does guarantee a global buffer space that is proportional to the largest `nz` a container has ever been constructed with. This buffer can be retrieved via calls to
```
const size_t nbytes = internal::template getCurrentBufferSize< char >();
char * buffer = internal::template getReferenceBuffer< char >( nbytes );
```
(The above example retrieves it as a raw byte buffer. In the current context, `nbytes` could be used to derive how many `local_visited_columns` are supported by the current global buffer, and select `T` accordingly. The reference backends during the construction of vector and matrix containers and during any call to `grb::resize` ensures that the global buffer is always of size `Theta( max{ m, n, nz } )`.)

The motivation here are performance semantics -- we would like ALP primitives to execute without performing any system calls, which entails pre-allocating buffers like these ones, without incurring unscalable behaviour.

byjtew: Very clear answer thank you.

I tried to go another way without any additional allocation and threading, not sure about it though. 
See the masked variant of reference/blas3.hpp::fold_masked_generic @anyzelman

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +919,330 @@ namespace grb {
...
+				RC local_rc = rc;
+				size_t start = 0;
+				size_t end = A_nnz;
+				auto local_x = monoid.template getIdentity< typename Monoid::D3 >();
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+				config::OMP::localRange( start, end, 0, A_nnz );
+#endif
+				for( size_t idx = start; idx < end; ++idx ) {
+					// Get A value
+					const InputType a_val = A_raw.getValue( idx, InputType() );
```
anyzelman, line 978: This should not be a default-constructed InputType here, but rather a semiring identity (one, not zero(!)).

Recent changes already static-assert that void matrices will not pass to this function if the monoid variant is called. The fix here would be to add an optional identity argument to the `fold_unmasked_generic` (e.g., a pointer to a one that is NULL when entered from a monoid variant).

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +919,330 @@ namespace grb {
...
+			Descriptor descr,
+			class Monoid, bool left_handed,
+			typename InputType, typename IOType,
+			typename RIT, typename CIT, typename NIT
+		>
+		RC fold_unmasked_generic(
+			IOType &x,
+			const Matrix< InputType, reference, RIT, CIT, NIT > &A,
+			const Monoid &monoid
+		) {
```
anyzelman, line 932: I believe this function should static_assert that descr does not contain add_identity?

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +919,330 @@ namespace grb {
...
+			return rc;
+		}
+
+		template<
+			Descriptor descr,
+			class Ring, bool left_handed,
+			typename InputType, typename IOType,
+			typename RIT, typename CIT, typename NIT
+		>
+		RC fold_add_identity_unmasked_generic(
```
anyzelman: Note on recent change: this function is only called in the semiring case, and therefore can safely require a Ring instead of a monoid

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +919,330 @@ namespace grb {
...
+					for(
+						auto k = A_raw.col_start[ start ];
+						k < A_raw.col_start[ i + 1 ];
+						k++
+					) {
+						const auto j = A_raw.row_index[ k ];
+
+						// Check if the element of the main diagonal has been touched in this row
+						identity_element_computed |=
+							(descr & descriptors::add_identity) && i == j;
```
anyzelman: I believe this will go wrong when called from the BSP1D backend -- we need to get the distribution in here as a template argument, and the check (if really to be implemented like this) would then have to first translate i and j from local to global coordinates

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +919,330 @@ namespace grb {
...
+
+						// descriptors::add_identity logic
+						const InputType identity_increment = ( i == j )
```
anyzelman: same here (noting since it is critical -- currently, the bsp1d and hybrid unit tests for this MR fail due to segfault, I think likely because of this)

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +919,330 @@ namespace grb {
...
+										local_x_before, a_val + identity_increment,
+										op
+									);
+							} else {
+								local_rc = grb::apply< descr >(
+										local_x,
+										a_val + identity_increment, local_x_before,
+										op
+									);
+							}
```
anyzelman: Note on recent changes: the orientation of the fold (left or right) matters, but was previously ignored. This pattern using the `left_handed` template bool arg is now used consistently everywhere.

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +919,330 @@ namespace grb {
...
+			auto mask_k = mask_raw.col_start[ 0 ];
+			auto A_k = A_raw.col_start[ 0 ];
+
+			for( size_t i = 0; i < m_A; ++i ) {
+				coors.clear();
+
+				for(; mask_k < mask_raw.col_start[ i + 1 ]; ++mask_k ) {
+					if( !ignore_mask_values &&
+						mask_raw.getValue( mask_k, identity_mask )
+					) { continue; }
```
anyzelman: should this not use `utils::interpretMask` to handle the various mask descriptors?

(Note that with recent changes, it is not allowed for a descriptor to contain both inversion and structural, so the use of utils::interpretMask should handle all allowed cases)

include/graphblas/reference/blas3.hpp
```diff
@@ -918,6 +919,330 @@ namespace grb {
...
+				if( coors.nonzeroes() == 0 ) { continue; }
+
+				for(; A_k < A_raw.col_start[ i + 1 ]; ++A_k ) {
+					const auto j = A_raw.row_index[ A_k ];
+
+					// Skip if the coordinate is not in the mask
+					if( !coors.assigned( j ) ) { continue; }
+
+					// descriptors::add_identity logic for the main diagonal of A
+					const InputType identity_increment = add_identity && ( i == j )
```
anyzelman: same issue with local vs. global coordinates (noting explicitly since critical)

anyzelman (changes requested):
Intermediate request for changes -- I did not finish reviewing the entire MR yet.

I did revise the spec and brought the tests and current implementation in line with the revised spec. One major change is that the foldl/foldr from matrix to scalar now can take either a monoid or a semiring, and (thus) one more additional rfc is to expand the unit test to repeat all non-add-identity tests twice, once with monoids, and once with semirings.

byjtew:
Since this MR should be closed soon, would it be okay to discard the add_identity logic and simply handle the trivial case instead @anyzelman ?

include/graphblas/reference/blas3.hpp
anyzelman: I'm resolving this by adding a static-assert that InputType is never void (and after checking that this is indeed the case at present). In such case, the default-constructed element will of course be never used:)

anyzelman:
> Since this MR should be closed soon, would it be okay to discard the add_identity logic and simply handle the trivial case instead @anyzelman ?

Ah sorry -- yes this is OK, but we should keep the Semiring variants. Could we roll-back / revert the last commit and surgically remove only the bsp1d-broken add_identity logic instead, as well as remove from the base specification that add_identity would be supported, and finally also add an issue for milestone v0.9 to add support for add_identity? I'll move on to some other MRs first but after that one could also give the above suggested changes a shot

anyzelman (approved):

anyzelman:
Final tests running, will merge if OK. Concept release notes:

This MR specifies and implements matrix-to-scalar reduction routines. The matrix may be input masked (by a separate, second given matrix). The reductions take either a commutative monoid or a semiring -- if the latter, its additive monoid will be used for reduction (which, by definition of a semiring, is also commutative). The monoid variants cannot handle pattern matrices, and cannot handle the `add_identity` descriptor-- for these, the "one" identity of a semiring is necessary.
Since with ALP/GraphBLAS, operators, from which monoids and semirings are comprised, may have different input and different output domains (i.e., `A x B -> C`), this MR introduces left-looking and right-looking variants of this reduction operation and calls them `foldl` and `foldr`, respectively. This is in line with the previously already included primitives for scalar-to-scalar and vector-to-scalar reduction.

The MR introduces exhaustive unit testing for the new functionality, and, as always, includes code style fixes in files related and unrelated to this new feature. Notable other inclusions:
- an internal utility has been introduced that helps asserting return codes are synchronised across multiple user processes;
- a systematic way to derive index-translation lambdas for so-called "union coordinates" has been introduced. These lambdas are by convention used as a mechanism to let the BSP1D backend (or any distributed backend) translate process-local calls to any single-process final backend. This new feature may yet be further improved-- see GitHub issue #322 .

Many thanks to Benjamin for the implementation and testing for this one!

PR #199, opened 2023-06-07
639 Single-source shortest path (SSSP) algorithm

### LINKED ISSUE #198: Single-Source Shortest-Path (SSSP) algorithm
Implementation of a Single-Source Shortest-Path (SSSP) algorithm.

byjtew:
Closes #198.


anyzelman:
Depends on #213 which in turn depends on #271

anyzelman:
TODO:

- [ ] remove (or move) the utility functions,
- [ ] remove the need for the distances vector,

and then it's ok. The main diff. with mpv is early stopping criteria.

PR #200, opened 2023-06-08
642 separating built from tested backends

alberto-scolari:
This solves the issue with the internal CI by by separating backends being tested (i.e., run tests) and the backends being built, which depend on the user's selction. E.g.:

1. the user selects the hybrid backend and excludes all other backends
2. both the hybrid and the reference_omp backends are built because of the dependency between backends
3. only the tests for the hybrid backend are built and run

This is in line with the backends "composability".
Thanks a lot to @byjtew for providing the code for this fix.


anyzelman:
From the description I don't totally get what's currently going wrong on the internal CI -- could you (@byjtew or @alberto-scolari) please clarify? Many thanks!!

.gitlab-ci.yml
```diff
@@ -6,42 +6,51 @@ default:
...
+        MULTIPLE_COMPILERS_ENABLED: "yes"
     - if: $CI_COMMIT_BRANCH =~ /^v.*-rc.*$/
       variables:
-        IS_MAIN_PIPELINE: "yes"
-  # rules to avoid duplicate pipelines on merge requests
+        EXTRA_TESTS_ENABLED: "yes"
+        LPF_TESTS_ENABLED: "yes"
+        MULTIPLE_COMPILERS_ENABLED: "yes"
+    # "manual" execution strategies are allowed to force run
+    - if: $CI_PIPELINE_SOURCE != "merge_request_event" && $CI_PIPELINE_SOURCE != "push"
```
anyzelman, line 32: Does a manual trigger of the pipeline require manually setting the `*_ENABLED` flags? (On Jenkins the GUI would expose the available flags and make it easy to change them, when manually triggering a run -- not sure if Gitlab allows the same?)

CMakeLists.txt
```diff
@@ -52,6 +52,11 @@ endif()
 option( WITH_REFERENCE_BACKEND "With Reference backend" ON )
 option( WITH_OMP_BACKEND "With OMP backend" ON )
 option( WITH_HYPERDAGS_BACKEND "With Hyperdags backend" ON )
+if( WITH_HYPERDAGS_BACKEND )
+	if ( NOT DEFINED WITH_HYPERDAGS_USING )
```
anyzelman, line 56: stray space after `if`

CMakeLists.txt
```diff
@@ -128,6 +131,64 @@ if( NOT WITH_REFERENCE_BACKEND AND
...
+# Enabling reference(+omp) backend if bsp1d is active
+if( WITH_BSP1D_BACKEND )
+	set( WITH_REFERENCE_BACKEND ON )
+	set( WITH_OMP_BACKEND ON )
+	message( STATUS "Enabling compilation of reference backend: required by bsp1d" )
+endif()
+
+# Enabling reference(+omp) backend if hybrid is active
+if( WITH_HYBRID_BACKEND )
+	set( WITH_OMP_BACKEND ON )
```
anyzelman, line 164: should probably be `WITH_REFERENCE_BACKEND` ?

CMakeLists.txt
```diff
@@ -128,6 +131,64 @@ if( NOT WITH_REFERENCE_BACKEND AND
...
+	set( WITH_OMP_BACKEND ON )
+	message( STATUS "Enabling compilation of reference_omp backend: required by hybrid" )
```
anyzelman, line 166: should maybe be `reference and reference_omp` instead of only `reference_omp` ?

CMakeLists.txt
```diff
@@ -128,6 +131,64 @@ if( NOT WITH_REFERENCE_BACKEND AND
...
+	elseif(WITH_HYPERDAGS_USING STREQUAL "reference_omp" )
+		set( WITH_OMP_BACKEND ON )
+		message( STATUS "Enabling reference_omp backend for hyperdags" )
+	elseif(WITH_HYPERDAGS_USING STREQUAL "nonblocking" )
+		set( WITH_NONBLOCKING_BACKEND ON )
+		message( STATUS "Enabling non-blocking backend for hyperdags" )
+	else()
+		message( FATAL_ERROR "Unrecognised backend for hyperdags: ${WITH_HYPERDAGS_USING}.\nAvailable options: reference, reference_omp, nonblocking" )
+	endif()
+endif()
```
anyzelman, line 190: Should this code block not appear before line 136? (Or is it the idea that the tests of the backends hyperdags depend on are not enabled unless explicitly asked for?)

tests/unit/CMakeLists.txt
```diff
@@ -240,12 +239,20 @@ add_grb_executables( outer outer.cpp
 	BACKENDS reference reference_omp hyperdags nonblocking
 )
 
+# must generate the golden output for other tests
+force_add_grb_executable( mxv mxv.cpp
+	BACKEND reference
+)
```
anyzelman, line 245: nice!

tests/unit/buildMatrixUnique.cpp
```diff
@@ -598,7 +598,8 @@ void grbProgram( const void *, const size_t, int &error ) {
 			{ spmd<>::nprocs(), spmd<>::nprocs() },
 			{ 77, 70 },
 			{ 130, 139 },
-			{ 1463, 5376 }
+			{ 146, 5376 }
+			// { 1463, 5376 } // MPI in CI has issues with this size
```
anyzelman, line 602: Should we not fix that issue, rather than disabling that size?

anyzelman (changes requested):
Cursory review based on current understanding-- many thanks!

.gitlab-ci.yml
byjtew: Gitlab is able to recognise the variables and expose them in the interface. 
It is now possible, see [this commit](https://github.com/Algebraic-Programming/ALP/pull/200/commits/530e2a1e4d3a373c97f41d2ae05dc88dc72616e9)

CMakeLists.txt
byjtew: It is precisely for that idea. 
We allow the user to test the backends he asked for only, and then enable the compilation for the necessary ones.

CMakeLists.txt
byjtew: My bad. 
Fixed [here](https://github.com/Algebraic-Programming/ALP/pull/200/commits/5b972d4890323e1b92237b8928a8cb4c468ac038)

alberto-scolari:
> From the description I don't totally get what's currently going wrong on the internal CI -- could you (@byjtew or @alberto-scolari) please clarify? Many thanks!!

There two problems currently:

1. some tests fail (this you know)
2. currently, when we run unit tests for LPF (bsp1d and hybrid), we also run unit tests for reference and reference_omp, because of the dependencies among backends; in other words: the dependencies among backends also reflect onto tests

Point 2 makes the LPF unit tests (job `tests_unit_lpf`) much longer (with the risk of timeouts) and is useless, because unit tests for reference and reference_omp already run within job `tests_unit` (which runs for any pipeline). Therefore, this PR introduces a separation between "which backends are built" and "which backends are tested": the user chooses the latter in the CMake configuration (or the `bootstrap.sh` script), the internal logic configures the former. This feature is also much needed for the upcomig PR on coverage reports, where each backend is profiled in a separate job to avoid timeouts (due to profiling, compilers optimizations must be disabled, which greatly increases runtime). Hence, we decided to anticipate this separation as part of this PR, instead of hacking now a fix and then do the separation (and remove the hack).

## More thoughts
As a further thought on point 2., the fundamental problem is that unit tests run for *all* backends and the user (or the CI) has no way to select subsets (e.g., category==unit && backend==reference) when launching. This is a limitation of our current infrastructure, which will be solved once and for all in the next months with the introduction of CTests. With this infrastructure, a user will be able to enable potentially all backends and run only a specific subset of tests (down to a single one).

tests/unit/buildMatrixUnique.cpp
alberto-scolari: I recently opened an internal issue about this change, which may require some (long) investigation.
To summarize, the problem is that the tests fails *only* in these conditions (in logical `AND`):

1. bsp1d backend
2. 16 or more processes
3. these matrix sizes (not smaller)
4. within the CI container

This is not a memory problem, because I tested on a machine with more RAM than needed (and no limit on the container memory) and it still fails. Furthermore, I identified the point where the failure happens, which is quite deep into our backend logic and may thus be hard to reproduce (more details in the internal ticket).

Since we are using the `mpimsg` LPF backend, I suspect this is a problem of MPI + Docker, which may require trying multiple MPI versions (the container one is Ubuntu 20.04 default, an old MPICH 3.3.2). Also, only this test fails, and only in the aforementioned conditions, which made me create the separate, internal ticket.

tests/unit/CMakeLists.txt
alberto-scolari: This is caused by our "un-orthodox" testing modality: the result of `mxv_bsp1d` is compared against the result of `mxv_reference`, which is a "golden output". Therefore, `mxv_reference` must be compiled even when the reference backend was not selected for testing, hence the "force_" prefix (suggestions for a better naming are welcome!).

In other words, dependencies in the testing phase (output of `mxv_bsp1d` checked against output of `mxv_reference`) must be reflected into the compilation phase (compile `mxv_bsp1d` implies compile `mxv_reference`).

anyzelman: Yeah I got it, I meant just to compliment the approach=)

CMakeLists.txt
byjtew: Same as below 
Fixed [here](https://github.com/Algebraic-Programming/ALP/pull/200/commits/5b972d4890323e1b92237b8928a8cb4c468ac038)

alberto-scolari:
> Hi @alberto-scolari , sorry, I don't get how the two items you mention cause a linker error in the internal CI? I see things like
> 
> ```
> /usr/bin/ld: ../../src/graphblas/bsp1d/hybrid/libgraphblas.a(init.cpp.o): in function `grb::RC grb::finalize<(grb::Backend)7>()':
> /builds/zrc-von-neumann-lab/spatial-computing/graphblas/src/graphblas/bsp1d/init.cpp:87: undefined reference to `grb::RC grb::finalize<(grb::Backend)1>()'
> ```
> 
> Is the cause to this series of issues resolved by this PR? (If so, how?)

Problem 2. causes this error (and the other ones about compilation). In the CI, we excluded the reference_omp backend for LPF-related tests (because we don't want to run those tests both in `tests_unit` and in `tests_unit_lpf`, and possibly timeout), but in doing so we excluded not only the tests for the reference_omp backend, but also the reference_omp backend itself (`(grb::Backend)1`), on which the hybrid backend (`(grb::Backend)7`) depends; that's why it doesn't link. This is why this PR separates the backends it builds from the backends it tests: it builds *all* backends needed for the backends under test. This PR solves problem 2., hence these issues.

>
> Also I'm not aware of any test failing for bsp1d or hybrid -- which ones are those? (Or do you mean the ones that require some golden output?)

Indeed, those fail because the test producing the golden output (reference backend) is not built, and hence not run (you can see in the CI logs that the file with the golden output is missing). This PR:

1. fixes problem 2. by allowing the reference backend to be built, hence the reference test to be built
2. forces the build of "golden tests" via the `force_grb_executable()` you noticed
3. forces their run even if the reference backend is not tested (example at https://github.com/Algebraic-Programming/ALP/blob/5b972d4890323e1b92237b8928a8cb4c468ac038/tests/unit/unittests.sh#L667)

tests/unit/buildMatrixUnique.cpp
alberto-scolari: done https://github.com/Algebraic-Programming/ALP/pull/200/files#diff-da268222d5a4b946df12bb16fff811370487b63ce2adcde7f0f856dd50537076R602-R603

CMakeLists.txt
alberto-scolari: fixed https://github.com/Algebraic-Programming/ALP/commit/60cf3440307cc3a7214d8def74c9aa08db3f0294

anyzelman:
Got it, many thanks!!

Did some small code style fix, now waiting for CIs to OK before merge. Many thanks both

anyzelman (approved):

PR #204, opened 2023-06-21
636 eWiseApply([out] Matrix, [in] Matrix, [in] Matrix, ... ) variants

### LINKED ISSUE #203: Unexpected behaviour of eWiseApply([out] Matrix, [in] Matrix, [in] Matrix, ... ) variants
An unexpected behaviour using the two variants of __eWiseApply__ for matrices was noticed.

> [Documentation of the two variants](https://github.com/Algebraic-Programming/ALP/blob/54d340942d95e2cae6524c34a1bab890d287bed3/include/graphblas/base/blas3.hpp#L317)

- `eWiseApply([out] Matrix, [in] Matrix, [in] Matrix, [in] Monoid)` is supposed to apply the given monoid's operator over all elements of the two matrices, __including__ the {non_zero, zero} couples (_UNION_), using the provided monoid's identity as a replacement for the zero value.
- `eWiseApply([out] Matrix, [in] Matrix, [in] Matrix, [in] Operator)` is supposed to apply the given monoid's operator over all elements of the two matrices, __except__ for the {non_zero, zero} couples (_INTERSECTION_).

The observed behaviour showed that only the _INTERSECTION_ behaviour was implemented. A very user-friendly unit-test has been created for the occasion: __eWiseApplyMatrix_variants__.

Testing quickly:
```bash
make test_eWiseApplyMatrix_variants_debug_reference
./tests/unit/eWiseApplyMatrix_variants_debug_reference
```

Extra note: [This unit-test](https://github.com/Algebraic-Programming/ALP/blob/develop/tests/unit/eWiseApplyMatrixReference.cpp) should probably be re-evaluated.

byjtew:
Closes #203


byjtew:
Ready for review.

Implementation of grb::eWiseApply( C, A, B, Monoid ) and adaptation of the existing unit test finished.
The current implementation has been made according to the Operator variant code - hence not using OpenMP and provided without benchmarking data.

The implementation using OpenMP and a more efficient iteration pattern over matrices will be part of a future PR.

byjtew:
<h3>Benchmark - eWiseApply( C, A, B, op ):</h3>

*The prefix-sum is serial for the two implementations*

<ul>
<li><em>Stackless</em>:</li>
</ul>
<pre><code class="language-bash">./tests/unit/eWiseApplyMatrix_variants_ndebug_reference 200000000
This is functional test ./tests/unit/eWiseApplyMatrix_variants_ndebug_reference
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
Info: grb::Matrix&lt; T, reference &gt;::resize allocated 5 GB of memory, reason: , for 399999999 nonzeroes in an 200000000 times 200000000 matrix.
.
Info: grb::Matrix&lt; T, reference &gt;::resize freed (or will eventually free) 5 GB of memory, for 200000000 nonzeroes that this matrix previously contained.
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
-- eWiseApply using Operator, supposed to be annihilating non-zeroes -&gt; INTERSECTION
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
resize time: 344558238 ns
prefix sum time: 746306807 ns
apply time: 477102422 ns
Total execute time: 1223409229 ns
Result (operator) is correct
Test OK
</code></pre>
<ul>
<li><em>Stack</em>:</li>
</ul>
<pre><code class="language-bash">./tests/unit/eWiseApplyMatrix_variants_ndebug_reference 200000000
This is functional test ./tests/unit/eWiseApplyMatrix_variants_ndebug_reference
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
Info: grb::Matrix&lt; T, reference &gt;::resize allocated 5 GB of memory, reason: , for 399999999 nonzeroes in an 200000000 times 200000000 matrix.
.
Info: grb::Matrix&lt; T, reference &gt;::resize freed (or will eventually free) 5 GB of memory, for 200000000 nonzeroes that this matrix previously contained.
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
-- eWiseApply using Operator, supposed to be annihilating non-zeroes -&gt; INTERSECTION
Info: grb::Matrix&lt; T, reference &gt;::Matrix() allocated 9 GB of memory, reason: initial capacity allocation.
initialisation time: 420471744 ns
resize time: 1080189914 ns
Total resize: 1500661658 ns
initialisation time: 59941740 ns
prefix sum time: 623644765 ns
apply time: 1354553083 ns
Total execute time: 2038139588 ns
Result (operator) is correct
Test OK
</code></pre>

  | STACK | STACKLESS
-- | -- | --
RESIZE (ns) | 15 x 10^8 | 3.4 × 10^8
EXECUTE (ns) | 20 x 10^8 | 12 x 10^8
Prefix-sum / EXECUTE (%) | 30.60 | 61.00

aleksamilisavljevic:
If `A` and `B` are the same matrix, the following code from `eWiseApply_matrix_generic_union`:

```c++
			char * arr1, * arr2, * arr3, * buf1, * buf2, * buf3;
			arr1 = arr2 = buf1 = buf2 = nullptr;
			InputType1 * vbuf1 = nullptr;
			InputType2 * vbuf2 = nullptr;
			OutputType * valbuf = nullptr;
			internal::getMatrixBuffers( arr1, buf1, vbuf1, 1, A );
			internal::getMatrixBuffers( arr2, buf2, vbuf2, 1, B );
			internal::getMatrixBuffers( arr3, buf3, valbuf, 1, C );
			// end buffer retrieval

			// initialisations
			Coordinates< reference > coors1, coors2;
			coors1.set( arr1, false, buf1, n );
			coors2.set( arr2, false, buf2, n );
```

sets the same buffer for coors1 and coors2, which can be a problem if the descriptor `transpose_right` is set.

byjtew:
Right, I forgot to check if the IDs of the matrices were different, would you like to fix this @aleksamilisavljevic ?

byjtew:
> If `A` and `B` are the same matrix
> sets the same buffer for coors1 and coors2, which can be a problem if the descriptor `transpose_right` is set.

I just realised what you meant, that's a very good catch thanks. I'll push a fix in a minute

aleksamilisavljevic:
Seems good now, thanks

include/graphblas/bsp1d/blas3.hpp
```diff
@@ -131,38 +131,23 @@ namespace grb {
...
-		}
-		return internal::checkGlobalErrorStateOrClear( C, local_rc );
+		RC ret = eWiseApply< descr >(
+			internal::getLocal( C ),
+			internal::getLocal( A ),
+			internal::getLocal( B ),
+			mul,
+			phase
+		);
+		return internal::checkGlobalErrorStateOrClear( C, ret );
```
anyzelman: This rewrite changes the semantics: if the `RESIZE` fails, the semantics say nothing should happen to `C`. Would suggest to either revert or to pull the if( phase == RESIZE ) switch after the call to eWiseApply.

include/graphblas/bsp1d/blas3.hpp
```diff
@@ -179,29 +164,22 @@ namespace grb {
...
 			phase
 		);
-		if( phase == RESIZE ) {
-			if( collectives<>::allreduce( ret, operators::any_or< RC >() ) != SUCCESS ) {
-				return PANIC;
-			} else {
-				return SUCCESS;
-			}
-		}
-		assert( phase == EXECUTE );
```
anyzelman: I wanted to say "same as above" but actually the original implementation here is broken also: old line 201 should say `return ret`, not `return SUCCESS`

include/graphblas/nonblocking/blas3.hpp
```diff
@@ -535,39 +500,28 @@ namespace grb {
...
-			"called with an output matrix C that does not match the output domain "
-			"of the given multiplication operator"
-		);
-		static_assert( ( !(
-				std::is_same< InputType1, void >::value ||
-				std::is_same< InputType2, void >::value )
-			), "grb::eWiseApply (nonblocking, matrix <- matrix x matrix, operator): "
-			"the operator version of eWiseApply cannot be used if either of the "
-			"input matrices is a pattern matrix (of type void)"
-		);
```
anyzelman: Why removed? (For clarity of C++ errors, it is better that the user-facing functions immediately catch all static errors-- otherwise the C++ signature huge error traces are quickly encountered, lessening our user-friendliness)

include/graphblas/reference/blas3.hpp
```diff
@@ -57,6 +57,34 @@
...
+		const D *array, size_t i, const T&,
+		typename std::enable_if< !std::is_void< D >::value >::type * const = nullptr
+	) { return array[i]; }
+
+	template< typename T >
+	static T getValue( const void *, size_t, const T& identity ) { return identity; }
+
+} // namespace grb::internal
+
+#endif // _H_GRB_REFERENCE_BLAS3_ACCESSORS
```
anyzelman: Separate .hpp files or remove the ifndef/define/endif block

include/graphblas/reference/blas3.hpp
```diff
@@ -918,142 +946,108 @@ namespace grb {
...
+				#pragma omp parallel for simd default(none) \
+					shared(CCS_raw) firstprivate(n)
 #endif
-				for( size_t j = start; j < end; ++j ) {
+				for( size_t j = 0; j <= n; ++j ) {
 					CCS_raw.col_start[ j ] = 0;
 				}
-#ifdef _H_GRB_REFERENCE_OMP_BLAS3
 			}
-#endif
```
anyzelman: Two things here -- this switches from a static schedule to a dynamic one. The work here is totally balanced (meaning the work per iterand in the for-loop over `j` is the same), and thus the principle in ALP is to pick a static openmp schedule.

Second, `firstprivate' makes an (unnecessary) thread-local copy of the read-only `n`.

(The simd we don't usually employ -- we use the trick in `internalops.hpp` to guide the compiler into vectorisation. Indeed though here vectorisation is missing and I don't see an easy way to (mis/re)use the internalops.hpp tricks to otherwise effect vectorisation here, so I'm ok with adding simd there. A nicer fix may be to make setting-col_start-to-zero a primitive of the coordinates class)

include/graphblas/reference/blas3.hpp
```diff
@@ -918,142 +946,108 @@ namespace grb {
...
 			// end initialisations
 
 			// nonzero count
 			size_t nzc = 0;
 
 			// symbolic phase
 			if( phase == RESIZE ) {
+				nzc = 0;
```
anyzelman, line 1023: This happened a few lines above here as well?

include/graphblas/reference/blas3.hpp
```diff
@@ -1063,22 +1057,33 @@ namespace grb {
...
 				// retrieve additional buffer
-				config::NonzeroIndexType * const C_col_index = internal::template
-					getReferenceBuffer< typename config::NonzeroIndexType >( n + 1 );
+				auto* const C_col_index = getReferenceBuffer< config::NonzeroIndexType >( n + 1 );
+
+				if( !crs_only ) {
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+					#pragma omp parallel for simd default(none) \
+						shared(C_col_index) firstprivate(n)
+#endif
```
anyzelman: Nice catch! Please make it a static schedule and don't firstprivate n

include/graphblas/reference/blas3.hpp
```diff
@@ -1091,7 +1096,9 @@ namespace grb {
 						const size_t l_col = B_raw.row_index[ l ];
 						if( coors1.assigned( l_col ) ) {
 							(void) ++nzc;
-							(void) ++CCS_raw.col_start[ l_col + 1 ];
+							if( !crs_only ) {
```
anyzelman, line 1072: nice catch!

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+					coors1.clear();
+
 #ifdef _H_GRB_REFERENCE_OMP_BLAS3
-				#pragma omp parallel
-				{
-					size_t start, end;
-					config::OMP::localRange( start, end, 0, n );
+					#pragma omp parallel default(none) \
+							shared(coors1, valbuf) \
+							firstprivate(i, A_raw, dummy_identity)
```
anyzelman: again firstprivate (there are all read-only vars right)?

include/graphblas/reference/blas3.hpp
anyzelman: btw if the openmp default is static (initially I thought dynamic, but in hindsight and a little bit later I'm not sure), please still make it explicitly static to avoid ambiguity. Also note that by using `OMP::localRange`, ALP also controls exactly how the static schedule is derived, which I find preferable. Localrange also takes into account false sharing issues.

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+						// update CCS
+						if( !crs_only ) {
+							++C_col_index[ j ];
+							const size_t CCS_index =  CCS_raw.col_start[ j+1 ] - C_col_index[ j ];
+							CCS_raw.row_index[ CCS_index ] = i;
+							CCS_raw.setValue( CCS_index, result_value );
+						}
+
+						// update count
+						(void)++nzc;
```
anyzelman: space before `++`

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+
+					CRS_raw.col_start[ i + 1 ] = nzc;
+
+				}
+
+
+#ifdef _DEBUG
+				std::cout << "CRS_raw.col_start = [ ";
+				for( size_t j = 0; j <= m; ++j )
+					std::cout << CRS_raw.col_start[ j ] << " ";
```
anyzelman: curly brackets even if the inner code block is a single statement (elsewhere below also)

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+			// Note: By using the buffer of the output matrix C, we can
+			//       allow A and B to be the same matrix (with the same buffer)
+			Coordinates< reference > coors1, coors2;
+			coors1.set( arr1, false, buf1, n );
+			coors2.set( arr3, false, buf3, n );
+			// end initialisations of the coordinates
+
+			if( !crs_only ) {
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+				#pragma omp parallel for simd default(none) shared(CCS_raw) firstprivate(n)
```
anyzelman: make it (explicitly) statically scheduled (suggest to use localRange), no firstprivate on read-only n

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+			// end initialisations
+
+			// nonzero count
+			size_t nzc = 0;
+
+			// symbolic phase
+			if( phase == RESIZE ) {
+				nzc = 0;
+
+
```
anyzelman: remove one superfluous empty line

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+				{
+					for( size_t i = 0; i < m; ++i ) {
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+						#pragma omp single
+#endif
+						{
+							coors1.clear();
+						}
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+						#pragma omp barrier
```
anyzelman: I would expect it to be cheaper to start the parallel section at line 1378, rather than using a single for the clear (also does the single not imply a barrier at the end of it?)

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+						for( size_t k = A_raw.col_start[ i ]; k < A_raw.col_start[ i + 1 ]; ++k ) {
+							const size_t k_col = A_raw.row_index[ k ];
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+							if( !coors1.asyncAssign( k_col, local_update1 ) ) {
+                                assignValue( vbuf1, k_col, A_raw.getValue( k, identity_A ) );
+                                if( ++assigns1 == maxAsyncAssigns1 ) {
+                                    coors1.joinUpdate( local_update1 );
+                                    assigns1 = 0;
+                                }
+                            }
```
anyzelman: spaces instead of tabs for identation

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+#else
+							(void)coors1.assign( k_col );
```
anyzelman: space before coors1 (and directly below also)

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+						}
+					}
 					for( size_t l = B_raw.col_start[ i ]; l < B_raw.col_start[ i + 1 ]; ++l ) {
 						const size_t l_col = B_raw.row_index[ l ];
-						if( coors1.assigned( l_col ) ) {
-							coors2.assign( l_col );
-							(void)grb::apply( valbuf[ l_col ], valbuf[ l_col ], B_raw.getValue( l,
-								mulMonoid.template getIdentity< typename Operator::D2 >() ), oper );
+						if( !coors1.assigned( l_col ) ) {
+							(void)++nzc;
```
anyzelman: space before `++`

anyzelman: and directly below also

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+#ifndef NDEBUG
+					assert( CCS_raw.col_start[ 0 ] == 0 );
+					assert( CCS_raw.col_start[ n ] == nzc );
+#endif
+				}
+
+				// set C_col_index to all zero
+				if( !crs_only ) {
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+					#pragma omp parallel for simd
```
anyzelman, line 1351: explicitly select static schedule with right blocksize (use of OMP::localRange is preferred)

(simd can stay as per above also)

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+				nzc = 0;
+				CRS_raw.col_start[ 0 ] = 0;
+				for( size_t i = 0; i < m; ++i ) {
+					coors1.clear();
+					coors2.clear();
+
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+				#pragma omp parallel default(none) \
+						shared(coors1, vbuf1, coors2, vbuf3) \
+						firstprivate(i, A_raw, identity_A, B_raw, identity_B )
```
anyzelman: read-only?

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+							if( !coors1.asyncAssign( k_col, local_update1 ) ) {
+								if( ++assigns1 == maxAsyncAssigns1 ) {
+									coors1.joinUpdate( local_update1 );
+									assigns1 = 0;
+								}
+							}
+#else
+							(void)coors1.assign( k_col );
```
anyzelman: missing space (I'll stop flagging these, suggest to use global find/replace)

include/graphblas/reference/blas3.hpp
```diff
@@ -1111,92 +1118,545 @@ namespace grb {
...
+
+					CRS_raw.col_start[ i + 1 ] = nzc;
+				}
+
+				if( !crs_only ) {
 #ifdef _DEBUG
-					std::cout << "\n";
+					std::cout << "CRS_raw.col_start = [ ";
+					for( size_t j = 0; j <= m; ++j )
+						std::cout << CRS_raw.col_start[ j ] << " ";
```
anyzelman: curly-brackets also for one-liner bodies, see below also

include/graphblas/reference/blas3.hpp
```diff
@@ -1282,23 +1743,27 @@ namespace grb {
...
+		const typename std::enable_if<
+				!grb::is_object< OutputType >::value &&
+				!grb::is_object< InputType1 >::value &&
+				!grb::is_object< InputType2 >::value &&
+				grb::is_operator< Operator >::value,
+				void
+			>::type * const = nullptr
 	) {
+		typedef typename std::conditional<std::is_void<InputType1>::value, typename Operator::D1, InputType1>::type ActualInputType1;
+		typedef typename std::conditional<std::is_void<InputType2>::value, typename Operator::D2, InputType1>::type ActualInputType2;
```
anyzelman: line length overflow

include/graphblas/reference/blas3.hpp
```diff
@@ -1282,23 +1743,27 @@ namespace grb {
...
 		// static checks
 		NO_CAST_ASSERT( ( !( descr & descriptors::no_casting ) ||
-			std::is_same< typename Operator::D1, InputType1 >::value ),
+			std::is_same< typename Operator::D1, ActualInputType1 >::value ),
```
anyzelman, line 1567: I'm not sure what is being fixed here -- is it now not the case that `void`  A passes no_casting even if the monoid left-hand input is not void? (Would that not be unexpected behaviour?)

include/graphblas/reference/blas3.hpp
```diff
@@ -1282,23 +1743,27 @@ namespace grb {
...
 			"grb::eWiseApply (reference, matrix <- matrix x matrix, operator)",
 			"called with a prefactor input matrix A that does not match the first "
 			"domain of the given multiplication operator"
 		);
 		NO_CAST_ASSERT( ( !( descr & descriptors::no_casting ) ||
-			std::is_same< typename Operator::D2, InputType2 >::value ),
+			std::is_same< typename Operator::D2, ActualInputType2 >::value ),
```
anyzelman, line 1573: same here

include/graphblas/reference/blas3.hpp
```diff
@@ -1309,20 +1774,17 @@ namespace grb {
...
-		static_assert( ( !(
-				std::is_same< InputType1, void >::value ||
-				std::is_same< InputType2, void >::value )
-			), "grb::eWiseApply (reference, matrix <- matrix x matrix, operator): "
-			"the operator version of eWiseApply cannot be used if either of the "
-			"input matrices is a pattern matrix (of type void)"
+		static_assert(
+			!std::is_void< OutputType >::value,
+			"grb::eWiseApply: the elementwise mxm cannot be used if the"
+			" output matrix is a pattern matrix (of type void)"
```
anyzelman: isn't the description wrong here? The issue is if the output is non-void, while both inputs are void (in the operator case at least)

tests/unit/eWiseApply_matrix.cpp
```diff
@@ -94,4 +94,3 @@ int main( int argc, char ** argv ) {
 		return 0;
 	}
 }
-
```
anyzelman, line 97: revert?

anyzelman (changes requested):
Initial review (note to self: did not review unit tests yet)

include/graphblas/reference/blas3.hpp
byjtew: It's just to select which of the types will be used: 
* in a void case it will be the operator input type
* in a non-void case it will be the one of the matrix

tests/unit/eWiseApply_matrix.cpp
byjtew: ?

anyzelman:
Provides union variants (intersection variant already existed but was erroneously called for monoid variants as well). This one is ready for merge also (aside from review & conflicts)

PR #214, opened 2023-06-28
643 Provide tril([out] Matrix, [in] Matrix, int, Phase)

### LINKED ISSUE #213: Provide tril([out] Matrix, [in] Matrix, int, Phase)
Allocate and gather the lower/upper triangular portion of a matrix

Parameters:
* L / U: Output matrix
  The lower/upper triangular portion of A in sparse format.
* A: Input matrix
  Matrix to extract the triangular portion from.
* k: integer (optional)
  The topmost diagonal of the triangular portion to extract

Signature of __tril(L, A, k, phase)__ (_triu(L, A, k, phase)_ is implicit)
```c++
/**
 * Return the lower triangular portion of a matrix, strictly below the diagonal
 * starting at the k-th column (included).
 * 
 * Out-of-place variant.
 *
 * @tparam descr      The descriptor to be used (descriptors::no_operation
 * 					  if left unspecified).
 * @tparam InputType  The type of the elements in the supplied ALP/GraphBLAS
 *                    matrix \a A.
 * @tparam OutputType The type of the elements in the supplied ALP/GraphBLAS
 *                    matrix \a L.
 *
 * @param[out] L       The lower triangular portion of \a A, strictly below
 * 					   the k-th diagonal.
 * @param[in]  A       Any ALP/GraphBLAS matrix.
 * @param[in]  k       The smallest column of the diagonal above which to zero out \a A.
 * @param[in]  phase   The #grb::Phase in which the primitive is to proceed.
 *
 * @return grb::SUCCESS  When the call completed successfully.
 * @return grb::MISMATCH If the dimensions of \a L and \a A do not match.
 * 
 * \parblock
 * \par Parameter \a k
 * If you want to extract the strict lower triangular portion of a matrix \a A 
 * (without its main diagonal), you can use the following call:
 *        grb::tril( L, A, 1 );
 * \par Extract the main diagonal
 * If you want to extract the main diagonal of \a A, 
 * you can use the following calls:
 * 			grb::triu( U, A, 0 );
 *        	grb::tril( L, U, 0 );
 * \endparblock
 *
 * \parblock
 * \par Allowed descriptors
 * - transpose_matrix: Consider A^T instead of A.
 * - no_casting: If the types of \a L and \a A differ, the primitive
 * 				 will fail.
 * \endparblock
 */
template<
	typename InputType,
	typename OutputType,
	typename RIT_L, typename CIT_L, typename NIT_L,
	typename RIT_A, typename CIT_A, typename NIT_A,
	Backend implementation
>
RC tril(
	Matrix< OutputType, implementation, RIT_L, CIT_L, NIT_L > & L,
	const Matrix< InputType, implementation, RIT_A, CIT_A, NIT_A > & A,
	const long int k,
	const Phase &phase = Phase::EXECUTE,
		!grb::is_object< OutputType >::value &&
		!grb::is_object< InputType >::value &&
		std::is_convertible< InputType, OutputType >::value
	>::type * const = nullptr
) {
	(void) L;
	(void) A;
	(void) k;
	(void) phase;
#ifdef _DEBUG
	std::cerr << "Selected backend does not implement grb::tril()\n";
#endif
#ifndef NDEBUG
	const bool selected_backend_does_not_support_tril = false;
	assert( selected_backend_does_not_support_tril );
#endif
	const RC ret = grb::clear( L );
	return ret == SUCCESS ? UNSUPPORTED : ret;
}

/**
 * Return the lower triangular portion of a matrix, strictly below the diagonal
 * starting at the k-th column (included).
 * 
 * Out-of-place variant.
 * 
 * This primitive is strictly equivalent to calling
 * grb::tril( L, A, 0, phase ).
 *
 * see grb::tril( L, A, k, phase ) for full description.
 */
template<
	typename InputType,
	typename OutputType,
	typename RIT_L, typename CIT_L, typename NIT_L,
	typename RIT_A, typename CIT_A, typename NIT_A,
	Backend implementation
>
RC tril(
	Matrix< OutputType, implementation, RIT_L, CIT_L, NIT_L > & L,
	const Matrix< InputType, implementation, RIT_A, CIT_A, NIT_A > & A,
	const Phase &phase = Phase::EXECUTE,
		!grb::is_object< OutputType >::value &&
		!grb::is_object< InputType >::value &&
		std::is_convertible< InputType, OutputType >::value
	>::type * const = nullptr
) {
	return tril< descr >( L, A, 0, phase );
}
```

### LINKED ISSUE #230: Utils CRS and CCS print

byjtew:
Closes #213

~Requires #231~
~Requires #241~

Generalisation of this feature in #271

### DIFF
--- a/include/graphblas/base/blas3.hpp
+++ b/include/graphblas/base/blas3.hpp
@@ -490,6 +490,7 @@ namespace grb {
 	) {
 		(void) L;
 		(void) A;
+		(void) k;
 		(void) phase;
 #ifdef _DEBUG
 		std::cerr << "Selected backend does not implement grb::tril()\n";
@@ -530,6 +531,97 @@ namespace grb {
 		>::type * const = nullptr
 	) {
 		return tril< descr >( L, A, 0, phase );
+	}
+
+	/**
+	 * Return the upper triangular portion of a matrix, strictly above the k-th 
+	 * diagonal (excluded).
+	 *
+	 * @tparam descr      The descriptor to be used (descriptors::no_operation
+	 * 					  if left unspecified).
+	 * @tparam InputType  The type of the elements in the supplied ALP/GraphBLAS
+	 *                    matrix \a A.
+	 * @tparam OutputType The type of the elements in the supplied ALP/GraphBLAS
+	 *                    matrix \a U.
+	 *
+	 * @param[out] U       The upper triangular portion of \a A, strictly above 
+	 * 					   the k-th diagonal.
+	 * @param[in]  A       Any ALP/GraphBLAS matrix.
+	 * @param[in]  k       The diagonal above which to zero out \a A.
+	 * @param[in]  phase   The #grb::Phase in which the primitive is to proceed.
+	 *
+	 * @return grb::SUCCESS  When the call completed successfully.
+	 * @return grb::MISMATCH If the dimensions of \a U and \a A do not match.
+	 *
+ 	 * \parblock
+	 * \par Allowed descriptors
+	 * - transpose_matrix: Consider A^T instead of A.
+	 * - no_casting: If the types of \a T and \a A differ, the primitive
+	 * 				 will fail.
+	 * \endparblock
+	 */
+	template<
+		Descriptor descr = descriptors::no_operation,
+		typename InputType,
+		typename OutputType,
+		typename RIT_U, typename CIT_U, typename NIT_U,
+		typename RIT_A, typename CIT_A, typename NIT_A,
+		Backend implementation
+	>
+	RC triu(
+		Matrix< OutputType, implementation, RIT_U, CIT_U, NIT_U > & U,
+		const Matrix< InputType, implementation, RIT_A, CIT_A, NIT_A > & A,
+		const long int k,
+		const Phase &phase = Phase::EXECUTE,
+		const typename std::enable_if<
+			!grb::is_object< OutputType >::value &&
+			!grb::is_object< InputType >::value &&
+			std::is_convertible< InputType, OutputType >::value
+		>::type * const = nullptr
+	) {
+		(void) U;
+		(void) A;
+		(void) k;
+		(void) phase;
+#ifdef _DEBUG
+		std::cerr << "Selected backend does not implement grb::triu()\n";
+#endif
+#ifndef NDEBUG
+		const bool selected_backend_does_not_support_triu = false;
+		assert( selected_backend_does_not_support_triu );
+#endif
+		const RC ret = grb::clear( A );
+		return ret == SUCCESS ? UNSUPPORTED : ret;
+	}
+
+	/**
+	 * Return the upper triangular portion of a matrix,
+	 * strictly above main diagonal (excluded).
+	 *
+	 * This primitive is strictly equivalent to calling
+	 * grb::triu( U, A, 0, phase ) 
+	 * 
+	 * see grb::triu( U, A, k, phase ) for full description.
+	 */
+	template<
+		Descriptor descr = descriptors::no_operation,
+		typename InputType,
+		typename OutputType,
+		typename RIT_U, typename CIT_U, typename NIT_U,
+		typename RIT_A, typename CIT_A, typename NIT_A,
+		Backend implementation
+	>
+	RC triu(
+		Matrix< OutputType, implementation, RIT_U, CIT_U, NIT_U > & U,
+		const Matrix< InputType, implementation, RIT_A, CIT_A, NIT_A > & A,
+		const Phase &phase = Phase::EXECUTE,
+		const typename std::enable_if<
+			!grb::is_object< OutputType >::value &&
+			!grb::is_object< InputType >::value &&
+			std::is_convertible< InputType, OutputType >::value
+		>::type * const = nullptr
+	) {
+		return triu< descr >( U, A, 0, phase );
 	}
 
 	/**
--- a/include/graphblas/base/blas3.hpp
+++ b/include/graphblas/base/blas3.hpp
@@ -472,12 +472,13 @@ namespace grb {
 		Descriptor descr = descriptors::no_operation,
 		typename InputType,
 		typename OutputType,
-		typename RIT, typename CIT, typename NIT,
-		Backend backend
+		typename RIT_L, typename CIT_L, typename NIT_L,
+		typename RIT_A, typename CIT_A, typename NIT_A,
+		Backend implementation
 	>
 	RC tril(
-		Matrix< OutputType, backend, RIT, CIT, NIT > &L,
-		const Matrix< InputType, backend, RIT, CIT, NIT > &A,
+		Matrix< OutputType, implementation, RIT_L, CIT_L, NIT_L > & L,
+		const Matrix< InputType, implementation, RIT_A, CIT_A, NIT_A > & A,
 		const long int k,
 		const Phase &phase = Phase::EXECUTE,
 		const typename std::enable_if<
@@ -512,12 +513,13 @@ namespace grb {
 		Descriptor descr = descriptors::no_operation,
 		typename InputType,
 		typename OutputType,
-		typename RIT, typename CIT, typename NIT,
-		Backend backend
+		typename RIT_L, typename CIT_L, typename NIT_L,
+		typename RIT_A, typename CIT_A, typename NIT_A,
+		Backend implementation
 	>
 	RC tril(
-		Matrix< OutputType, backend, RIT, CIT, NIT > &L,
-		const Matrix< InputType, backend, RIT, CIT, NIT > &A,
+		Matrix< OutputType, implementation, RIT_L, CIT_L, NIT_L > & L,
+		const Matrix< InputType, implementation, RIT_A, CIT_A, NIT_A > & A,
 		const Phase &phase = Phase::EXECUTE,
 		const typename std::enable_if<
 			!grb::is_object< OutputType >::value &&
--- a/include/graphblas/base/blas3.hpp
+++ b/include/graphblas/base/blas3.hpp
@@ -443,7 +443,8 @@ namespace grb {
 	}
 
 	/**
-	 * Return the lower triangular portion of a matrix, below the k-th diagonal.
+	 * Return the lower triangular portion of a matrix, strictly below the k-th 
+	 * diagonal (excluded).
 	 *
 	 * @tparam descr      The descriptor to be used (descriptors::no_operation
 	 * 					  if left unspecified).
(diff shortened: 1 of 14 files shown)


include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,220 @@ namespace grb {
...
 	}
 
+	/**
+	 * Return the lower triangular portion of a matrix, strictly below the diagonal
+	 * starting at the k-th column (included).
+	 * 
+	 * Out-of-place variant.
+	 *
+	 * @tparam descr      The descriptor to be used (descriptors::no_operation
+	 * 					  if left unspecified).
```
anyzelman, line 452: funky spacing here (maybe tabs instead of spaces? Elsewhere also)

include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,220 @@ namespace grb {
...
+		typename InputType,
+		typename OutputType,
+		typename RIT_L, typename CIT_L, typename NIT_L,
+		typename RIT_A, typename CIT_A, typename NIT_A,
+		Backend implementation
+	>
+	RC tril(
+		Matrix< OutputType, implementation, RIT_L, CIT_L, NIT_L > &L,
+		const Matrix< InputType, implementation, RIT_A, CIT_A, NIT_A > &A,
+		const long int k,
```
anyzelman, line 497: Use `size_t` instead? (There is no reason to limit the range of `k`, while `size_t` will always be the maximum range addressable)

include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,220 @@ namespace grb {
...
+	 * 					  if left unspecified).
+	 * @tparam InputType  The type of the elements in the supplied ALP/GraphBLAS
+	 *                    matrix \a A.
+	 * @tparam OutputType The type of the elements in the supplied ALP/GraphBLAS
+	 *                    matrix \a L.
+	 *
+	 * @param[out] L       The lower triangular portion of \a A, strictly below
+	 * 					   the k-th diagonal.
+	 * @param[in]  A       Any ALP/GraphBLAS matrix.
+	 * @param[in]  k       The smallest column of the diagonal above which to zero out \a A.
```
anyzelman, line 461: Can `k` be negative?

include/graphblas/bsp1d/blas3.hpp
```diff
@@ -195,16 +195,118 @@ namespace grb {
...
+#ifdef _DEBUG
+		std::cout << "In grb::tril( BSP1D )\n";
+#endif
+		assert( phase != TRY );
+		const RC ret = tril< descr >(
+			internal::getLocal( L ),
+			internal::getLocal( A ),
+			k,
+			phase
+		);
```
anyzelman, line 236: Simply delegating has an issue that local coordinates may first need to be mapped to global coordinates before we may determine what is above or below the kth diagonal. Here we are using a row-wise block-cyclic distribution, which means local rows have a hidden offset-- e.g., the local row 64 at the first user process actually corresponds to row 128 globally, and therefore it should filter column indices greater-than-or-equal to 128, not 64.

Some primitives like the reference `vxm` for this reason take a hidden `std::function` parameter that allows the reference backend to execute the necessary conversions. An alternative could be a hidden template argument that passes the appropriate `Distribution` class.

A different implementation may instead rely on iterators -- we could iterate over the nonzeroes of `A` in `PARALLEL` I/O mode, meaning we iterate only over the contents of `A` that are locally stored. Iterators already convert local indices to global indices. Thus we can filter the resulting iterator (`operator++` can skip elements that are above the `k`th diagonal), and pass the filtered iterator to a `buildMatrix` routine on `L`. With this type of approach, there are two remarks:

1. `tril` and `triu` could reside in `grb::algorithms`, which frees up new backends from having to implement `tril` and `triu`explicitly (they just have to implement parallel output iteration and matrix ingestion from iterators, which they would have to anyway);

2. however, currently, output matrix iterators are forward iterators, not random access -- we would have to fix that first.

include/graphblas/bsp1d/blas3.hpp
```diff
@@ -195,16 +195,118 @@ namespace grb {
...
+#endif
+		assert( phase != TRY );
+		const RC ret = triu< descr >(
+			internal::getLocal( U ),
+			internal::getLocal( A ),
+			k,
+			phase
+		);
+
+		return internal::checkGlobalErrorStateOrClear( U, ret );
```
anyzelman, line 288: An alternative way of implementing triu might be to call tril with a `transpose` descriptor?

include/graphblas/nonblocking/blas3.hpp
```diff
@@ -571,6 +571,110 @@ namespace grb {
...
+		const typename std::enable_if<
+			!grb::is_object< OutputType >::value &&
+			!grb::is_object< InputType >::value &&
+			std::is_convertible< InputType, OutputType >::value
+		>::type * const = nullptr )
+	{
+#ifdef _DEBUG
+		std::cout << "In grb::tril( nonblocking )\n";
+#endif
+		// nonblocking execution is not supported
```
anyzelman, line 594: please print a warning whenever a blocking placeholder implementation is used (as done e.g. in the present "nonblocking" mxm)

include/graphblas/reference/blas3.hpp
```diff
@@ -1204,6 +1204,210 @@ namespace grb {
...
+			}
+
+			const auto &A_raw = descr & descriptors::transpose_matrix
+								? internal::getCCS( A )
+								: internal::getCRS( A );
+
+			if( phase == Phase::RESIZE ) {
+				size_t nzc = 0;
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+#pragma omp parallel for reduction( + : nzc ) \
```
anyzelman, line 1259: This pragma can be aligned with the `for` on line 1262 it parallelises

include/graphblas/reference/blas3.hpp
```diff
@@ -1204,6 +1204,210 @@ namespace grb {
...
+			Descriptor descr = descriptors::no_operation,
+			typename InputType, typename OutputType,
+			typename RIT_L, typename CIT_L, typename NIT_L,
+			typename RIT_A, typename CIT_A, typename NIT_A
+		>
+		RC trilu_generic(
+			Matrix< OutputType, reference, RIT_L, CIT_L, NIT_L > &L,
+			const Matrix< InputType, reference, RIT_A, CIT_A, NIT_A > &A,
+			const long int k,
+			const Phase &phase
```
anyzelman, line 1218: if taking a lambda-function approach to resolving local vs. global indexing, this is probably where it should go. I'll hold off a deep review on this function until it's clear whether or not such a thing will be added=)

include/graphblas/reference/blas3.hpp
```diff
@@ -1204,6 +1204,210 @@ namespace grb {
...
+				L_crs_raw.col_start[ 0 ] = 0;
+				L_ccs_raw.col_start[ 0 ] = 0;
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+#pragma omp parallel for simd
+#endif
+				for( size_t i = 0; i < m; ++i ) {
+					L_ccs_raw.col_start[ i + 1 ] = 0;
+				}
+
+				// Prefix sum computation into L.CRS.col_start
```
anyzelman, line 1300: Is this similar to the various other prefix sums elsewhere in the code base (e.g., `reference/matrix.hpp`? At this point it is probably wise to have like a single standard function for it in `graphblas/utils` e.g.

anyzelman (changes requested):
A couple of high-level questions, if you could have a look. The tests look very nice to me, although in particular I worry about why this would work in the `BSP1D` or `hybrid` cases. Might it be that the test does not check for missing nonzeroes (i.e., if tril/triu filter too many nonzeroes)?

include/graphblas/base/blas3.hpp
byjtew: k can be negative, as it represents an offset (+/-)

include/graphblas/bsp1d/blas3.hpp
byjtew: Already the case if you have a look at the _referenec_ implementation: `RC trilu_generic(...)`

include/graphblas/reference/blas3.hpp
byjtew: Already working on that 👍

include/graphblas/bsp1d/blas3.hpp
byjtew: The thing is that we have to be very clear on what we designate with the word `transpose`:
- If we talk about the input matrix **only**, then `tril` is not equivalent to `triu<transpose>`
e.g 
```
<<< tril<transpose_matrix>(A) >>>
=== MATRIX ===
Size: 10 x 10
0  0  0  0  0  0  0  0  0  0
0  0  0  0  0  0  0  0  0  0
0  0  0  0  0  0  0  0  0  0
3 0  0  9 0  0  0  0  0  0
0  0  0  0  0  0  0  0  0  0
0  0  0  11 13 0  0  0  0  0
6 0  0  0  0  0  18 0  0  0
0  9 0  0  0  0  0  0  0  0
0  0  0  0  0  0  0  0  0  0
0  0  0  0  0  0  0  23 0  0
==============

<<< triu(A) >>>
=== MATRIX ===
Size: 10 x 10
0  0  0  3 0  0  6 0  0  0
0  0  0  0  0  0  0  9 0  0
0  0  0  0  0  0  0  0  0  0
0  0  0  9 0  11 0  0  0  0
0  0  0  0  0  13 0  0  0  0
0  0  0  0  0  0  0  0  0  0
0  0  0  0  0  0  18 0  0  0
0  0  0  0  0  0  0  0  0  23
0  0  0  0  0  0  0  0  0  0
0  0  0  0  0  0  0  0  0  0
==============
```

- If we talk about the input and the output matrix, then `tril` **is** equivalent to `triu<transpose>`. I am not exactly sure how to formally describe such logic using the available descriptors

include/graphblas/base/blas3.hpp
byjtew: Indeed it can be it's an offset

include/graphblas/bsp1d/blas3.hpp
byjtew: I will try to come up with something for the local coordinates issue.

I am not really prone to the idea of transforming `tril+u` into an algorithm.
This would first of all hamper any backend-specific optimisations. 
Moreover, this would not make sense if we plan to generalise `tril+u` into a `filter` primitive using special (boolean) operators capable of handling vector/matrix coordinates (like in other GraphBLAS implementations)

include/graphblas/bsp1d/blas3.hpp
byjtew: Forget the comment above I was only considering square matrices, plus it's a twisted logic that is absolutely not necessary

anyzelman:
If we have have select in v0.9 (#271) then this issue should be resolved as an algorithm (maybe `grb::algorithms::matrices<>::tril` or `grb::algorithms::tril`).

anyzelman:
Given we have #271 slated for v0.8, should this MR still be retained actually?

byjtew:
> Given we have #271 slated for v0.8, should this MR still be retained actually?

I think it should be discarded

PR #215, opened 2023-07-05
Ctest infrastructure

alberto-scolari:
This PR replaces the current testing infrastructure with a new one based on CTest, with the following main advantages:

* generation of test binaries and test cases in a single file and on top of the same infrastructure (CMake), hence no coherence issues anymore: e.g., if you rename a test executable but not the corresponding test cases, you will get an error during CMake configuration
* more abstract API to generate test cases: you specify the list of backends, the infrastructure does the rest
* ability to filter tests to run from the command line along to multiple dimensions (backends, categories, name via regex): you can choose each time which tests to run, down to a single one; the infrastructure figures out the rest (launcher, environment variables, ...)
* better reporting, with commands reported during run (to make it easy to reproduce it manually); you can easily re-run at each level (test, launcher, binary) just by copy-and-paste'ing from the test log
* integration with internal CI: CTest can export a test in JUnit format, which the CI can understand and display in the GUI; this standard file format is also easy to parse with common tools like Python (which includes a standard library with XQuery support)

As a further note, this PR removes more code than it adds, although adding more features: +2024, -2848. This mostly comes from removing the scripts invoking each test (documentation and golden outputs not accounted: `git diff develop..HEAD --stat '*.cpp' '*.hpp' '*.sh' '*.cmake' '*.txt' '*.py' '*.yml' '*.in'`).

The old infrastructure is indeed removed in a dedicated commit ("removing old infrastructure based on Bash test runners), which is a decision we may revert or at least we may want to remember (for future debugging) via a dedicated git tag.

Commits sequence:
* commits 1-4: fixes of bugs found along the way
* commits 5-13: CTest infrastructure and its use all over
* commit 14: removing old test infrastructure


byjtew:
To investigate:

image

byjtew (changes requested):

tests/unit/utilParserTest.cpp
```diff
@@ -23,13 +23,11 @@
...
-
 	if( argc != 2 ) {
-		std::cerr << "please, give path to west0497.mtx" << std::endl;
-		std::cout << "Test FAILED" << std::endl;
-		return 255;
+		std::cerr << "path to input file west0497.mtx is needed" << std::endl;
+		std::exit( 1 );
 	}
-
+
```
byjtew, line 30:
```suggestion

```

byjtew (changes requested):

tests/unit/utilParserTest.cpp
alberto-scolari: should be fixed in the latest changes

alberto-scolari:
> To investigate:
> 
> image

Coverage issue fixed, now slightly better
coverage

PR #224, opened 2023-07-13
Provide operators::logical_not

### LINKED ISSUE #222: Provide operators::negate
Operator wrapper allowing to negate its result.

The point of this feature it to not need to fully implement new operators using operators::negate. 
For instance _operators::logical\_nand_ becomes: 
-  _operators::negate< operators::logical_and < T > >_
- OR  _operators::not_op< operators::logical_and < T > >_

byjtew:
closes #222


include/graphblas/base/internalops.hpp
```diff
@@ -39,9 +39,127 @@ namespace grb {
 
 	namespace operators {
 
+
+
```
anyzelman, line 43: Too many empty lines?

include/graphblas/base/internalops.hpp
```diff
@@ -39,9 +39,127 @@ namespace grb {
...
+			 * 			- \b D2: The right-hand input domain.
+			 * 			- \b D3: The output domain, must be convertible to bool.
+			 * 			- \b operator_type: The internal::operator type to negate.
+			 */
+			template<
+				class Op,
+				enum Backend implementation = config::default_backend
+			>
+			class logical_not {
+				public:
```
anyzelman, line 65: missing empty line after class def (elsewhere also)

include/graphblas/base/internalops.hpp
```diff
@@ -39,9 +39,127 @@ namespace grb {
...
+
+					/** Whether this operator has an inplace foldr. */
+					static constexpr bool has_foldr = Op::operator_type::has_foldr;
+
+					/**
+					 * Whether this operator is \em mathematically associative; that is,
+					 * associative when assuming equivalent data types for \a IN1, \a IN2,
+					 * and \a OUT, as well as assuming exact arithmetic, no overflows, etc.
+					 */
+					static constexpr bool is_associative = Op::operator_type::is_associative;
```
anyzelman, line 87: Associativity of the underlying operator is actually not sufficient for the negated operator to be associative -- counterexample:
```
!(!(a or b) or c)=(a or b) and !c=(a and !c) or (b and !c)
while
!(a or !(b or c))=!a and (b or c)=(!a and b) or (!a and c)
```
So e.g., a = true, b=false, and c = false leads to a counter-example.

include/graphblas/base/internalops.hpp
```diff
@@ -39,9 +39,127 @@ namespace grb {
...
+
+					/**
+					 * Whether this operator is \em mathematically commutative; that is,
+					 * commutative when assuming equivalent data types for \a IN1, \a IN2,
+					 * and \a OUT, as well as assuming exact arithmetic, no overflows, etc.
+					 */
+					static constexpr bool is_commutative = Op::operator_type::is_commutative;
```
anyzelman, line 94: By contrast, if the underlying operator is commutative then for sure the negated operator is commutative also, simply because no matter the order of operands, the negation is applied to the same value. Given the non-triviality of the associative case, it would be good to note why commutativity is "inherited" in the doxygen here.

include/graphblas/base/internalops.hpp
```diff
@@ -39,9 +39,127 @@ namespace grb {
...
+					static void apply(
+						const left_type * __restrict__ const a,
+						const right_type * __restrict__ const b,
+						result_type * __restrict__ const c,
+						const typename std::enable_if<
+							std::is_convertible< result_type, bool >::value,
+							void
+						>::type * = nullptr
+					) {
+
```
anyzelman, line 114: superfluous newline (elsewhere also)

include/graphblas/base/internalops.hpp
```diff
@@ -39,9 +39,127 @@ namespace grb {
...
+						Op::operator_type::foldl( c, b );
+						*c = !*c;
+
+					}
+			};
```
anyzelman, line 161: missing empty line before `};`

include/graphblas/ops.hpp
```diff
@@ -1142,6 +1166,11 @@ namespace grb {
 		static const constexpr bool value = true;
 	};
 
+	template< class Op >
+	struct is_idempotent< operators::logical_not< Op >, void > {
+		static const constexpr bool value = is_idempotent< Op >::value;
```
anyzelman, line 1171: This is also correct -- if given the same value to both operands returned that same value, then also that same value is negated and idempotency is retained. Please also note this in an (internal) comment above this class def.

tests/unit/operator_logical_not.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+#include <iostream>
+#include <sstream>
+
+#include <graphblas.hpp>
+
+using namespace grb;
+
+void grb_program( const size_t & n, grb::RC & rc ) {
+	(void)n;
```
anyzelman, line 26: missing space before `n`

tests/unit/operator_logical_not.cpp
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+#include <iostream>
+#include <sstream>
+
+#include <graphblas.hpp>
+
+using namespace grb;
+
+void grb_program( const size_t & n, grb::RC & rc ) {
```
anyzelman, line 25: superfluous space after `&` (twice)

tests/unit/operator_logical_not.cpp
```diff
@@ -0,0 +1,83 @@
...
+		bool value = true;
+		rc = rc ? rc : foldl( value, true, operators::logical_not< operators::logical_and< bool > >() );
+
+		// Check results
+		if( value != false || rc != SUCCESS ) {
+			rc = FAILED;
+			std::cerr << "Test logical_not< logical_and< bool > > FAILED\n";
+			return;
+		}
+	}
```
anyzelman, line 50: Three comments:
1. should we not also test negation of or and xor (do we even have xor actually?)
2. should we not test for false associative property of nand and nor?
3. should we not test the full truth table (two Booleans that can each take four values, i.e., four tests per operator)?

For bonus points, I don't think other operators have this exhaustively tested though, maybe all of foldl, foldr, and apply should be tested. Maybe with some nice software engineering this is actually doable with not too much repeated code

tests/unit/operator_logical_not.cpp
```diff
@@ -0,0 +1,83 @@
...
+	std::cout << "This is functional test " << argv[ 0 ] << "\n";
+	grb::Launcher< AUTOMATIC > launcher;
+	grb::RC out = SUCCESS;
+	size_t unused = 0;
+	if( launcher.exec( &grb_program, unused, out, true ) != SUCCESS ) {
+		std::cerr << "Launching test FAILED\n";
+		return 255;
+	}
+	if( out != SUCCESS ) {
+		std::cerr << "Test FAILED (" << grb::toString( out ) << ")" << std::endl;
```
anyzelman, line 78: The `test FAILED` output should go to `cout`. Before writing it, `cerr` should be flushed

include/graphblas/base/internalops.hpp
anyzelman: So this means that actually this type trait should be `false`. However, there is something annoying-- consider the type `logical_not< logical_not< OP< T > > >`. The associativity of the doubly-negated `OP` is of course the same as the associativity of `OP` itself.

One potential fix is to introduce an algebraic type trait `is_lnegated<OP>::value` (with lnegated short for logically negated) so that this line can be corrected to:
```
static constexpr const bool is_associative = is_lnegated<OP>::value ? OP::operator_type::is_associative : false;
```

include/graphblas/base/internalops.hpp
```diff
@@ -4179,6 +4297,9 @@ namespace grb {
 					/** The output domain of this operator. */
 					typedef typename OperatorBase< OP >::D3 D3;
 
+					/** The type of the operator OP. */
+					typedef OP operator_type;
```
anyzelman, line 4301: Nice addition!

Code style though -- should this not be OperatorType? (Classes are CamelCase)

anyzelman (changes requested):
With apologies -- chiefly the algebraic property propagation is a bit more tricky than currently accounted for

tests/unit/operator_logical_not.cpp
byjtew: > Three comments:
> 
>     1. should we not also test negation of or and xor (do we even have xor actually?)

We don't have xor, but why not.

>     2. should we not test for false associative property of nand and nor?

Absolutely.

>     3. should we not test the full truth table (two Booleans that can each take four values, i.e., four tests per operator)?

Will do.

> For bonus points, I don't think other operators have this exhaustively tested though, maybe all of foldl, foldr, and apply should be tested. Maybe with some nice software engineering this is actually doable with not too much repeated code

I'll see what I can think of but it would be nice to have such testing.

byjtew:
I continue the issue discussion here to avoid _outdated_ code review.

> One potential fix is to introduce an algebraic type trait `is_lnegated<OP>::value` (with lnegated short for logically negated) so that this line can be corrected to:
> 
> ```
> static constexpr const bool is_associative = is_lnegated<OP>::value ? OP::operator_type::is_associative : false;
> ```

This fix would does not work for this reason:
```
(1): logical_not< 
    (2): logical_not<
        (3): logical_and< bool >::is_associative = never reached
    >::is_associative = ( is_lnegated(3) <=> **false** ) ? (3)::is_associative : **false**
>::is_associative = ( is_lnegated(2) <=> **true** ) ? **(2)::is_associative** : false
```

Instead I modified the logical of __lnegated__ (renamed logically_negated because lnegated sounds like _left_-negated) like this:
```c++
template< typename OP >
struct is_logically_negated<
	operators::logical_not< OP >,
	typename std::enable_if< is_operator< OP >::value, void >::type
> {
	static constexpr const bool value = not is_logically_negated< OP >::value;
};
```

Then, I had to edit each potential access to the static field __is\_associative__ of the internal operator, from the _type\_traits_.

Now the annoying part:
This implied also deleting the public function __is\_associative()__ from the class _OperatorBase_ which is used here and there in the codebase, waiting for your feedback @anyzelman.

See last commit for these modifications.

include/graphblas/base/internalops.hpp
byjtew: See https://github.com/Algebraic-Programming/ALP/pull/224#issuecomment-1637667944

anyzelman:
Can we keep substract as an alias of subtract (or the other way around)

Can the MR description be updated to describe everything this MR does?

PR #225, opened 2023-07-13
647 Support for gcc-13

### LINKED ISSUE #212: Support for gcc-13
I noticed some errors while trying to manually compile ALP using gcc-13.

A step in the CI testing for this gcc version will be added, as well as the necessary modifications in the codebase to support it.

Porting documentation: https://gcc.gnu.org/gcc-13/porting_to.html

byjtew:
Closes #212

### DIFF
--- a/include/graphblas/base/config.hpp
+++ b/include/graphblas/base/config.hpp
@@ -33,7 +33,7 @@
 
 #include <assert.h>
 #include <unistd.h> //sysconf
-#include <cstdint>
+#include <cstdint> // for uintptr_t
 
 #include <graphblas/backends.hpp>
 
--- a/include/graphblas/base/config.hpp
+++ b/include/graphblas/base/config.hpp
@@ -33,6 +33,7 @@
 
 #include <assert.h>
 #include <unistd.h> //sysconf
+#include <cstdint>
 
 #include <graphblas/backends.hpp>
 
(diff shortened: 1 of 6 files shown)


include/graphblas/base/config.hpp
```diff
@@ -33,6 +33,7 @@
 
 #include <assert.h>
 #include <unistd.h> //sysconf
+#include <cstdint>
```
anyzelman, line 36: maybe add `// for size_t`, if indeed it is added because of that?

(also elsewhere)

anyzelman (changes requested):
Looks good to me, only a minor rfc.

Also a question: is GCC 13 added to the CI as one of the standard images tested against?

byjtew:
> Looks good to me, only a minor rfc.
> 
> Also a question: is GCC 13 added to the CI as one of the standard images tested against?

Not yet as we are testing using an Ubuntu image, and gcc-13 is not available yet using `apt-get`. 
This is in progress though, we would like to switch to spack modules. Will come in a future PR.

anyzelman:
Most recent commit addresses:
- reference coordinates.hpp misses cstdint include
- reference init.hpp misses cstdint include

Pending TODOs:
- [ ] check for more issues on either ARM or x86

anyzelman:
Alberto suggests to coincide with GCC 13 within the internal CI

AJ suggest to also have an ARM runner (GCC 13 threw different warnings on ARM)

PR #232, opened 2023-08-04
650 Potential bug in bps1d

### LINKED ISSUE #229: Potential bug in bps1d
The code below shows this initialisation of the **_local** matrix for the *bsp1d* backend.

The two `HERE`comments are there to show that `_id` is assigned to `id` after the call to `_local.initialize(&_id, ...)`. 
The problem is that `_id` is equal to `std::numeric_limits< uintptr_t >::max()` when passed to `_local.initialize(&_id, ...)`, which would assign every local matrices' id in a program to `std::numeric_limits< uintptr_t >::max()`.

Is this a typo where `_local.initialize(&_id, ...)` should be `_local.initialize(&id, ...)` or am I missing something ?

```c++
// ...
assert( _id == std::numeric_limits< uintptr_t >::max() );
// ...

// see if we can get an ID
const auto ptr = new char[ 1 ];
const auto id = data.mapper.insert(
	reinterpret_cast< uintptr_t >( ptr )
);
// now we changed our internal state, so we should be careful how to
// handle any subsequent exceptions
size_t global_cap = 0;
try {
	// complete local initialisation


	_local.initialize( &_id, local_m, local_n, local_nz );     // HERE (1)


	// sync global capacity
	global_cap = capacity( _local );
	if( collectives< BSP1D >::allreduce(
			global_cap,
			operators::add< size_t >()
		) != SUCCESS
	) {
		std::cerr << "Fatal error while synchronising global capacity\n";
		throw std::runtime_error( toString( PANIC ) );
	}
} catch( ... ) {
	// unwind the state change on the mapper
	data.mapper.remove( id );
	// then pass up the exception
	throw;
}

// all OK, so assign all local fields


_id = id;   // HERE (2)


_ptr = ptr;
_m = rows;
_n = cols;
_cap = global_cap;
```

byjtew:
Closes #229

### DIFF
--- a/include/graphblas/bsp1d/matrix.hpp
+++ b/include/graphblas/bsp1d/matrix.hpp
@@ -197,7 +197,7 @@ namespace grb {
 				// check default fields that should have been set by public constructor
 				assert( _m == 0 );
 				assert( _n == 0 );
-				assert( _id = std::numeric_limits< uintptr_t >::max() );
+				assert( _id == std::numeric_limits< uintptr_t >::max() );
 				assert( _ptr == nullptr );
 				assert( _cap == 0 );
 				// these default values correspond to an empty matrix and which the
@@ -265,7 +265,7 @@ namespace grb {
 					size_t global_cap = 0;
 					try {
 						// complete local initialisation
-						_local.initialize( &_id, local_m, local_n, local_nz );
+						_local.initialize( &id, local_m, local_n, local_nz );
 
 						// sync global capacity
 						global_cap = capacity( _local );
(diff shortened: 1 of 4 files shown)


anyzelman (approved):
Review OK

anyzelman:
Manual run of new unit test (with MPICH) OK. Waiting manual smoke tests

anyzelman:
Internal CI OK, GitHub CI OK, manual smoke test run (w/ LPF, MPICH) OK, will merge.

Concept release notes:

Benjamin found a bug where a BSP1D matrix would not initialise the ID of the sub-backends properly-- it would always assign an invalid ID. There is no known user-facing issue because of this bug-- when retrieving a container ID, the "front" backend is queried for it, not the sub-backend.

Nonetheless, without the fix this MR includes, it might have caused issues in some rare or future conditions. The MR also includes a new unit test that checks for all distributed backends (presently, `bsp1d` and `hybrid`), that the sub-backend IDs are correctly instantiated.

PR #233, opened 2023-08-04
228 iterators for populating matrices and vectors with some regular structure

### LINKED ISSUE #228: Iterators for populating matrices and vectors with some regular structure
This is especially useful for unit tests which currently rely on passing pointers to raw index & value arrays to `buildMatrix` and family.

Instead, specific nonzero patterns could be generated via (random access) iterators. HPCG is an example that already does it that way.

This issue proposes two phases:
1. provide a family of iterators that covers most standard patterns that are currently employed;
2. modify existing unit, smoke, and performance tests to make use of such iterators.

The end result would be faster, more robusts, and more scalable tests.

anyzelman:
This MR introduces several iterators that mimic iteration over containers with some regular structure, such as e.g. a constant (dense) vector or a range. These iterators are implemented in some minimal way, meaning a single base implementation is to cater to various such regular structures. Containers that return / make use of these iterators are also introduced.

They key feature is that iteration over n elements of some structure happens with Theta(1) memory overhead only.

The iterators are random access and STL-compatible. The containers are likewise STL-compatible.


anyzelman:
In the current version ([ab4945e](https://github.com/Algebraic-Programming/ALP/pull/233/commits/ab4945ebadc292fdc1e6659c5b5e531b0ed77f1e)) all unit tests pass (with LPF)

tests/unit/iteratorFilter.cpp
```diff
@@ -0,0 +1,385 @@
...
+		auto beginCopy2 = itPair.first;
+		auto endCopy = itPair.second;
+		if( n < 18 ) {
+			expected = n;
+		} else {
+			expected = n - 1;
+		}
+		testOneOut( n, 17, expected, beginCopy, itPair.second, local_rc );
+		if( local_rc != grb::SUCCESS ) {
+			std::cerr << "Test 2 FAILED\n";
```
byjtew:
```suggestion
			std::cerr << "Test 2 FAILED: rc is " + grb::toString( rc ) + "\n";
```

Everywhere in this file ?

tests/unit/iteratorFilter.cpp
```diff
@@ -0,0 +1,385 @@
...
+	{
+		auto itPair = grb::utils::iterators::createIteratorFilter(
+			v.begin(), v.end(),
+			[] (const size_t val) { return val == 7; }
+		);
+		if( n < 8 ) {
+			expected = n;
+		} else {
+			expected = n - 1;
+		}
```
byjtew, line 110:
```suggestion
		expected = ( n < 8 ) ? n : n-1;
```

Ternary usage in the whole file ?

tests/unit/iteratorFilter.cpp
```diff
@@ -0,0 +1,385 @@
...
+			rc = grb::FAILED;
+			return;
+		}
+	}
+	if( count != expected ) {
+		std::cerr << "Error: count should be " << n << ", "
+			<< "but the filtered iterator returned " << count << " elements instead\n";
+		rc = grb::FAILED;
+	}
+}
```
byjtew, line 54:
```suggestion
template< typename FwdIt >
void testOneOut(
	const size_t &n, const size_t filtered,
	FwdIt &it, const FwdIt &end,
	grb::RC &rc
) {
	const size_t expected_size = n - std::count( it, end, filtered );
	
	// Technically, n could also be replaced by:
	// const size_t n = std::distance( it, end );
	
	rc = grb::SUCCESS;
	size_t count = 0;
	for( ; it != end; ++it, ++count ) {
		if( *it != *(it.operator->()) ) {
			std::cerr << "Error: dereference operator returns " << *it << ", "
				<< "but dereferencing the pointer operator returns " << (*(it.operator->()))
				<< "\n";
			rc = grb::FAILED;
			return;
		}
		if( *it == filtered ) {
			std::cerr << "Error: found " << filtered << ", "
				<< "which should have been filtered out\n";
			rc = grb::FAILED;
			return;
		}
	}
	if( count != expected_size ) {
		std::cerr << "Error: count should be " << expected_size << ", "
			<< "but the filtered iterator returned " << count << " elements instead\n";
		rc = grb::FAILED;
	}
}
```

byjtew (commented):
The naming of the files should also be changed to be coherent

tests/unit/iteratorFilter.cpp
anyzelman: I did this everywhere (will commit soon); thanks!

tests/unit/iteratorFilter.cpp
anyzelman: I see it hadn't been committed yet, and I'm too lazy to do this one (the difference in readability is minimal). Hope ok to skip?

tests/unit/iteratorFilter.cpp
anyzelman: I'm not sure what the difference is between the blocks, OK to ignore?

tests/unit/iteratorFilter.cpp
byjtew: Absolutely!

anyzelman:
Hi @byjtew -- I had processed the suggestions (see above) and rebased. However, the new factories unit test fails in the distributed memory case, the easiest of which being bsp1d on two nodes. Could you check what goes wrong there?

tests/unit/iteratorFilter.cpp
byjtew: The **expected** value is computed in the function instead of begin passed as an argument

byjtew:
> Hi @byjtew -- I had processed the suggestions (see above) and rebased. However, the new factories unit test fails in the distributed memory case, the easiest of which being bsp1d on two nodes. Could you check what goes wrong there?

Sure thing

tests/unit/iteratorFilter.cpp
anyzelman: Ah I see, sneaky hahah. I prefer the old variant because it explicitly forces the unit tester to think about what they expect=)

byjtew: Sounds good

byjtew:
Fixed ! @anyzelman

anyzelman:
> Fixed ! @anyzelman

Confirmed to pass with LPF now:) GitHub CI OK.

Waiting internal CI and final code review (me) before merge.

anyzelman:
@byjtew Hi Benjamin -- I have the following on a first deeper review of this MR:
1. could you please check my edits of matrix_factory.hpp to see if it's ok with you?
2. there are a couple of FIXMEs in that file, in particular in relation to random matrix generation (which I think can break). Could you have a look?
3. I think the factories unit test does not test diagonal offsets -- does it indeed not do that?
4. Likewise, I think the factories unit test is presently not testing the putting arbitrary values from some iterator on the diagonal -- does it indeed not do that? (If not, and/or also if the above no. 3 does not, then it should be added to the unit test, or otherwise maybe we "promote" it to a smoke test instead-- although otoh the unit test is already pretty encompassing, this seems like a worthwhile and minor addition.)
5. ChainedIterators -- is a better name not InterleavedIterators? Also the corresponding container could then be called InterleavedVector.
6. The ChainedIterator currently implements += and -= as a loop-over ++ and --. The implicit contract for += is that it can skip forward multiple places in O(1) time, otherwise it's like std::forward instead
7. Relatedly, the iterator category of the chained iterator / interleaved iterator should be a forward iterator or a bi-directional iterator -- even if it derives exclusively from random access iterators. The reason is item no. 6 directly above this point. (See e.g. also the filteredIterator, which also discards any random access property of the sub-iterator for similar reasons.)

anyzelman:
Hang on -- I fell into a trap yesterday: I was thinking these factories only work with sequential semantics given that we want to implement them on the algorithm level, rather than the core primitive level, and also given that we presently do not have the necessary introspection on the distribution a backend may employ behind the scenes. While the latter is indeed a difficult problem, the trap is that for parallel I/O to work, we actually do *not* need to match exactly the backend's distribution-- it would be ideal, but not required to achieve scalability. I will rework the MR to fully support parallel I/O.

anyzelman:
Making a list of pending items:

- [x] enable parallel I/O based on structured iterators
- [x] make factories templated in backend (hidden) as well as matrix nonzero type (simplifies its use)
- [ ] revise chainedIterators w.r.t. iterator tag type (I'm now thinking actually it could have a proper parallel implementation if the subiterators are random access-- just use modulo and division to figure out how far each iterator should move)
- [x] instead of using chainedIterations, we can use position-based iterators for both row and column indices (would be slightly more efficient than chainedIterators, and clearer how to effect parallel I/O in that case)
- [x] improve factory unit tests (see above)
- [x] fix random factory (see above)

anyzelman:
Now all matrix factory methods except random matrix creation are shared-memory parallel. The way it is currently fixed no longer makes it required to "upgrade" the ChainedIterator to a random access iterator.

@byjtew , would you
1. nonetheless want to make that change and keep it in this MR,
2. would you instead like to put ChainedIterators in a separate MR and handle it further once another use case for it comes in, or
3. keep implementation as-is but follow the above cursory edits, which are, to summarise, override sub-iter tag and force that the ChainedIterator is a forward iterator, remove unnecessary-for-forward-iterators member functions (such as operator+= etc.), and rename ChainedIterator->InterleavedIterator?

The next step on my list is to enable parallel I/O for the matrix factory classes -- the above commits already modify the regular iterators with the required functionality, and unit tests those. I'm first gonna run the unit test suite to make sure no bugs are present in the current state before continuing.

anyzelman:
> All smoke and unit tests with LPF OK at this point

At this point (after refactoring the factory class;) all tests still OK

anyzelman:
So in sequential I/O mode everything works. When switching to parallel I/O mode, however, some things break for bsp1d and hybrid. This is yet to be debugged, some time over the holidays.

Some thoughts on the diag function:
- there should be diag from vector to matrix and back. I'm not sure if that already had an issue somewhere.
- I'm not sure that there should be a diag in the factory class -- the more I think about it, the semantics are messy because it takes any iterator as input. Those iterators may not match the distribution of row/column iterators internal to the factory function. Therefore, actually, sequential I/O semantics are forced -- which means it's not scalable.
- One possible fix to the above item is to define diag more akin to the first bullet above. That is, we automatically derive the size of the diagonal matrix from the input iterator, and then the row/column iterators can simply follow the given iterator. However, we must still impose a sort of "variable block distribution" of the map of iterator contents to the diagonal, while the order of blocks would be somewhat arbitrary (e.g., the ordered according to PID) when in parallel I/O mode.
- As a final note, in most tests, bullet point 1 above should actually suffice: one could use sequential mode buildVector to turn the diagonal data into a `grb::Vector`, and then use diag to promote it to a `grb::Matrix`. This suffices for all use cases in the test suite.

byjtew:
> So in sequential I/O mode everything works. When switching to parallel I/O mode, however, some things break for bsp1d and hybrid. This is yet to be debugged, some time over the holidays.

🤷‍♂️

> Some thoughts on the diag function:
> 
>     * there should be diag from vector to matrix and back. I'm not sure if that already had an issue somewhere.
>     * I'm not sure that there should be a diag in the factory class -- the more I think about it, the semantics are messy because it takes any iterator as input. Those iterators may not match the distribution of row/column iterators internal to the factory function. Therefore, actually, sequential I/O semantics are forced -- which means it's not scalable.

I'd rather use the `matrices<D>::identity(rows, cols, values_iterator)` to build an identity-shaped matrix with the given values, at least for the unit-tests.

>     * As a final note, in most tests, bullet point 1 above should actually suffice: one could use sequential mode buildVector to turn the diagonal data into a `grb::Vector`, and then use diag to promote it to a `grb::Matrix`. This suffices for all use cases in the test suite.

That would also work, less user-friendly though

anyzelman:
> 🤷‍♂️

I'm back on it. Meanwhile Alberto and I couldn't remember why this issue had priority, I thought it was something related to ALP/Spark but seems not-- do you maybe remember?
 
> I'd rather use the `matrices<D>::identity(rows, cols, values_iterator)` to build an identity-shaped matrix with the given values, at least for the unit-tests.

To clarify, the issue is that the distribution of rows, cols, and values should match when in parallel I/O. I don't think we should expose such a erquirement to the user at all. The promotion idea avoids the issue by having only a single distributed iterator over the values, in which case any distribution is fine for parallel I/O to work.

anyzelman:
At this point I'm fairly certain the bug is squashed. Remaining items, for brevity re-listing them, are:

- [ ] revise chainedIterators w.r.t. iterator tag type (I'm now thinking actually it could have a proper parallel implementation if the subiterators are random access-- just use modulo and division to figure out how far each iterator should move)
- [x] remove diag from factories and make sure an issue exists that promotes a vector to a diagonal matrix
- [x] improve factory unit tests (see above)
- [x] remove random functionalities or tag them as experimental

@byjtew I wanna close this one up asap as it seems maybe we shouldn't have prioritised this one over some of the other issues. Is it therefore OK if we skip the first of the above TODOs and create another issue & MR that introduces Chained/Interleaved iterators, and is it ok for you to remove the diag from the factories and instead rely on vector-to-diagonal-matrix-promition as a separate algorithm or core primitive?

I'm now going to take a lock on item no. 3 and finish that in the next hours or so, as that one seems necessary to merge and move on from this MR in any case=)

byjtew:
> @byjtew I wanna close this one up asap as it seems maybe we shouldn't have prioritised this one over some of the other issues. Is it therefore OK if we skip the first of the above TODOs and create another issue & MR that introduces Chained/Interleaved iterators, and is it ok for you to remove the diag from the factories and instead rely on vector-to-diagonal-matrix-promition as a separate algorithm or core primitive?
> 
> I'm now going to take a lock on item no. 3 and finish that in the next hours or so, as that one seems necessary to merge and move on from this MR in any case=)

Absolutely, I never understood why this branch was considered urgent. 
I rebased on top of my changes from GitLab and just pushed, waiting for the CI.

anyzelman:
Hi @byjtew --

OK, then let's try and finish up. Something seems horribly wrong with the above rebase though -- it seems to revert to a much older version of the factories unit test. I could resolve the conflicts I see now but I would worry some other old stuff is re-instated as well. Is it OK to undo it, could you do that (I'm signing off in this time zone now)? If you can't then I can force push the previous one.

In general, if there exists a GitHub branch, we should always consider the GitHub one to be the master, rather than the GitLab. I prefer to resolve conflicts on the same branch via merge commits rather than rebasing-- that makes it more clear what is happening (whereas in this case I have no clue what is lost and how).

I did some more checking -- in fact a bunch of changes I had made seem lost, in particular what I saw lost is the change from ChainedIterators towards adapted iterators (so that we have full benefits of the random access iteration).

anyzelman:
Also, separately to note progress: the version before the latest rebase from Benjamin is confirmed to pass all unit and smoke tests with LPF enabled.

The version I can't push atm due to conflicts tests diagonal offsets to identity and passes for the reference backend. TODO is to also test the eye function in a similar way, but I wanted to first check if this version passes for all backends.

byjtew:
I merged not rebased, you just have to pop the last commit.
I had to modify the unit-tests to adapt for the correct calls to the factories. 
As for the Iterators it shouldn't have changed.

In case it's not correct, could you push your current branch state in a new one with a suffix like "-save"

anyzelman:
> I merged not rebased, you just have to pop the last commit. I had to modify the unit-tests to adapt for the correct calls to the factories. As for the Iterators it shouldn't have changed.
> 
> In case it's not correct, could you push your current branch state in a new one with a suffix like "-save"

ah ok then I can just force-push right, it will safely ignore your merge? (I don't think you consciously e.g. removed the dependence on the adapter iterator right?)

byjtew:
> > I merged not rebased, you just have to pop the last commit. I had to modify the unit-tests to adapt for the correct calls to the factories. As for the Iterators it shouldn't have changed.
> > 
> > In case it's not correct, could you push your current branch state in a new one with a suffix like "-save"
> 
> ah ok then I can just force-push right, it will safely ignore your merge? (I don't think you consciously e.g. removed the dependence on the adapter iterator right?)

Yes you can, I haven't touched that part of the codebase (at least I hope)

anyzelman:
> > > I merged not rebased, you just have to pop the last commit. I had to modify the unit-tests to adapt for the correct calls to the factories. As for the Iterators it shouldn't have changed.
> > > In case it's not correct, could you push your current branch state in a new one with a suffix like "-save"
> > 
> > 
> > ah ok then I can just force-push right, it will safely ignore your merge? (I don't think you consciously e.g. removed the dependence on the adapter iterator right?)
> 
> Yes you can, I haven't touched that part of the codebase (at least I hope)

Done..and somehow it did touch that part -- there was a commit on like the 12th of January if I remember right (the force push removed the exact commit) that undid at least the move from ChainedIterators to adapters as well as undid some of the code style fixes to the factories unit test (and probably more).

I think if you try to merge or rebase again (on this GitHub branch) it should raise conflicts, if manually resolving them it should be fine. If the conflicts are unclear how to resolve, you can ask me to merge (probably easiest to create a new branch+MR that merges into this one:)

I'll run the unit test overnight on the current state to make sure all backends can handle identity with diagonal offsets and then continue expanding the unit test tomorrow HK time:)

anyzelman:
@byjtew As I went through the factories implementation earlier today I noticed one more forgotten issue: the random generation of matrices has two remaining issues:

1. it seems bugged (the random sampling could generate the same coordinates more than once, which buildMatrixUnique disallows);
2. random matrix generation cannot proceed in parallel, neither shared-memory parallel (it would need parallel RNGs for that), nor in a distributed-memory parallel fashion (neither in sequential or parallel I/O modes). Or more precisely, the code would execute serially in all those cases.

We could either remove the functionality for now, or tag it as experimental -- what do you think / prefer?

byjtew:
> > > > I merged not rebased, you just have to pop the last commit. I had to modify the unit-tests to adapt for the correct calls to the factories. As for the Iterators it shouldn't have changed.
> > > > In case it's not correct, could you push your current branch state in a new one with a suffix like "-save"
> > > 
> > > 
> > > ah ok then I can just force-push right, it will safely ignore your merge? (I don't think you consciously e.g. removed the dependence on the adapter iterator right?)
> > 
> > 
> > Yes you can, I haven't touched that part of the codebase (at least I hope)
> 
> Done..and somehow it did touch that part -- there was a commit on like the 12th of January if I remember right (the force push removed the exact commit) that undid at least the move from ChainedIterators to adapters as well as undid some of the code style fixes to the factories unit test (and probably more).
> 
> I think if you try to merge or rebase again (on this GitHub branch) it should raise conflicts, if manually resolving them it should be fine. If the conflicts are unclear how to resolve, you can ask me to merge (probably easiest to create a new branch+MR that merges into this one:)
> 
> I'll run the unit test overnight on the current state to make sure all backends can handle identity with diagonal offsets and then continue expanding the unit test tomorrow HK time:)

As it's not used for the moment, I would suggest removing it.

anyzelman:
The unit tests are expanded with off-diagonal unit tests. At current, the following unit tests fail (tick when fixed):
- [x] factories (BSP1D, P=16, NDEBUG)
- [x] eWiseApply_matrix (BSP1D, P=16, NDEBUG)
- [x] factories (hybrid, P=7, T=2, NDEBUG)
- [x] eWiseApply_matrix (hybrid, P=7, T=2, NDEBUG)
- [x] factories (hybrid, P=7, T=12, NDEBUG)
- [x] eWiseApply_matrix (hybrid, P=7, T=12, NDEBUG)
- [x] factories (BSP1D, P=2, DEBUG)
- [x] ... (other tests still running)

anyzelman:
Fixed three bugs (one of which a meta-bug). Manually tested factories, bsp1d, P=2, debug now passes correctly.

Rerunning unit test suite now to see whether that also fixes the other failed tests.

byjtew:
Fixed one potential bug and improved documentation, running CI on GitLab

anyzelman:
@byjtew I see your last merge crosses the current test suite run -- this one again looks confusing to me, as it seems to remove the offset from the identity function (which i think I added just yesterday), while the unit test keeps testing identity with offsets?

What also seems to go wrong is that eWiseApply_matrix should use eye, not identity, as the code that is changed works using non-square matrices?

I think to close this down (the tests are finally likely to pass this time round), let's discuss quickly the semantics. Prior to merge we had:
1. eye as possibly non-square and possibly with diagonal offsets, while keeping
2. identity as square with possibly diagonal offsets. 

The rationale is I'm not to keen on non-square identities-- given the A in ALP it seems nicer to stick with more rigid interpretations of the math involved. Eye, being a separate function from the identity, I thought can be more generic.

I was therefore indeed in doubt about allowing diagonal offsets to identity though, as that also of course doesn't result in a proper identity. I ended up adding it, but am OK with removing that again as you already started doing, but then the tests should be fixed up (as in identity should not be tested with offsets any more), while the "missing" offset tests that identity-with-offset currently tests for (I think 1 and -2) should then be added as additional tests for eye-with-offsets.

The eWiseApply_matrix should def. always use eye, I'm not sure what happens currently but I don't think it will pass. (Note BTW that the regular CI tests may not be sufficient for this MR-- it needs to run them with LPF and for the full set of #processes.)

The docs etc. ofc very welcome.

Indeed btw updating on the test run -- all ndebug mode tests flagged as failing earlier today now pass. The test suite is still running for debug mode, but the only one I saw failed is also confirmed fixed already.

byjtew:
> The rationale is I'm not to keen on non-square identities-- given the A in ALP it seems nicer to stick with more rigid interpretations of the math involved. Eye, being a separate function from the identity, I thought can be more generic.

But then you can easily write incorrect code like `identity( 10, 10 )` thinking that you declared an identity matrix of 10x10 with a default value of 1 while you actually initialised a 10x10 with a value of 10. A trade-off would be to remove any default value for the parameters.

> I was therefore indeed in doubt about allowing diagonal offsets to identity though, as that also of course doesn't result in a proper identity. I ended up adding it, but am OK with removing that again as you already started doing, but then the tests should be fixed up (as in identity should not be tested with offsets any more), while the "missing" offset tests that identity-with-offset currently tests for (I think 1 and -2) should then be added as additional tests for eye-with-offsets.

Done

> The eWiseApply_matrix should def. always use eye, I'm not sure what happens currently but I don't think it will pass. (Note BTW that the regular CI tests may not be sufficient for this MR-- it needs to run them with LPF and for the full set of #processes.)

There is no offset involved in this test, hence `identity` should be preferred, no ?

> The docs etc. ofc very welcome.
> 
> Indeed btw updating on the test run -- all ndebug mode tests flagged as failing earlier today now pass. The test suite is still running for debug mode, but the only one I saw failed is also confirmed fixed already.

anyzelman:
> But then you can easily write incorrect code like `identity( 10, 10 )` thinking that you declared an identity matrix of 10x10 with a default value of 1 while you actually initialised a 10x10 with a value of 10. A trade-off would be to remove any default value for the parameters.

The removal of the default may actually compound the issue you're sketching. I see the following options:
* leave it as is (two arguments, second is default 1)
* use explicit (two arguments, second is default 1 but uses the explicit keyword to avoid implicit casting)
* instead of an (optional) identity value, give an optional semiring

I'm actually leaning towards the latter, what about you?

> There is no offset involved in this test, hence identity should be preferred, no ?

If identity can only generate square matrices and the test uses rectangular ones, then it should use eye

anyzelman:
Issue in unit test fixed, now re-running both unit and smoke tests with LPF

anyzelman:
More issues found and fixed in unit tests + LPF. I'm first checking if they are truly fixed; if yes, I will fix another metabug which resulted in a deadlock, namely, the following:

When using bsp1d or hybrid with p>1 user processes, it could happen that several q>0 user processes report an error while p-q processes do not. (This happened with the above meta-bug.)

In this case, because the error programming pattern used in this unit test, some processes may exit before others do. This (in hindsight and perhaps) may also be why the other unit tests use a fall-through pattern (errors are propagated to the end of the unit test, and synchronised with other processes at the end or when on error a test would not be able to proceed). This ensures processes exit at the same point, without leaving some processes stuck in an attempt to communicate with exited process.

anyzelman:
Testing update: at commit [6f696da](https://github.com/Algebraic-Programming/ALP/pull/233/commits/6f696dafabaa0ad966cde0e3b6686c2dc9ddda35) all is good again for tests with LPF.

@byjtew -- thanks for that update on the error handling. Only one thing:
1. this variant on error will still skip all remaining tests -- was that on purpose?
 
The `rc = rc ? rc : ` pattern I typically use in algorithms. I only use it in tests if failure of a preceding step leads to a non-meaningful check later on -- e.g., in a test for `vxm`, if the initial set fails, then I'm not interested in the rest of the test:

```
rc = grb::set( y, 0 );
rc = rc ? rc : grb::vxm( y, x, A, someSemiring );
```
In this case, however, even if nnz(M)==0 fails, I may still be interested in knowing whether nrows(M)==0 also failed, e.g. So something like the following may be preferred:

```
if( nnz(M) != 0 ) {
     std::cerr << somemessage;
     rc = FAILED;
}
if( nrows(M) != 0 ) {
     std::cerr << somemessage;
     rc = FAILED;
}
....
```

Then also, since this way of testing may cause a value for rc that is not synchronous between processes. Thus at the end of the test, code like the following should appear:

```
if( grb::collectives<>::allreduce( rc, grb::operators::any_or< grb::RC >() ) != grb::SUCCESS ) {
    std::cerr << somemessage;
    rc = PANIC;
}
```

This ensures that the `rc` code is fully in sync on program exit.

anyzelman:
Just updating: only the collectives are needed to update-- the pattern in the unit test does not include executing the primitive. The factories test uses `rc ? rc : FAILED`, meaning the rc is only set once but the remainder test still executes everything even if detecting earlier failures.

anyzelman:
OK, I'm happy:) @byjtew , do you wanna check the latest couple of changes and if all is ok for you, approve this one by review?

I triggered internal CI with LPF, and started manual unit + smoke tests with LPF. GitHub CI is already happy.

tests/unit/factories.cpp
```diff
@@ -20,17 +20,16 @@
...
 
 
 using namespace grb;
 using namespace grb::algorithms;
 
-#define ERROR( rc, msg ) {										\
-	std::cerr << "  [Line " << __LINE__ << "] Test FAILED: "	\
-		<< (msg) << std::endl;									\
-	(rc) = FAILED;												\
+#define ERROR( rc, msg ) {                                       \
```
byjtew, line 29: :sob:

tests/unit/factories.cpp
```diff
@@ -849,10 +848,12 @@ void grb_program( const size_t &n, RC &rc ) {
...
-	// Check return code for distributed backends
-	ASSERT_RC_SUCCESS(
-		collectives<>::allreduce( rc, grb::operators::any_or< RC >() )
-	);
+	// synchronise return code for distributed backends
+	if( collectives<>::allreduce( rc, grb::operators::any_or< RC >() )
+		!= SUCCESS )
+	{
+		rc = PANIC;
+	}
```
byjtew: We will lose the original error by assigning PANIC, maybe a print to stderr before ?

include/graphblas/algorithms/matrix_factory.hpp
```diff
@@ -227,6 +232,14 @@ namespace grb::algorithms {
 			 *
 			 * The \a V_iter and \a V_end iterator pair must contain exactly the number
 			 * of elements that should appear on the requested diagonal.
+			 *
+			 * \warning In parallel I/O mode, the given iterator pair must match in
+			 *          terms of both contents and order of returned elements, to those
+			 *          of the row- and column-indices returned by the \em internally-
+			 *          defined iterators, which (currently) are given by:
+			 *            - #grb::utils::iterators::Range .
```
byjtew: extra dot ?

byjtew:
```suggestion
			 *            - #grb::utils::iterators::Range
```

anyzelman:
Concept release notes:

This MR introduces regular iterators to the ALP/GraphBLAS utilities. Regular iterators iterate over virtual containers that have some regular structure, such as ranges (1, 2,..., n) or constant vectors (c, c, ..., c). Immutable containers corresponding to such virtual containers are added as well, and retrieving iterators over those containers returns the new regular iterators. These iterators and container classes are fully STL compliant and adhere to all requirements of random access iterators.

By overriding some default arguments, the containers and iterators furthermore allow for parallel I/O, by iterating over only a chunk of the (virtual) container-- thereby, a constant vector or a range can be stored in Theta(P) space and ingested into ALP containers, roughly, in O((n+m)/P) time, where P are the number of user processes, n the maximum ALP container dimension size, and m the number of elements (nonzeroes).

This functionality makes possible an easy way to generate some standard matrices, in a fully scalable way, both on shared-memory parallel machines (through random access iteration and shared-memory backends, as already available) as well as on distributed-memory parallel machines (through parallel I/O, as also already was available). These standard matrices include the identity matrix, diagonal matrices, or dense matrices. This MR introduces these in `include/graphblas/algorithm/matrix_factories.hpp`, following a factory pattern; e.g., `grb::algorithms::matrices<double>::identity( 400 );` will return a `grb::Matrix<double>` instance initialised as a 400 x 400 identity matrix.

In addition to the regular iterators, the MR also introduces an adapter iterator, which takes any valid iterator pair and produces a matching iterator pair which transforms the values iterated over by the former pair according to a given `std::function`. This iterator allows adapting on the fly values returned by a source iterator, without having to copy (and then modify) the underlying container. The adapter iterator is presently only defined for random access iterators, is also used to generate some standard structured matrices, and can wrap around "parallel I/O iterators".

Finally, the MR includes an iterator filter which can take any input iterator pair and return a matching iterator pair that filters the elements the former iterates over according to an user-defined function. This iterator filter necessarily demotes the category of any iterator it wraps around to a forward iterator, which limits its applicability (e.g., in ALP, forward iterators do not allow for shared-memory parallel ingestion into ALP containers).

Unit tests have been added to test all new functionalities: the matrix factory, the iterator filter, the regular iterators (both sequential & parallel), and the iterator adapter. Additionally, some unit tests which relied on sequential I/O and hard-coded or dynamic raw C++ arrays before ingestion into ALP containers have been modified to make use of the factory classes instead.

This MR also includes some unrelated bugfixes:
- the `getOne` and `getZero` functions of a semiring are defined as stateful, but also constexpr, which was misleading at best. This MR removes `constexpr` from those functions;
- the order of unit tests has been adapted so that tests for function X that depend on function(s) Y has the test for Y before X, for those tests that depend on changes this MR introduces. 

Other changes include:
- unit tests that compare matrix contents now rely on test/utils, thus reducing code duplication;
- internal CI now saves test log files;
- documentation fixes (formatting, typos);
- enable user-level documentation of grb::util features related to this MR;
- improvement of the `print_vec_mat.hpp` test/util;
- the usual code style fixes mostly in files related to this MR.

tests/unit/factories.cpp
anyzelman: what happened? The alignment for me was off, I guess because it used tabs instead of spaces. But maybe now it doesn't look great in your editor is it?

tests/unit/factories.cpp
anyzelman: ok, making (and testing) the change

include/graphblas/algorithms/matrix_factory.hpp
anyzelman: The dot should be there (the sentence continues from the preceding line), but the space is not necessary (tested with doxygen just now). However, iterators->containers. I'll make both changes.

tests/unit/factories.cpp
byjtew: On GitHub it looks messed up but not locally so it's fine

anyzelman:
Hi @byjtew , your suggestions should be processed.

I also found that actually doxygen was not yet enabled for these utilities, so I fixed that up-- and in checking the user-level docs I found the iteratorFilter's factory method had a diverging signature from those of the adapter, chained/interleaved, sequence, and repeated iterators-- so I fixed that up as well.

I'm re-running the tests everywhere again, meanwhile, the same review OK question to you once more on the recent changes? Many thanks!

byjtew:
> I'm re-running the tests everywhere again, meanwhile, the same review OK question to you once more on the recent changes? Many thanks!

Looks good to me!

anyzelman:
> > I'm re-running the tests everywhere again, meanwhile, the same review OK question to you once more on the recent changes? Many thanks!
> 
> Looks good to me!

can you press approve?=)

byjtew (approved):

anyzelman:
Thanks! Meanwhile both CIs and manual tests with LPF OK, and the reviews are OK, so merging now=)

Issue #235, opened 2023-08-08
Failing assertion in BSP1D exec routine


alberto-scolari:
This assertion fails

https://github.com/Algebraic-Programming/ALP/blob/dbe9a76ad9ba56e7bb6a19a98a32761083a15281/include/graphblas/bsp1d/exec.hpp#L88

and it is probably wrong: indeed, the value is for all processes `sizeof(T)`.

Furthermore, at line

https://github.com/Algebraic-Programming/ALP/blob/dbe9a76ad9ba56e7bb6a19a98a32761083a15281/include/graphblas/bsp1d/exec.hpp#L209

there is a memory leak, since the variable `data_in` is allocated but never freed.

finally, the routines `_grb_exec_spmd` and `_grb_exec_varin_spmd` are almost identical, and can probably be refactored to avoid duplicated bugs.

anyzelman:
Related issues fixed in #239 , this issue should no longer apply

PR #239, opened 2023-08-28
235 failing assertion in bsp1d exec routine

alberto-scolari:
This PR fixes several issues found in the BSP1D Launcher and also refactors a lot of duplicated code between Launcher and Benchmarker; hence, it removes more code than it adds, despite implementing the same number of functionalities.

A **unit** test is also added for the Launcher and Benchmarker, checking that they work fine in all scenarios (AUTOMATIC, FROM_MPI, MANUAL).

@anyzelman some smoke tests manually implement similar functionalities, not relying on Launcher, so there is probably overlap with the test this PR adds. I noticed though that the code of these tests is a bit "tangled", so it's not entirely clear to me what and how exactly they test the various launching mechanisms. I also suspect these tests date back to when we had no Launcher. What should we do about them?
1. remove, if they test the same things (though implemented there by hand)
2. port them to using the Launcher (duplicates?)
3. leave as is


anyzelman (changes requested):
@alberto-scolari -- while reviewing the unit test, I ran into the following two issues for `bsp1d` and `hybrid`:

1. there is no check that in `AUTOMATIC` no-broadcast mode, that the input struct received by processes `s>0` is empty;
2. I assume that in `MANUAL` or `FROM_MPI` *with broadcasting*, that any given inputs on processes `s>0` will be overwritten by the input on process 0; while without broadcasting, the idea is that their initial inputs remain. The test does not test the differing behaviour, however.

Starting with the latter-- is there another poem you could use to initialise the in-struct with for processes `s>0`? The final check should then differentiate between broadcasting or non-broadcasting -- in the former, the original poem should be returned on all processes, while in the latter case the new poem should remain for processes `s>0` and the old poem only for process 0.

For the former, the implementation in `bsp1d/exec.hpp` seems to always require a broadcast when the `AUTOMATIC` mode is requested-- this is not intended. If the broadcast flag is false, then the program should continue with "empty" input for the user processes with ID `s>0` (recall the assert mentioned in the originating issue).
For typed ALP programs, I do recall the issue with default-constructibility, so what I would suggest is the following:
- for untyped ALP programs, automatic mode with broadcast `false` results into a null `in`  pointer and zero `in_size` at processes `s>0`;
- for typed ALP programs with default-constructible input POD types, automatic mode with broadcast `false` results in default-constructed inputs at processes `s>0`;
- for typed ALP programs with non-default-constructible input types, automatic mode with broadcast `false` does nothing except return `ILLEGAL`.

In all above cases ALP does not initiate any broadcast. For variants where broadcast is `true` the behaviour should remain unchanged compared to current behaviour.

Does this make sense?

anyzelman:
For the question in the MR btw -- number 3, but together with opening up a new issue for it (probably with a cleanup tag:)

alberto-scolari:
> For the question in the MR btw -- number 3, but together with opening up a new issue for it (probably with a cleanup tag:)

I'd propose directly 1., but if we want to merge ASAP then 3. + cleanup issue is also fine.

alberto-scolari:
> @alberto-scolari -- while reviewing the unit test, I ran into the following two issues for `bsp1d` and `hybrid`:
> 
> 1. there is no check that in `AUTOMATIC` no-broadcast mode, that the input struct received by processes `s>0` is empty;
> 2. I assume that in `MANUAL` or `FROM_MPI` _with broadcasting_, that any given inputs on processes `s>0` will be overwritten by the input on process 0; while without broadcasting, the idea is that their initial inputs remain. The test does not test the differing behaviour, however.
> 
> Starting with the latter-- is there another poem you could use to initialise the in-struct with for processes `s>0`? The final check should then differentiate between broadcasting or non-broadcasting -- in the former, the original poem should be returned on all processes, while in the latter case the new poem should remain for processes `s>0` and the old poem only for process 0.
> 
> For the former, the implementation in `bsp1d/exec.hpp` seems to always require a broadcast when the `AUTOMATIC` mode is requested-- this is not intended. If the broadcast flag is false, then the program should continue with "empty" input for the user processes with ID `s>0` (recall the assert mentioned in the originating issue). For typed ALP programs, I do recall the issue with default-constructibility, so what I would suggest is the following:
> 
> * for untyped ALP programs, automatic mode with broadcast `false` results into a null `in`  pointer and zero `in_size` at processes `s>0`;
> * for typed ALP programs with default-constructible input POD types, automatic mode with broadcast `false` results in default-constructed inputs at processes `s>0`;
> * for typed ALP programs with non-default-constructible input types, automatic mode with broadcast `false` does nothing except return `ILLEGAL`.
> 
> In all above cases ALP does not initiate any broadcast. For variants where broadcast is `true` the behaviour should remain unchanged compared to current behaviour.
> 
> Does this make sense?

About POD types. I didn't implement this check, 'cause it likely breaks existing code. POD is a quite strict requirement, which e.g., requires that the default constructor is trivial (i.e., not user specified). A better constrains should be `[is_standard_layout](https://en.cppreference.com/w/cpp/types/is_standard_layout)`, but this is also violated in some pre-existing tests (maybe 1 or 2 -- we can fix them).

anyzelman:
> About POD types. I didn't implement this check, 'cause it likely breaks existing code. POD is a quite strict requirement, which e.g., requires that the default constructor is trivial (i.e., not user specified). A better constrains should be [is_standard_layout](https://en.cppreference.com/w/cpp/types/is_standard_layout), but this is also violated in some pre-existing tests (maybe 1 or 2 -- we can fix them).

Where does it break it? The '20 paper and the current spec clearly defines only POD types are allowed, so at the very least such occurrences would not be compliant and should be fixed.

If I take this as a suggestion to relax the POD requirement to allowing standard layouts, then my main worry is the following: standard layouts in their intent do not preclude the use of pointers-- in fact, intuitively, I would be fine with standard layouts containing pointer data, while I would question that for POD types (even though the C++ spec's fine with it, at least in 11 and 14 I believe).

One of the two main original intents of defining it as POD is to make clear none of the struct members should contain pointers (because [1]). The other main intent is to ensure trivial copies are possible (again to facilitate broadcasts between multiple user processes). It gets a bit ugly but I could see standard layout minus pointers and minus triviality being ok, but we should document it. 

[1] If I broadcast a (struct containing) pointers from one user process to the other, the destination process would receive an invalid pointer (or struct with invalid pointers) since valid pointers on process A are not necessarily valid on process B.

alberto-scolari:
> Where does it break it? The '20 paper and the current spec clearly defines only POD types are allowed, so at the very least such occurrences would not be compliant and should be fixed.
> 

Tests found so far:
- pinnedVector
- gmres
- gmres_complex
- hpcg

So not a long list after all.

> If I take this as a suggestion to relax the POD requirement to allowing standard layouts, then my main worry is the following: standard layouts in their intent do not preclude the use of pointers-- in fact, intuitively, I would be fine with standard layouts containing pointer data, while I would question that for POD types (even though the C++ spec's fine with it, at least in 11 and 14 I believe).
>
> One of the two main original intents of defining it as POD is to make clear none of the struct members should contain pointers (because [1]). The other main intent is to ensure trivial copies are possible (again to facilitate broadcasts between multiple user processes). It gets a bit ugly but I could see standard layout minus pointers and minus triviality being ok, but we should document it.
> 

From the point of view of pointers allowance, both `is_pod` and `is_standard_layout` are indeed the same. PODs require both trivial constructibility and copyability https://en.cppreference.com/w/cpp/language/classes#Trivial_class, which prevents the user from defining a default constructor ("has one or more eligible default constructors such that each is trivial"). Indeed, the following code snippet

```c++
struct aaa {
	int a;
	aaa() : a( 0 ) {}
};
static_assert( std::is_pod< aaa >::value );
```

does not compile, just because of the user-defined default constructor. Nonetheless, the `aaa` type is trivially copyable.

The idea of POD is that the code defining the type `aaa` works fine both in C and C++, which excludes any constructor (the only admitted constructors are the ones the C++ compiler can "trivially" provide). Hence, whether to require POD depends on the usage scenario: do we allow processes written in C to talk to processes written in C++? I expect not, because the `Launcher` is a facility used within ALP, where C++ is required, and where the execution model is SPMD. Under this assumption, `is_standard_layout` should then suffice. Also, no constraint enforces the absence of pointers (and POD is also deprecated with C++20), so this further constraint can only be mentioned in the specification.

Furthermore, we may further relax the constraint `is_standard_layout` to `is_trivially_copyable`. `is_standard_layout` disallows inheritance (unless all base classes have only static members, which is unlikely). therefore, the following snippet

```c++
struct aaa {
	int a;
	aaa() : a( 0 ) {}
};

struct bbb: aaa {
	float b;
	bbb() : aaa(), b( 0.0f ) {}
};

static_assert( std::is_standard_layout< bbb >::value );
```

does not compile. Changing `is_standard_layout` to `is_trivially_copyable` makes it compile, and is just what we need to serialize objects between ALP SPMD processes. Also, we would have less compilation failures if we check only `is_trivially_copyable` (probably just `pinnedVector`)

On the other side, one practical reason to retain `is_pod` is that we already wrote it in the specs.

anyzelman:
So concretely, relax the constraints to `is_trivially_copyable` and checks for `is_default_constructible` when it is needed right? That seems OK to me.

> Hence, whether to require POD depends on the usage scenario: do we allow processes written in C to talk to processes written in C++

Again, the intent of the constraint here is solely to make broadcasting work in the case of distributed-memory auto-parallelisation. I checked up to the latest C++ standard and did not find type traits to check for the presence of pointer members, so `is_trivially_copyable` indeed seems to correspond to the minimal thing we want *and* can check for (aside from default constructibility in some cases).

I'll have a quick check on what happens in the pinnedVector=)

anyzelman:
Manual tests with LPF running. If pass, will continue with code review

anyzelman (approved):

anyzelman:
Waiting CIs OK, another OK manual re-run of unit & smoke tests with LPF, composition of merge notes, and @alberto-scolari OK on recent changes before merge.

anyzelman:
Merge notes concept:

This MR reviews the implementation of the BSP1D (and hybrid) launcher and benchmarker for all launcher modes (automatic, manual, and from MPI). In so doing it fixes various errors, and given it modifies 42 files, perhaps also answers some things about life, the universe, and everything.

Specification:
- in automatic mode, the output types have to be default constructible,
- in automatic mode with broadcasting and with typed calls, the input type must be default constructible. This can only be caught at run-time and thus the spec now requires an error be returned (`grb::ILLEGAL`) when this situation is encountered.
- the specification did not document all errors that could be thrown by the `grb::Launcher` constructor;
- the specification erroneously asserted that `grb::Launcher::finalize` was necessary. It was not, and indeed tests and examples in the current implementation (as before this MR) never called the function before exiting, without error. The specification has hence been modified to state the use of this function is optional.
- the specification standardised the `grb::Launcher` constructor arguments, but the implemented `FROM_MPI` mode uses a different signature. The spec has been updated to allow for non-standard modes with custom constructor signatures.
- the specification now includes phrasing that the obligatory arguments in automatic and manual mode may have implementation-specific constraints.
- the specification of the launcher and benchmarker classes have been revised for improved overall clarity.

Implementation:
- both the BSP1D launcher and BSP1D benchmarker did not check for default-constructability (DC) of output types when in automatic mode, now fixed (static-assert);
- both the BSP1D launcher and BSP1D benchmarker did not check for DC of input types when in automatic mode, without broadcasting, and when making a typed call, now fixed (run-time check);
- the BSP1D benchmarker now relies on the launcher implementation, strongly reducing the no. of lines required for the former;
- the BSP1D launcher (and thus BSP1D benchmarker) now makes a minimum check on requirements on types that should be broadcast;
- the BSP1D launcher and benchmarker, contrary to the spec, had default broadcast option `true`, now fixed;
- many unit tests erroneously assumed the default broadcast option is `true`, now fixed;
- fixes a manual launch use case that originally spawned issue #235;
- introduces an exhaustive unit test for launchers and benchmarkers of all benchmarks and all combinations of modes and broadcasting options.

Many thanks to Alberto Scolari (@alberto-scolari) for detecting the issue and proposing the fix.

anyzelman:
> Waiting CIs OK, another OK manual re-run of unit & smoke tests with LPF, composition of merge notes, and @alberto-scolari OK on recent changes before merge.

All CIs OK, manual run of unit, smoke, and performance tests on some inputs with LPF OK, concept merge notes done. Waiting for feedback before merge.

PR #246, opened 2023-08-31
655 Provide grb::set( matrix, mask, value )

### LINKED ISSUE #245: Provide grb::set( matrix, value )
Same as [this variant](http://albert-jan.yzelman.net/alp/user/group__IO.html#ga1d6880056a08497b967de557b576bca5) but for Matrix.

### LINKED ISSUE #321: Provide grb::set( matrix, mask, value )

byjtew:
~Closes #245~
Closes #321

This MR includes the following bugfixes:
- copy-contructor of pattern matrix iterators for more than one user processes did not copy PID information correctly;
- the `fold_matrix_to_scalar` unit test relied on undefined behaviour, here fixed;
- matrix reduction into a scalar could fail if the matrix is small w.r.t. the number of user processes;
- selection operators erroneously did not appear in user documentation;
- the specification for resize and execute phases was outdated: if in execution phase it is detected that the output container has insufficient capacity, then primitives should return `ILLEGAL`;
- the BSP1D set-vector-to-value erroneously returned `FAILED` instead, herewith fixed.


include/graphblas/base/io.hpp
```diff
@@ -1078,6 +1078,72 @@ namespace grb {
...
+	 *
+	 * @param[in,out] C The vector of which every element is to be set to equal
+	 *                  \a val. On output, the number of elements shall be equal
+	 *                  to the size of \a C.
+	 * @param[in]   val The value to set each element of \a C to.
+	 * @param[in] phase Which #grb::Phase the operation is requested. Optional;
+	 *                  the default is #grb::EXECUTE.
+	 *
+	 * In #grb::RESIZE mode:
+	 * @returns #grb::SUCCESS  All the time, no de-/allocation is necessary.
```
anyzelman: This is incorrect -- if there is a matrix C with capacity zero, then running this primitive in RESIZE mode must reallocate, is it not?

include/graphblas/bsp1d/io.hpp
```diff
@@ -823,6 +823,49 @@ namespace grb {
 		return ret;
 	}
 
+	template<
```
anyzelman, line 935: The backend should define performance semantics (see other primitives for examples, should be straightforward in this case)

include/graphblas/base/io.hpp
```diff
@@ -1078,6 +1078,72 @@ namespace grb {
...
+	 *   -# grb::descriptors::no_casting
+	 * \endparblock
+	 *
+	 * @param[in,out] C The vector of which every element is to be set to equal
+	 *                  \a val. On output, the number of elements shall be equal
+	 *                  to the size of \a C.
+	 * @param[in]   val The value to set each element of \a C to.
+	 * @param[in] phase Which #grb::Phase the operation is requested. Optional;
+	 *                  the default is #grb::EXECUTE.
+	 *
```
anyzelman, line 1116: Should we include a `\warning` that says on successful execution of this primitive, `C` shall be dense? (Usually, this is not what a user should want)

include/graphblas/bsp1d/io.hpp
```diff
@@ -823,6 +823,49 @@ namespace grb {
...
+		Descriptor descr = descriptors::no_operation,
+		typename DataType, typename RIT, typename CIT, typename NIT,
+		typename ValueType = DataType
+	>
+	RC set(
+		Matrix< DataType, BSP1D, RIT, CIT, NIT  > &C,
+		const ValueType& val,
+		const Phase &phase = EXECUTE
+	) noexcept {
```
anyzelman, line 951: a static_assert for no_casting is missing

include/graphblas/bsp1d/io.hpp
```diff
@@ -823,6 +823,49 @@ namespace grb {
...
+		const size_t m = nrows( C );
+		const size_t n = ncols( C );
+		const size_t nz = nnz( C );
+
```
anyzelman, line 974: dynamic error checks seem missing: for example, a guard against overflow if m * n > 2^64-1. I think in such a case ILLEGAL should be returned, which in turn then also modifies the base specification. This may play for more primitives actually and we should open up an issue for that

include/graphblas/bsp1d/io.hpp
```diff
@@ -823,6 +823,49 @@ namespace grb {
...
+		if( m == 0 || n == 0 || nz == 0 ) { return SUCCESS; }
+
```
anyzelman, line 979: Something like the following is missing:
```
const size_t full_nz = m * n; // assumes overflow is guarded against
if( phase == RESIZE ) {
    if( grb::capacity( C ) < full_nz ) {
        return grb::resize( C, full_nz );
    }
}
```

include/graphblas/bsp1d/io.hpp
```diff
@@ -823,6 +823,49 @@ namespace grb {
...
+			internal::getLocal( C ), val, phase
+		);
+
+		if( collectives< BSP1D >::allreduce( ret, operators::any_or< RC >() )
+			!= SUCCESS
+		) {
+			return PANIC;
+		}
+
+		if( phase == EXECUTE ) {
```
anyzelman: With the above change, this if-statement can become an assert instead (no need to consider `TRY` at the moment).

include/graphblas/bsp1d/io.hpp
```diff
@@ -823,6 +823,49 @@ namespace grb {
...
+			if( ret == SUCCESS ) {
+				ret = internal::updateNnz( x );
+			} else if( ret == FAILED ) {
```
anyzelman: ALP primitives can never return `FAILED`, so this branch should never apply

include/graphblas/bsp1d/io.hpp
```diff
@@ -823,6 +823,49 @@ namespace grb {
...
+				const RC clear_rc = clear( x );
+				if( clear_rc != SUCCESS ) {
+					ret = PANIC;
+				}
+			} else {
+				assert( ret == PANIC );
```
anyzelman, line 1064: If taking the above suggestion, ret could possibly also be `ILLEGAL` to indicate overflow

include/graphblas/hyperdags/io.hpp
```diff
@@ -332,6 +332,38 @@ namespace grb {
 		return ret;
 	}
 
+	template<
```
anyzelman: Perf semantics are missing (just a comment that it inherits the performance semantics of the underlying backend would be sufficient)

include/graphblas/hyperdags/io.hpp
```diff
@@ -332,6 +332,38 @@ namespace grb {
...
+			internal::getMatrix( C ), val, phase
+		);
+		if( ret != SUCCESS ) { return ret; }
+		if( phase != EXECUTE ) { return ret; }
+		if( nrows( C ) == 0 || ncols( C ) == 0 ) { return ret; }
+		std::array< const void *, 0 > sourcesP{};
+		std::array< uintptr_t, 1 > sourcesC{
+			getID( internal::getMatrix(C) )
+		};
+		std::array< uintptr_t, 1 > destinations{
```
anyzelman: I think also val should be registered as a source scalar. There should be multiple examples of that already in the hyperdags backend, for example, for folding a scalar into a vector (foldl/foldr)

include/graphblas/nonblocking/io.hpp
```diff
@@ -1136,6 +1136,59 @@ namespace grb {
 
 	} // end namespace internal::grb
 
+	template<
```
anyzelman: Perf semantics should be added, otoh I'm not sure if nonblocking does that anywhere at the moment. If not, ok for me to skip, but in that case please make an issue of missing perf semantics for nonblocking)

include/graphblas/reference/compressed_storage.hpp
```diff
@@ -501,6 +501,39 @@ namespace grb {
...
+					// The value can not be of type void
+					static_assert(
+						!std::is_void< ValueType >::value,
+						"ValueType must not be void"
+					);
+					// The value must be convertible to D
+					static_assert(
+						std::is_convertible< ValueType, D >::value,
+						"ValueType must be convertible to D"
+					);
```
anyzelman: This is a correct assertion, but for readability of compiler errors, the code standard is to put these static_asserts right at the beginning of each possible entry point (meaning each backend implementation of set should have a static assert like this)

include/graphblas/reference/io.hpp
```diff
@@ -1076,6 +1076,80 @@ namespace grb {
 
 	} // end namespace internal::grb
 
+	template<
```
anyzelman: Performance semantics should be given as doxygen comment here

include/graphblas/reference/io.hpp
```diff
@@ -1076,6 +1076,80 @@ namespace grb {
...
+			), "grb::set (unmasked set to value): ",
+			"called with non-matching value types"
+		);
+
+		// dynamic checks
+		assert( phase != TRY );
+
+		// delegate
+		if( phase == RESIZE ) {
+			return SUCCESS;
```
anyzelman: This return-statement should be something like
```
const size_t full_nz = m * n; // and guard against overflow somewhere
if( grb::capacity( C ) < full_nz ) {
    return grb::resize( C, full_nz );
}
```

include/graphblas/reference/io.hpp
```diff
@@ -1076,6 +1076,80 @@ namespace grb {
...
+
+			const size_t m = nrows( C );
+			const size_t n = ncols( C );
+			const size_t nz = nnz( C );
+
+			if( m == 0 || n == 0 || nz == 0 ) { return SUCCESS; }
+
+#ifdef _H_GRB_REFERENCE_OMP_IO
+			#pragma omp parallel default( none ) \
+				shared( C ) firstprivate( val, m, n, nz )
```
anyzelman: firstprivate makes unnecessary copies of const values?

tests/unit/setMatrixValue.cpp
```diff
@@ -0,0 +1,133 @@
...
+	size_t in = 1000;
+
+	// error checking
+	if( argc > 2 ) {
+		printUsage = true;
+	}
+	if( argc == 2 ) {
+		size_t read;
+		std::istringstream ss( argv[ 1 ] );
+		if( ! ( ss >> read ) ) {
```
anyzelman, line 520: no space after `!` (elsewhere also)

include/graphblas/reference/io.hpp
```diff
@@ -1076,6 +1076,80 @@ namespace grb {
...
+#endif
+			{
+				{ // CRS section
+					size_t range = internal::getCRS( C ).copyFromRange( nz, m );
+					size_t start = 0, end = range;
+#ifdef _H_GRB_REFERENCE_OMP_IO
+					config::OMP::localRange( start, end, 0, range );
+#endif
+					internal::getCRS( C ).assignValue( nz, m, start, end, val );
```
anyzelman: I think this sets all current values to the given one, rather than all values in the matrix. This would be the correct implementation for masked sets, in particular, for `grb::set( C, C, val )`, where the second `C` indicates the matrix itself acts as the mask for the set operation. For `grb::set( C, val )` I would expect the entire matrix `C` to have the given `val` everywhere. (Also for the CCS section ofc.)

Btw if implementing the masked variants, take care to also consider the descriptors that modify the mask behaviours, which could get tricky

anyzelman (changes requested):
Also as the most major comment perhaps: should this MR not include a `grb::set( C, mask, value );`? I would expect that variant to be more useful than the fully dense case, but it may also be a confusion of what (I think) the intended semantics are and what they should be --see also the last added review comment

include/graphblas/reference/io.hpp
byjtew: The idea of this primitive (at leat in my head) was to assign the *existing* non-zero values to the given one. Which would be an alias for `grb::set( input/output=A, mask=A, value=x)`

byjtew:
Goal of this MR changed form now on to a masked-set for matrix

anyzelman:
TODOs:

- [x] Update spec
- [x] Ref impl already seems to exit
- [x] Other impl not, so need to provide
- [x] unit test needs to be extended for masked set to value

include/graphblas/bsp1d/io.hpp
anyzelman: Trivial resolution: removed this entire function;)

include/graphblas/bsp1d/io.hpp
anyzelman: see above

include/graphblas/bsp1d/io.hpp
anyzelman: see above

include/graphblas/bsp1d/io.hpp
anyzelman: see above

include/graphblas/nonblocking/io.hpp
anyzelman: #331 has been created

anyzelman:
At this point, all tests OK. The remaining TODO is to check that the test for the new functionality itself, is OK;)

anyzelman:
Collecting TODOs on this MR, suggested to handle in-order:
- [x] make the self-masked case work. The easiest is to provide a separate function that simply overwrites the values array, and leaves the other arrays alone;
- [x] make the void case work. At present, this should only entail fixing the test to handle the void case more gracefully;
- [x] the unit test should be expanded to test the structural and invert_mask descriptor;
- [x] the implementation of grb::set should call a different function for handling the structural and invert_mask cases. A separate generic function is needed since the current generic one cannot efficiently filter out such elements since it is based on raw copying of the mask's structure. This new generic function should also gracefully handle the self-masked case. Note that the spec disallows self-masking under mask inversion, so this should be possible without any intermediate buffers.

anyzelman:
MPICH 3.3.2 detects an issue with the fold_matrix_to_scalar test, and indeed that test implementation does not conform the ALP standard. It should also be fixed-- decide later if in this issue or in a separate one.

Just now in the wee hours made the self-masked case work. This triggered some bugs here and there now also resolved.

anyzelman:
Update: at this point, all unit tests with LPF ok except fold_matrix_to_scalar. All smoke tests with LPF OK. (MPICH 3.3.2)

anyzelman:
> MPICH 3.3.2 detects an issue with the fold_matrix_to_scalar test, and indeed that test implementation does not conform the ALP standard. It should also be fixed-- decide later if in this issue or in a separate one.

This is confirmed fixed in this MR

anyzelman:
All tests OK at this point (using the "difficult" MPICH version)

anyzelman:
All tests OK at this point (using the "difficult" MPICH version)

anyzelman:
> All tests OK at this point (using the "difficult" MPICH version)

At this point also. Now only need to increase unit test robustness and this one is done.

anyzelman:
Running the tests again. If OK, this MR is done (the tests now generate and check for differing outputs for different descriptors)

anyzelman:
> Running the tests again. If OK, this MR is done (the tests now generate and check for differing outputs for different descriptors)

Tests were ok but code review shows an issue that should cause the non-structural self-masked case to not work. Fixing the implementation first, and then will look at why the unit test doesn't fail because of it.

anyzelman (approved):
Code review OK, waiting for tests and internal CI to flag OK as well before merge.

anyzelman:
Concept release notes:

This MR formalises the `grb::set(matrix,matrix)` and `grb::set(matrix,mask,scalar)` variants. For the former, all backends did implement it and an exhaustive unit test existed prior to this MR, but a specification of the function was missing-- this MR adds this. For the latter, all backends did implement one variant of it which was tested in the unit test for the matrix-matrix `grb::set`. This MR specifies and implements all possible variants of `grb::set(matrix,mask,scalar)`, and adds a unit test that exhaustively tests all combinations under which `grb::set(matrix,mask,scalar)` may be called.

To realise this MR, the following features were introduced:

- the ALP utilities now include a prefix-sum algorithm, both sequential and in OpenMP. The implementation of some of the new functionalities depend on this operation. Existing prefix-sum computations should rely on this new utility function; GitHub issue #332 is to follow up on that.
- ALP objects now include ALP types such as `grb::Phase`. This was required to unambiguously dispatch between `grb::set(matrix,matrix)` and `grb::set(matrix,mask,scalar)` via SFINAE.

Additionally, this MR includes the following bugfixes:

1. copy-contructor of pattern matrix iterators for more than one user processes did not copy PID information correctly;
2. the `fold_matrix_to_scalar` unit test relied on undefined behaviour, here fixed;
3. matrix reduction into a scalar could fail if the matrix is small w.r.t. the number of user processes;
4. selection operators erroneously did not appear in user documentation;
5. the specification for resize and execute phases was outdated: if in execution phase it is detected that the output container has insufficient capacity, then primitives should return `ILLEGAL`;
6. the BSP1D set-vector-to-value erroneously returned `FAILED` instead, herewith fixed.

The impact of bugs 2-4 is expected very low as it was only present in the `develop` branch and only for a limited time. Issues 1 and 5 were present longer, however, prior to crafting this MR and as far as we know, seem never to have been triggered or noticed. Issue 5 furthermore did not induce any code changes, except for the bug in issue no. 6. That bug (no. 6), however, never triggers in the currently available back-ends because vectors have fixed full capacities (the bug would only apply for final backends that do not have such a fixed full capacity).

Last, and as always, the MR minor includes code style and documentation fixes in files related and unrelated to this MR.

Many thanks to Benjamin Lozes @byjtew for drafting the initial version of the MR!

anyzelman:
Manual unit & smoke tests with the "difficult" MPI all OK, internal CI OK, GitHub CI OK -- triggering merge=)

PR #248, opened 2023-09-01
656 grb::set vector empty-masked is broken

### LINKED ISSUE #245: Provide grb::set( matrix, value )

### LINKED ISSUE #247: grb::set vector empty-masked is broken
The buggy section below appears in some pieces of code, to check and fix.

```c++
template<
	typename DataType, typename MaskType, typename T,
	typename Coords
>
RC set(
	Vector< DataType, reference, Coords > &x,
	const Vector< MaskType, reference, Coords > &m,
	const T val,
	const Phase &phase = EXECUTE,
		!grb::is_object< DataType >::value && !grb::is_object< T >::value,
	void >::type * const = nullptr
) {
#ifdef _DEBUG
	std::cout << "In grb::set (vector-to-value, masked)\n";
#endif


	//...

        
// FROM HERE
	// If the mask is empty (no value to assign)
	if( size( m ) == 0 ) {
                // SHOULD INSTEAD CALL: clear( x, phase );
		return set< descr >( x, val, phase );
	}
// TO HERE


	// ...
}
```

byjtew:
Closes #247 

Would require #246 to be merged before, as this particular case could be tested into the new unit test.

### DIFF
--- a/include/graphblas/nonblocking/io.hpp
+++ b/include/graphblas/nonblocking/io.hpp
@@ -388,9 +388,9 @@ namespace grb {
 			"vector"
 		);
 
-		// catch empty mask
+		// If the mask is empty: clear the vector
 		if( size( m ) == 0 ) {
-			return set< descr >( x, val, phase );
+			return clear( x );
 		}
 
 		// dynamic sanity checks
--- a/include/graphblas/reference/io.hpp
+++ b/include/graphblas/reference/io.hpp
@@ -469,9 +469,9 @@ namespace grb {
 			"vector"
 		);
 
-		// catch empty mask
+		// If the mask is empty: clear the vector
 		if( size( m ) == 0 ) {
-			return set< descr >( x, val, phase );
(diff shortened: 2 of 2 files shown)


include/graphblas/nonblocking/io.hpp
```diff
@@ -388,9 +388,9 @@ namespace grb {
 			"vector"
 		);
 
-		// catch empty mask
+		// If the mask is empty: clear the vector
 		if( size( m ) == 0 ) {
-			return set< descr >( x, val, phase );
```
anyzelman, line 393: This is actually correct: `size( m ) == 0` is code for `NO_MASK`, so the this was correct

anyzelman:
See the review comment -- the behaviour was correct, closing without merge

Issue #264, opened 2023-11-20
Repeated performance tests should skip repeated file I/O


anyzelman:
The old performance testing setup also tests for I/O performance. For large input files, this drives up the required experiment time, as I/O tends to be by far the slowest component in every benchmark. This issue proposes two fixes:

- [x] every smoke test should read inputs only once. To cope with variability under differing allocations (esp. detected on ARM), the file should be read to an in-memory storage (e.g., an `std::vector` that stores a nonzero struct for each nonzero, or multiple `std::vector`s). The remainder smoke tests then remain unchanged, except every experiment now reads from this in-memory storage, but otherwise still performs new allocations for matrix and vector data;

- [ ] a new performance test should be introduced that tests I/O only -- i.e., reads the input multiple times, and reporting all the regular timing statistics (average, min, max, unbiased sample standard deviation). 

Thus, the new performance test shall form the only exception to the first item, retains I/O speed benchmarking capabilities, and speeds up all other performance testing in future.

anyzelman:
First item is done, second item has lower priority. De-assigning myself so that the remaining item is open for anyone else to pick up:)

PR #265, opened 2023-11-20
264 repeated performance tests should skip repeated file io

anyzelman:
This MR resolves the first item of issue #264, and addresses the ALP performance testing setup, in particular all performance tests that operate on input files.

The situation before this MR is that each experiment also tests I/O performance. For large input files, this drives up the required performance test time, as I/O tends to be by far the slowest component in every benchmark. Yet when running e.g. a performance test for BiCGstab, one is typically interested in the performance of BiCGstab, not in the performance of I/O. With "experiments", we here mean here the so-called "outer iterations" of the many performance tests included with ALP.

This MR modifies the performance testing framework to perform file I/O only once. During this initial I/O phase, the file contents are buffered in memory. The actual experiments will then repeatedly read from the in-memory file storage. We do not simply load and parse the file to an ALP container once and then keep reusing the same parsed containers, as on some platforms, the precise allocation (which can't always be precisely controlled) may have significant effect on performance-- and this is exactly the kind of effect that having a single benchmark execute the same experiment multiple times was designed to measure.

This MR also, as a bonus, includes
1. code style fixes in the Pregel interface: `num_vertices` -> `numVertices`, `num_edges` -> `numEdges`, `get_matrix` -> `getMatrix`. **This change does affect ALP/Pregel user code!**
2. a singleton utility class has been added,
3. code style review of all smoke and performance tests, as well as,
4. the usual smaller code style fixes in unrelated files. In particular, some filenames erroneously employed CamelCase naming, while they should have used camelCase (or maybe, rather, dromeDary).


tests/unit/distribution_bsp1d.cpp
```diff
@@ -117,41 +165,57 @@ int main( int argc, char ** argv ) {
...
 		localsize[ s ] += localsize[ s - 1 ];
 	}
 	for( size_t s = 0; s < P; ++s ) {
-		if( localsize[ s ] != internal::Distribution< BSP1D >::local_offset( n, s, P ) ) {
+		if( localsize[ s ] !=
+			internal::Distribution< BSP1D >::local_offset( n, s, P )
+		) {
 			error = static_cast< int >( 14 + s + P );
-			(void)fprintf( stderr,
+			(void) fprintf( stderr,
```
anyzelman: Enter before `stderr`

include/graphblas/utils/singleton.hpp
```diff
@@ -0,0 +1,79 @@
...
+		 * ::getData. The singleton includes storage.
+		 *
+		 * If multiple singletons of the same data type \a T are required, then each
+		 * such singleton should define a unique \a key.
+		 *
+		 * @tparam T   The data type of the singleton.
+		 * @tparam key The identifier of this singleton.
+		 *
+		 * The type \a T must be default-constructible.
+		 */
```
anyzelman, line 62: Add a warning that this should be used as sparsely as possible, and in any case only by the final application directly (it should not be used as an implementation component in ALP, nor in any library in fact)

include/graphblas/utils/singleton.hpp
```diff
@@ -0,0 +1,79 @@
...
+		template< typename T, size_t key = 0 >
+		class Singleton {
+
+			private:
+
+				/**
+				 * \internal Make sure no-one except ::getData can construct an instance.
```
anyzelman: ::getData also does not construct an instance.

tests/smoke/conjugate_gradient.cpp
```diff
@@ -25,24 +25,57 @@
...
-
 using BaseScalarType = double;
 #ifdef _CG_COMPLEX
  using ScalarType = std::complex< BaseScalarType >;
 #else
  using ScalarType = BaseScalarType;
 #endif
 
+/** Parser type */
+typedef grb::utils::MatrixFileReader< ScalarType,
```
anyzelman: enter before ScalarType

tests/smoke/gmres.cpp
```diff
@@ -305,34 +431,28 @@ void grbProgram( const struct input &data_in, struct output &out ) {
...
+			const RC rc = buildMatrixUnique(
+				P,
+				utils::makeNonzeroIterator<
+					grb::config::RowIndexType, grb::config::ColIndexType, ScalarType
+				>( data.cbegin() ),
+				utils::makeNonzeroIterator<
+					grb::config::RowIndexType, grb::config::ColIndexType, ScalarType
+				>( data.cend() ),
+				PARALLEL
+			);*/
```
anyzelman, line 455: This code block should be in the directly preceding inner code block

tests/smoke/gmres.cpp
```diff
@@ -693,6 +831,6 @@ int main( int argc, char **argv ) {
 	std::cout << std::endl;
 
 	// done
-	return out.rc;
+	return (50 + out.rc);
```
anyzelman: Unless out.rc is zero, then this should return 0 also, not 50. So in fact this code should read something like

`return (out.rc == 0 ? 0 : (50 + out.rc));`

tests/smoke/hpcg.cpp
```diff
@@ -35,36 +35,43 @@
...
 #ifdef HPCG_PRINT_STEPS
 
 // HPCG_PRINT_STEPS requires defining the following symbols
 
-/**
- * @brief simply prints \p args on a dedicated line.
- */
-#define DBG_println( args ) std::cout << args << std::endl;
+ /**
+  * @brief simply prints \p args on a dedicated line.
```
anyzelman: Remove all uses of `@brief`

tests/unit/auto_launcher.cpp
```diff
@@ -29,6 +29,7 @@
                           "be negative. This program must be called #processes times " \
 			  "on any number of connected nodes.\n";
 
+
```
anyzelman, line 26: This empty lines should be above the preceding macro defines

anyzelman: (in fact, removing those defines may be nicest)

tests/smoke/gmres.cpp
anyzelman: actually no, it is in the correct place -- it seems git diff was misleading here. Resolving without any action

anyzelman:
At this point, several (most?) smoke tests are now caching the files in-memory, after reading them. Those smoke tests then report as I/O the ingestion time from memory. At this point, all smoke tests (with LPF) pass, and a code review has been conducted. What remains is to 1) identify other smoke or performance tests that may perform repeated file I/O when benchmarking something, and 2) to also have them switch over to the use of singletons for file caching.

anyzelman:
Review OK

anyzelman:
GitHub CI OK, smoke tests with LPF OK, waiting for internal CI.

anyzelman:
Testing found some bugs, now fixed. Code review OK.

Waiting CIs, waiting manual smoke and perf tests (with LPF).

anyzelman:
All CIs OK, manual smoke *and* performance tests with LPF OK. Will merge.

PR #267, opened 2023-11-30
The non-blocking backend is using sysconf

byjtew:
Probably an oversight. 
Noticed during benchmarking an absolutely constant scalability.


anyzelman (approved):
OK

PR #271, opened 2023-12-01
What a selection operator could look like

### LINKED ISSUE #270: Provide grb::select( [out] Matrix, [in] Matrix, grb::operator(r,c,v) -> bool )
The `select` primitive(s) is required for multiple incoming algorithms. 
For instance to threshold under/over a specified value, or in `tril` in order to filter depending on the coordinates.

The idea would be to generalise the incoming `tril`, `triu` and `diag` primitives.
- `grb::tril( L, A )` -> `grb::select( L, A, grb::boolean_operators::is_lower_part<RIT, CIT>() )`
- `grb::triu( U, A )` -> `grb::select( U, A, grb::boolean_operators::is_upper_part<RIT, CIT>() )`
- `grb::diag( D, A )` -> `grb::select( D, A, grb::boolean_operators::is_diag<RIT, CIT>() )`
- `grb::diag( A )` -> `grb::select( A, grb::boolean_operators::is_diag<RIT, CIT>() )`

The **diag** boolean_operator would then have the following `apply` function:
```c++
template< 
  typename RIT, typename CIT, typename T,
  typename BoolOrEquivalent = bool
>
BoolOrEquivalent apply( RIT i, CIT j, T value ) {
  return static_cast<BoolOrEquivalent>( i == j );
}
```

The naming is to be discussed, of course.

byjtew:
Closes #270 

- [x] Wrapping the selection operators in their own namespace to avoid confusion: `grb::operators::select::`
- [x] Implement selectLambda
- [x] Distributed backends implementation

- In-place variants (should come in a separate PR)
- Masked variants (should come in a separate PR)
- Pattern matrices support (should come in a separate PR)


byjtew:
Next steps:
- [ ] Wrapping the selection operators in their own namespace to avoid confusion
- [ ] Implement selectLambda
- [ ] In-place variants (may be in a separate PR)
- [ ] Masked variants (may be in a separate PR)

anyzelman:
The current state does not have bsp1d nor hybrid implementations of select, that's correct right? I think this is a super useful enhancement, but before merge it should have those distributed-memory implementations also

byjtew:
> The current state does not have bsp1d nor hybrid implementations of select, that's correct right? I think this is a super useful enhancement, but before merge it should have those distributed-memory implementations also

Absolutely, I also wanted to switch the namespace from `operators::` to `select_operators::` or `operators::select::` since they have different signatures than usual `operators::`

I'll ping you once it will be implemented

anyzelman:
> > The current state does not have bsp1d nor hybrid implementations of select, that's correct right? I think this is a super useful enhancement, but before merge it should have those distributed-memory implementations also
> 
> Absolutely, I also wanted to switch the namespace from `operators::` to `select_operators::` or `operators::select::` since they have different signatures than usual `operators::`
> 
> I'll ping you once it will be implemented

maybe just `grb::select`?

byjtew:
> maybe just `grb::select`?

But then it could be either the primitive or the operator's namespace, what about `grb::select::operators` ?

anyzelman:
> > maybe just `grb::select`?
> 
> But then it could be either the primitive or the operator's namespace, what about `grb::select::operators` ?

Oh right.. It could still work by having select as a class with the "primitive call" via `operator()` and the pre-defined lambdas as public static members, but that's perhaps a bit hacky. Maybe just think of something and we can look at it in more detail later, once it's more or less ready?

anyzelman:
Also doc/spec should be added to base/blas3.hpp

include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,78 @@ namespace grb {
...
+		typename Tin,
+		typename RITin, typename CITin, typename NITin,
+		typename Tout,
+		typename RITout, typename CITout, typename NITout,
+		Backend backend
+	>
+	RC select(
+		Matrix< Tout, backend, RITout, CITout, NITout >& out,
+		const Matrix< Tin, backend, RITin, CITin, NITin >& in,
+		const SelectionOperator &op,
```
anyzelman, line 550: isn't it nicer to typedef `SelectionOperator` as an `std::function` with a specific signature, which would then no longer require `selectLambda` as a separate primitive? (Different from `grb::algorithms`, we want to have an absolute minimum number of core primitives.)

include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,78 @@ namespace grb {
 		return ret == SUCCESS ? UNSUPPORTED : ret;
 	}
 
+	template<
```
anyzelman, line 538: Doc is still missing?

include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,78 @@ namespace grb {
...
+		(void) descr;
+		(void) out;
+		(void) in;
+		(void) op;
+		(void) phase;
+#ifdef _DEBUG
+		std::cerr << "Selected backend does not implement grb::select\n";
+#endif
+#ifndef NDEBUG
+		const bool selected_backend_does_not_support_select	= false;
```
anyzelman: tab instead of a space before `=` (elsewhere also)

include/graphblas/base/blas3.hpp
anyzelman: I now remember a concern about the performance of lambda functions and std::functions in particular.

There is another alternative, which is to specify that the `SelectionOperator` is a functor, i.e., a class of the following form:

```
template< typename RIT, typename CIT, typename D >
class MySelectionOperator {

    public:

        static bool operator()( const RIT i, const CIT j, const D v ) {
               // some logic based on i, j, v returning a bool
        }

};
```

I would have to check, but I suspect that the use of this (i.e., in the select function, there would be calls like `if( SelectionOperator( i, j, v ) ) { ... }`) would work even if SelectionOperator were an std::function or a lambda function, isn't it?

include/graphblas/base/internalops.hpp
```diff
@@ -4276,6 +4276,167 @@ namespace grb {
...
+
+			template< typename RIT1, typename CIT1, typename D1 >
+			static bool apply(
+				const RIT1 & x, const CIT1 & y, const D1 &
+			) {
+				const RIT a = static_cast< RIT1 >( x );
+				const CIT b = static_cast< CIT1 >( y );
+				return OP::apply( &a, &b, nullptr );
+			}
+		};
```
anyzelman: If following the above suggestion, then these base classes are not required

include/graphblas/base/internalops.hpp
```diff
@@ -4276,6 +4276,167 @@ namespace grb {
...
+			typedef D value_type;
+			typedef RIT row_type;
+			typedef CIT column_type;
+
+			static bool apply(
+				const row_type * __restrict__ const x,
+				const column_type * __restrict__ const y,
+				const value_type * __restrict__ const v
+			) { return not is_strictly_lower< D, RIT, CIT >::apply( x, y, v ); }
+		};
```
anyzelman: If following the above suggestion, then instead of `apply(`, these should be `static bool operator()(`

include/graphblas/ops.hpp
```diff
@@ -981,245 +981,339 @@ namespace grb {
 
 	} // namespace operators
 
-	template< typename D1, typename D2, typename D3, enum Backend implementation >
+	/**
+	 * This namespace holds various standard selection-operators such
+	 * as #grb::operators::is_diagonal or #grb::operators::is_strictly_lower.
+	 */
+	namespace operators::select {
```
anyzelman: I suggest `namespace selectors` (i.e., put the functors in `grb::selectors`)

include/graphblas/ops.hpp
```diff
@@ -981,245 +981,339 @@ namespace grb {
...
+
+		template<
+			typename D,
+			typename RIT = config::RowIndexType,
+			typename CIT = config::ColIndexType
+		>
+		class is_diagonal : public internal::MatrixSelectionOperatorBase<
+			internal::is_diagonal< D, RIT, CIT >, D
```
anyzelman: If following the above suggestion, then a direct functor implementation in `ops.hpp` that does not require this inheritance + reflection becomes possible.

The reason why ops.hpp is more complicated is because 1) it automatically derives foldl/foldr and array-based operations when no explicit ones are given, 2) because there is an algebraic type traits system, and 3) because the operators need to be composed into richer algebraic constructs. I don't think we have any of such a view or requirement for selectors right, so we can pick a simpler solution instead?

include/graphblas/reference/blas3.hpp
```diff
@@ -61,6 +61,200 @@ namespace grb {
...
+					config::OMP::localRange( start_row, end_row, 0, m );
+#endif
+					for( auto i = start_row; i < end_row; ++i ) {
+						for( auto k = in_raw.col_start[ i ]; k < in_raw.col_start[ i + 1 ]; ++k ) {
+							const auto j = in_raw.row_index[ k ];
+							const auto r = i + in_coordinates.first;
+							const auto c = j + in_coordinates.second;
+							const auto value = in_raw.getValue( k, identity );
+							if( op.apply( r, c, value ) ) {
+								nzc++;
```
anyzelman: `(void)` before `nzc++`

include/graphblas/reference/blas3.hpp
```diff
@@ -61,6 +61,200 @@ namespace grb {
...
+			if( m == 0 || n == 0 ) {
+				return SUCCESS;
+			}
+
+			if( phase == RESIZE ) {
+				size_t nzc = 0;
+
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+				#pragma omp parallel default(none) \
+					shared(out, in_raw, rc) firstprivate(m, op, identity, in_coordinates) \
```
anyzelman: `firstprivate` copies the given values, which in this case seem unnecessary copies. The default share-everything would be more performant?

include/graphblas/reference/blas3.hpp
```diff
@@ -61,6 +61,200 @@ namespace grb {
...
+					reduction(+:nzc)
```
anyzelman: I don't think the OpenMP standard specifies the performance semantics of reduction -- it is likely to be implementation-dependent. Therefore an OpenMP run-time implementation may dynamically allocate memory to facilitate the reduction. Instead relying on the internal ALP/GraphBLAS buffer and explicitly doing the reduction prevents this possibility and ensures the implementation adheres to the performance semantics (which, although presently not specified, is likely to include the promise that no system calls such as malloc or free are made).

include/graphblas/reference/blas3.hpp
```diff
@@ -61,6 +61,200 @@ namespace grb {
...
+
+				// Prefix sum of CCS.col_start
+				for( size_t j = 2; j < n + 1; ++j ) {
+					out_ccs.col_start[ j ] += out_ccs.col_start[ j - 1 ];
+				}
+
+				// Initialise the column counter array with zeros
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+				#pragma omp parallel for simd default(none) \
+					shared(col_counter) firstprivate(in_coordinates, n)
```
anyzelman: note `firstprivate` again (possibly elsewhere as well)

In this case also `in_coordinates` seem unused. In particular the use of `default(none)` in my experience always ends up annoying to maintain because code changes more often than one might think affects the shared and private clauses-- which might be what happened here (if an earlier version did use `in_coordinates`)

include/graphblas/reference/blas3.hpp
```diff
@@ -61,6 +61,200 @@ namespace grb {
...
+						const auto idx = col_counter[ j ] + out_ccs.col_start[ j ];
+						out_ccs.row_index[ idx ] = i;
+						out_ccs.setValue( idx, value );
+						++col_counter[ j ];
+					}
+
+					// Update CRS
+					out_crs.row_index[ nzc ] = j;
+					out_crs.setValue( nzc, value );
+					++nzc;
```
anyzelman: `(void)` here (and elsewhere possibly also)

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,308 @@
...
+	NOTE:
+	This function will fail for distributed backend because the local iterator of the matrix
+	does not reflect the global coordinates, which can lead to false negatives.
+	*/
+	bool valid = true;
+	for( const auto &each : B ) {
+		const auto entry = getMatrixEntry<D>(each);
+		if( not predicate.apply( std::get<0>(entry), std::get<1>(entry), std::get<2>(entry) ) ) {
+			std::cerr << "  /!\\ Predicate failed for ("
+				<< std::get<0>(entry) << ", " << std::get<1>(entry) << ", " << std::get<2>(entry) << ")"
```
anyzelman: line overflow?

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,308 @@
...
+) {
+	/*
+	NOTE:
+	This function will fail for distributed backend because the local iterator of the matrix
+	does not reflect the global coordinates, which can lead to false negatives.
+	*/
+	bool valid = true;
+	for( const auto &each : B ) {
+		const auto entry = getMatrixEntry<D>(each);
+		if( not predicate.apply( std::get<0>(entry), std::get<1>(entry), std::get<2>(entry) ) ) {
```
anyzelman: use `!` instead of `not`

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,308 @@
...
+			valid = false;
+			break;
+		}
+	}
+	assert( collectives<>::allreduce( valid, operators::logical_and<bool>() ) == SUCCESS );
+
+	return valid;
+}
+
+template<typename D, typename SelectionOperator, typename RIT, typename CIT, Backend implementation>
```
anyzelman: spaces after (before) < (>)

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,308 @@
...
+		if( rc != SUCCESS ) {
+			std::cerr << "(lambda variant): EXECUTE phase of test <" << test_name <<
+				"> failed, rc is \"" << toString(rc) << "\"" << std::endl;
+			return rc;
+		}
+
+		grb::wait( output );
+		printSparseMatrix<Debug>(output);
+
+		const bool valid = matrix_validate_predicate(output, op);
```
anyzelman: spaces after `(` and before `)`

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,308 @@
...
+template< typename D, typename RIT, typename CIT, typename NIT, Backend implementation >
+void buildMatrixUniqueWrapper(
+	Matrix<D, implementation, RIT, CIT, NIT>& mat,
+	const size_t* row_indices,
+	const size_t* col_indices,
+	const size_t nvals,
+	const IOMode io_mode,
+	typename std::enable_if< not std::is_void<D>::value >::type * = nullptr
+) {
+	std::vector<D> values(nvals, 1);
```
anyzelman: here and perhaps elsewhere too

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,308 @@
...
+		std::vector<size_t> const_indices_zero(n, 0);
+		std::vector<size_t> iota_indices(n, 0);
+		std::iota(iota_indices.begin(), iota_indices.end(), 0);
+		std::vector<size_t> reverse_iota_indices(n, 0);
+		for (long i = n-1; i >= 0; i--) { reverse_iota_indices[i] = i; }
+
+		buildMatrixUniqueWrapper(I, iota_indices.data(), iota_indices.data(), n, SEQUENTIAL);
+		printSparseMatrix<Debug>(I, "identity");
+
+		buildMatrixUniqueWrapper(I_transposed, iota_indices.data(), reverse_iota_indices.data(), n, SEQUENTIAL);
```
anyzelman: line overflow?

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,308 @@
...
+		if (out != SUCCESS) {
+			STDERR_WITH_LINE << "Test FAILED (" << toString(out) << ")" << std::endl;
+			return out;
+		}
+	}
+
+	std::cout << std::flush;
+	std::cerr << std::flush << "Test OK" << std::endl;
+	return 0;
+}
```
anyzelman, line 652: missing empty line at the bottom?

include/graphblas/bsp1d/blas3.hpp
```diff
@@ -205,6 +205,86 @@ namespace grb {
...
+		if( phase == RESIZE ) {
+			if( collectives<>::allreduce( ret, operators::any_or< RC >() ) != SUCCESS ) {
+				return PANIC;
+			} else {
+				return SUCCESS;
+			}
+		}
+		assert( phase == EXECUTE );
+		return internal::checkGlobalErrorStateOrClear( out, ret );
+	}
```
anyzelman: I don't think this is a correct implementation -- BSP1D employs a block-cyclic distribution, so local coordinates (i,j) are not simply an offset away from the global coordinates. Other primitives (like mxv) resolve this by passing a lambda that describes how to translate local to global coordinates. That lambda could replace the offset that you currently have implemented. Also, the test should likely be expanded to run into this error (apparently at present it does not right?)

anyzelman: Passing in a generic distribution / how to get from local to global has the added benefit that it will work with any other distributed-memory backend (which may make other distribution choices)

anyzelman (changes requested):
Thanks for this useful primitive!

I think this one can be simplified as in the comments. The suggested simplifications should benefit both usage as well asimplementation.

include/graphblas/base/blas3.hpp
byjtew: Indeed and I wanted to do that at first, but I copied the logic of the current operators in case someone absolutely needed to write the word *apply*

include/graphblas/base/internalops.hpp
byjtew: Indeed

include/graphblas/base/blas3.hpp
byjtew: Done.

include/graphblas/base/internalops.hpp
byjtew: For the void specialisation they are

include/graphblas/ops.hpp
byjtew: Type handling is easier this way, otherwise you need to specify it explicitly

include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,236 @@ namespace grb {
...
+	 * Any old entries of \a B will be removed after a successful call to this
+	 * primitive; that is, this primitive is out-of-place.
+	 *
+	 * \note When applying selection on sparse matrices using operators,
+	 *       there is a difference between annihilating already existing
+	 *       non-zero values and accumulating them. This functionality is
+	 *       provided by #grb::select depending on whether one or two
+	 *       operators are provided:
+	 *        - #grb::select( B, A, op, phase ): annihilating,
+	 *        - #grb::select( B, A, op, op, phase ): accumulating.
```
anyzelman: I think the latter variant was never implemented right? (I'm deleting it during review, please let me know if that was premature @byjtew )

byjtew: Absolutely

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,393 @@
...
+bool matrix_validate_predicate(
+	const Matrix< D, implementation, RIT, CIT > &B,
+	Func predicate,
+	typename std::enable_if< !std::is_void< D >::value >::type * = nullptr
+) {
+	/*
+	NOTE:
+	This function will fail for distributed backend because the local iterator of the matrix
+	does !reflect the global coordinates, which can lead to false negatives.
+	*/
```
anyzelman: @byjtew to check, this comment is outdated right?

tests/unit/selectMatrix.cpp
```diff
@@ -0,0 +1,393 @@
...
+			STDERR_WITH_LINE << "Launching test FAILED\n";
+			return 255;
+		}
+		if (out != SUCCESS) {
+			STDERR_WITH_LINE << "Test FAILED (" << toString(out) << ")" << std::endl;
+			return out;
+		}
+	}
+
+	{ // To be implemented
```
anyzelman: And this comment too?

anyzelman:
Note: during review, I was trying out lambda-based selection. I'm working on debugging what happens when modifying the unit tests to test that functionality, while also working on simplifying the interface to cope with both pre-defined selection ops as well as arbitrary lambdas. It's getting late though, so will continue this tomorrow.

tests/unit/selectMatrix.cpp
byjtew: Yes

tests/unit/selectMatrix.cpp
byjtew: Yes

anyzelman:
This one is ready to merge (after conflict resolution)

anyzelman:
> This one is ready to merge (after conflict resolution)

Correction: unit tests still detect a bug in BSP1D, fixing it is TODO

anyzelman:
Fixed

anyzelman (approved):

anyzelman:
Some final tests (with LPF) are running, as is the CI. Will merge if all OK.

Concept release notes:

This MR introduces the matrix selection feature -- given a operator that takes matrix nonzero coordinates and a value as arguments and produces a boolean, elements from an input matrix are selected to be copied in some output matrix. This primitive operates in an out-of-place fashion. In-place semantics may be achieved via a temporary (pattern) matrix and fold, and thus are currently not planned to be added. Standard matrix selection operators such as (strictly) lower triangular etc. are provided, including type traits for them. Non-standard selection operators provided by users are simple to add as the new `grb::select` primitive accepts arbitrary lambda functions as a selection operator as well.

This MR also fixes a bug for the BSP1D and hybrid eWiseApply on matrices with the resize mode, which could return `SUCCESS` even if it was not successful. The MR includes unit tests for matrix distributions and the new select operator, as well as code style fixes in related and unrelated files.

Issue #275, opened 2023-12-18
Internal CI performance tests time out


anyzelman:
Since latest merge ( #264 ) . Should investigate the cause and eliminate any bug that may have been introduced.

anyzelman:
Last successful perftests by the internal CI was three weeks ago, indeed just prior to merging #264, job ID 124139. Back then running the performance tests took 40 minutes. The present time-out happens at ~2 hours. Manually running the performance tests shows no obvious issue.

I've now replicated the data sets available in the CI and started a timed run. If the timing is indeed at 2+ hours, I'll compare the log versus that of the last successful run in the CI to see if e.g. new tests were added, and if not, to see what tests run significantly slower since then.

anyzelman:
What I will note though is that parsing cit-HepTh.txt is extremely slow, I surmise due to the use of indirect mapping. However, this behaviour should also be incurred prior to the merge of #264. Nevertheless, noting it here in case of running out of more likely culprits.

anyzelman:
> What I will note though is that parsing cit-HepTh.txt is extremely slow, I surmise due to the use of indirect mapping. However, this behaviour should also be incurred prior to the merge of #264. Nevertheless, noting it here in case of running out of more likely culprits.

This one is actually confirmed to be the issue. Tracking down what happened...

anyzelman:
Hypothesis: now the buildMatrixUnique with random access iterators is enabled, rather than the purely sequential one. This then triggers issue #276, which then, in turn, runs in the issue that the parallelised reads are orders of magnitude slower than the sequential one. Trying to confirm this hypthesis first.

anyzelman:
> Hypothesis: now the buildMatrixUnique with random access iterators is enabled, rather than the purely sequential one. This then triggers issue #276, which then, in turn, runs in the issue that the parallelised reads are orders of magnitude slower than the sequential one. Trying to confirm this hypthesis first.

disproved.

Unlikely hypothesis, but checking it anyway: the old benchmark ignored the indirect mapping request

anyzelman:
> > Hypothesis: now the buildMatrixUnique with random access iterators is enabled, rather than the purely sequential one. This then triggers issue #276, which then, in turn, runs in the issue that the parallelised reads are orders of magnitude slower than the sequential one. Trying to confirm this hypthesis first.
> 
> disproved.
> 
> Unlikely hypothesis, but checking it anyway: the old benchmark ignored the indirect mapping request

[...]

PR #277, opened 2023-12-18
275 matrix file iterator should include properties by reference

anyzelman:
The MatrixFileIterator for void-valued nonzero types copied-by-value the matrix properties. It should instead hold a reference to those properties. This manifested as a performance bug in `develop` after MR #264 that caused our internal CI to take more than 2 hours on a cursory performance test suite, instead of roughly 40 minutes.

MR #264 itself contained no issues-- all it did was include a programming pattern that kept re-creating an end-iterator, which, in turn, kept creating deep copies of the maps in the matrix file property class. Those maps furthermore are only populated when in indirect parsing mode.

This MR fixes the bug and includes code style fixes to the matrix parser files.

### DIFF
--- a/include/graphblas/utils/parser/matrixFileIterator.hpp
+++ b/include/graphblas/utils/parser/matrixFileIterator.hpp
@@ -720,7 +720,7 @@ namespace grb {
 						}
 						return buffer[ pos ].second;
 					}
-				};
+			};
 
 			/** An iterator over nonzeroes in the matrix file. Pattern matrix specialisation. */
 			template< typename S >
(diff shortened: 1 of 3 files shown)


anyzelman:
Code review OK. Checked that it fixes issue #275 . GitHub CI OK. Smoke tests without LPF OK. Waiting smoke tests with LPF and waiting internal CI before merge.

anyzelman:
Smoke tests with LPF OK, waiting internal CI

anyzelman:
Internal CI OK, will merge

PR #281, opened 2024-01-22
258 expose a transition path for solvers

### LINKED ISSUE #258: Expose a transition path for solvers
Initially, for demo-ing what that looks like.

alberto-scolari:
Adding interfaces to expose solvers on ALP transition path:
- (pre-existing) solver interface `solve_cg_*`
- (newly added) KML interface `KmlIss*` https://support.huawei.com/enterprise/en/doc/EDOC1100283141/eee6b84d/single-node-mode#EN-US_TOPIC_0000001468009088


bootstrap.sh
```diff
@@ -84,6 +84,7 @@ the location where LPF is installed"
 	echo "  --generator=<value>                 - set the generator for CMake (otherwise use CMake's default)"
 	echo "  --show                              - show generation commands instead of running them"
 	echo "  --delete-files                      - delete files in the current directory without asking for confirmation"
+	echo "  --spblas-prefix                     - delete files in the current directory without asking for confirmation"
```
anyzelman: Explantion is copied from delete-files instead of describing what --spblas-prefix does

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -154,19 +156,20 @@ namespace grb {
 				grb::identities::zero, grb::identities::one
 			>,
 			class Minus = operators::subtract< IOType >,
-			class Divide = operators::divide< IOType >
+			class Divide = operators::divide< IOType >,
+			typename RSI, typename NZI, Backend backend
```
anyzelman, line 159: Why would we want to expose RSI, NZI, and backend on the algorithm level? (We never do that right?)

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -310,23 +313,6 @@ namespace grb {
...
-				ret = ret ? ret : grb::dot< descr_dense >( bnorm, temp, b, ring );
-			} else {
-				ret = ret ? ret : grb::dot< descr_dense >( bnorm, b, b, ring );
-			}
-			assert( ret == SUCCESS );
-
-			if( ret == SUCCESS ) {
-				tol *= sqrt( grb::utils::is_complex< IOType >::modulus( bnorm ) );
-			}
-
```
anyzelman: The standard is to pass a relative tolerance to a solver, see e.g. https://docs.octave.org/latest/Specialized-Solvers.html. Two options: either roll this back, or allow for both variants on the ALP/GraphBLAS level (with the relative tolerance as default)

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -415,13 +401,6 @@ namespace grb {
 			// output that is independent of error code
 			iterations = iter;
 
-			// return correct error code
-			if( ret == SUCCESS ) {
-				if( sqrt( residual ) >= tol ) {
-					// did not converge within iterations
-					return FAILED;
-				}
-			}
```
anyzelman: Even if using an absolute norm the algorithm should return FAILED if it does not reach that tolerance

include/graphblas/nonblocking/vector.hpp
```diff
@@ -197,12 +197,44 @@ namespace grb {
 
 		friend class PinnedVector< D, nonblocking >;
```
anyzelman, line 199: Maybe add a comment here that these are native interface friends? (See e.g. nonblocking/matrix.hpp)

include/transition/kml_iss.h
```diff
@@ -0,0 +1,129 @@
...
+
+int KmlIssCgCleanSI(KmlSolverTask **);
+int KmlIssCgCleanDI(KmlSolverTask **);
+
+
+#ifdef __cplusplus
+} // end extern "C"
+#endif
+
+#endif // end ifdef _H_ALP_KML_ISS
```
anyzelman, line 145: Missing empty line below here

include/transition/kml_iss.h
```diff
@@ -0,0 +1,129 @@
+
+
```
anyzelman: One too many empty lines

include/transition/kml_iss.h
```diff
@@ -0,0 +1,129 @@
...
+#include <stddef.h> // for size_t
+
+#include "solver.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
```
anyzelman: Inconsistent use of two empty lines and one empty line for vertical spacing throughout the file, starting here -- please unify

src/transition/CMakeLists.txt
```diff
@@ -18,48 +18,70 @@
...
+	add_transition_library( spsolver_nonblocking_static STATIC "spsolver_nonblocking" ${CMAKE_CURRENT_SOURCE_DIR}/solver.cpp
+		PUBLIC_LINK_LIBRARIES backend_nonblocking
+	)
 
-	install( TARGETS sparseblas_omp_static
-		EXPORT GraphBLASTargets
-		ARCHIVE DESTINATION "${SHMEM_BACKEND_INSTALL_DIR}"
+	add_transition_library( kml_solver STATIC "kml_solver" ${CMAKE_CURRENT_SOURCE_DIR}/kml_iss.cpp
+		PUBLIC_LINK_LIBRARIES spsolver_nonblocking_static
 	)
```
anyzelman: I prefer that we by default only use the nonblocking backend to generate the solver library. Compiling it with OMP of course works but we want people to use the nonblocking backend, no?

Optionally, the emitting the libraries compiled with the blocking OpenMP backend can be enabled on request.

On the naming: the nonblocking solver libraries should just be called shmem or shmem_parallel, and should not be called `nonblocking`. Otherwise it is not immediately clear to users what it is for-- by contrast, users will always understand what is sequential or shmem parallel.
The blocking backend could be called shmem_blocking: since it would only be generated when the user asks for it, that user should know what is blocking and nonblocking, so we can use that nomenclature.

src/transition/CMakeLists.txt
```diff
@@ -18,48 +18,70 @@
...
+	)
 
-	install( TARGETS sparseblas_static
-		EXPORT GraphBLASTargets
-		ARCHIVE DESTINATION "${SHMEM_BACKEND_INSTALL_DIR}"
+	set( _SPBLAS_PREFIX "kml_" )
+	if( SPBLAS_PREFIX )
+		set( _SPBLAS_PREFIX "${SPBLAS_PREFIX}_" )
+	endif()
+	add_transition_library( ${_SPBLAS_PREFIX}sparseblas_static STATIC "${_SPBLAS_PREFIX}sparseblas"
```
anyzelman: Should the name not be something like `sparseblas_sequential` rather than "sparseblas"?

src/transition/kml_iss.cpp
```diff
@@ -0,0 +1,176 @@
...
+			break;
+		}
+	}
+	return result;
+}
+
+#define KML_CG_PREFIXED( name ) KmlIssCg ## name
+
+
+int KML_CG_PREFIXED(InitSI)( KmlSolverTask **handle, int n, const float *a, const int *ja, const int *ia ) {
```
anyzelman: Please check for line overflow (80 chars max, starting from first non-space char)

src/transition/kml_iss.cpp
```diff
@@ -0,0 +1,176 @@
...
+	return sparse_err_t_2_int( err );
+}
+
+int KML_CG_PREFIXED(InitDI)( KmlSolverTask **handle, int n, const double *a, const int *ja, const int *ia ) {
+	sparse_err_t err = sparse_cg_init_dii( handle, n, a, ja, ia );
+	return sparse_err_t_2_int( err );
+}
+
+
+
```
anyzelman: Remove superfluous empty lines (also below)

src/transition/kml_iss.cpp
```diff
@@ -0,0 +1,176 @@
...
+		if( r != NO_ERROR ) {									\
+			err = ( r == NULL_ARGUMENT ) ?						\
+				KMLSS_NULL_ARGUMENT : KMLSS_INTERNAL_ERROR;		\
+		}														\
+		break;													\
+	}															\
+	default:													\
+		err = KMLSS_BAD_SELECTOR;								\
+	}															\
+	return err;
```
anyzelman: The back-ticks are not space-aligned. Also, not better to define all macros somewhere at the top of this file?

src/transition/kml_iss.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+#define KMLSS_SOLVE( nb, sparse_err_t_val )				\
+	if( nb != 1 ) { return KMLSS_BAD_NB; }				\
+	int err = KMLSS_NO_ERROR;							\
+	const sparse_err_t r = sparse_err_t_val;			\
+	if( r != NO_ERROR ) {								\
+		err = ( r == NULL_ARGUMENT ) ?					\
+			KMLSS_NULL_ARGUMENT : KMLSS_INTERNAL_ERROR;	\
+	}													\
+	return err;
```
anyzelman: same here

anyzelman: (and further down this file also)

src/transition/solver.cpp
```diff
@@ -0,0 +1,686 @@
...
+
+#include <graphblas/algorithms/conjugate_gradient.hpp>
+
+
+template< typename T, typename NZI, typename RSI >
+class CG_Data {
+
+	private:
+
+		typedef grb::Matrix< T, grb::config::default_backend, NZI, NZI, RSI > Matrix;
```
anyzelman: line overflow?

tests/smoke/conjugate_gradient.cpp
```diff
@@ -136,6 +136,39 @@ void ioProgram( const struct input &data_in, bool &success ) {
...
+	grb::operators::add< ScalarType >, grb::operators::mul< ScalarType >,
+	grb::identities::zero, grb::identities::one
+>;
+
+template<
+	Backend backend
+> RC computeNorm(
+	const grb::Vector< ScalarType, backend > &v,
+	grb::Vector< ScalarType, backend > &temp,
+	BaseScalarType &norm
```
anyzelman: The code standard has that the output should be the first argument, not the last

tests/smoke/conjugate_gradient.cpp
```diff
@@ -136,6 +136,39 @@ void ioProgram( const struct input &data_in, bool &success ) {
...
+			}, temp
+		);
+		rc = rc ? rc : grb::dot< descriptors::dense >( bnorm, temp, v, ring );
+	} else {
+		rc = rc ? rc : grb::dot< descriptors::dense >( bnorm, v, v, ring );
+	}
+	assert( rc == SUCCESS );
+
+	norm = rc ? ring.template getZero< BaseScalarType >() :
+		sqrt( grb::utils::is_complex< ScalarType >::modulus( bnorm ) );
```
anyzelman: `sqrt`->`std::sqrt` ?

tests/smoke/conjugate_gradient.cpp
```diff
@@ -209,17 +242,24 @@ void grbProgram( const struct input &data_in, struct output &out ) {
 	set( x, static_cast< ScalarType >( 1 ) / static_cast< ScalarType >( n ) );
 	set( b, static_cast< ScalarType >( 1 ) );
```
anyzelman: `set`->`grb::set`?

tests/smoke/conjugate_gradient.cpp
```diff
@@ -298,6 +338,14 @@ void grbProgram( const struct input &data_in, struct output &out ) {
 		return;
 	}
 
+	const double residual_norm = sqrt( out.residual );
```
anyzelman: std::sqrt

tests/smoke/conjugate_gradient.cpp
```diff
@@ -298,6 +338,14 @@ void grbProgram( const struct input &data_in, struct output &out ) {
...
+	if( residual_norm >= tolerance ) {
+		// notify caller that the residual is too high
+		std::cerr << "Residual is too high: " << residual_norm << " >= " << tolerance << "\n";
+		out.error_code = 40;
+	}
```
anyzelman: In this case, the error code returned should also be `FAILED`

anyzelman (changes requested):
This is great, only super minor comments (and it may needs rebasing-- I did see a force push from me yesterday, but unsure if i did that after the latest merge or not)

include/graphblas/algorithms/conjugate_gradient.hpp
anyzelman: Self-updating: there is one extremely compelling (though futuristic) use case-- that of a dispatching backend that given an incoming compute request, decides on which backend/device the computation should run, and then sends it there. Such a dispatcher was already envisioned on the core primitive level, but also make sense (perhaps even makes more sense) on the algorithm level. Made issue #282 to resolve this for all algorithms at a later stage.

TL;DR this comment can be ignored / resolved without any code changes

bootstrap.sh
alberto-scolari: solved in recent commit

include/graphblas/algorithms/conjugate_gradient.hpp
alberto-scolari: modification reverted

include/graphblas/algorithms/conjugate_gradient.hpp
alberto-scolari: modification reverted

include/graphblas/algorithms/conjugate_gradient.hpp
alberto-scolari: modification reverted

tests/smoke/conjugate_gradient.cpp
alberto-scolari: modification reverted

tests/smoke/conjugate_gradient.cpp
alberto-scolari: modification reverted; other occurrences fixed

tests/smoke/conjugate_gradient.cpp
alberto-scolari: modification reverted

tests/smoke/conjugate_gradient.cpp
alberto-scolari: modification reverted

tests/smoke/conjugate_gradient.cpp
alberto-scolari: modification reverted

include/graphblas/nonblocking/vector.hpp
alberto-scolari: fixed

include/transition/kml_iss.h
alberto-scolari: fixed

include/transition/kml_iss.h
alberto-scolari: fixed

include/transition/kml_iss.h
alberto-scolari: fixed

src/transition/CMakeLists.txt
alberto-scolari: fixed naming and disabling reference[_omp]-based targets by default; optionally enabled via CMake or bootstrap script

src/transition/CMakeLists.txt
alberto-scolari: fixed

src/transition/kml_iss.cpp
alberto-scolari: fixed

src/transition/kml_iss.cpp
alberto-scolari: fixed

anyzelman:
Manually running full test suite with LPF, and waiting GitHub and internal CIs after final code review mods of [8bd4707](https://github.com/Algebraic-Programming/ALP/pull/281/commits/8bd4707ed50f803f3cc538f642075d7f30e55056)

anyzelman:
@alberto-scolari I tested the spblas prefix option but it seems to have no effect: the header and library for the KML solver are still called kml_, and the emitted prefix macro is unchanged as well. I'm also wondering about what prefix should be changed-- should the kml solver interface not always remain the same, and spblas should instead change the spblas.h only? (The doc at bootstrap.sh says only the kml prefix would change.)

The blas_sparse.h header I believe should never change prefixes, as that is the only properly standardised api (as by Duff et al.). At most the prefixes of the extended sparse_blas API (currently prefixed EXTBLAS_) could be configurable.

The spblas API is like an industry de-facto type standard where in fact it is common to allow for changed prefixes, so the option could well apply there.

The MKL solver API is non-standard afaik so can be kept without a capability of changing its prefix. Its implementation in relying on solver.h is really nice btw.

In summary,
1. the --spblas-prefix option should modify the function names in spblas.h only, and leave the mkl stuff unchanged (both its file names and function names);
2. the bootstrap.sh doc should be updated such (in that the option changes function names in spblas.h only);
3. (and primarily:) currently the option seems to do nothing and should be fixed (or removed and kept for a separate MR-- maybe easier)

I meanwhile rebased and relaunched all tests, this time also on ARM for a change with recent tests runs:)

anyzelman:
At this point, all manual tests (unit+smoke) on ARM with LPF pass; all CIs also pass.

anyzelman:
Many thanks @alberto-scolari -- it sadly did however reset a lot of my previous review so please allow me to re-review and merge it tomorrow;) I'll mean time rebase and run the test suite.

anyzelman (dismissed):

anyzelman:
Review OK, manual smoke test with LPF OK. Waiting CIs.

anyzelman:
All CIs OK

anyzelman (approved):

PR #286, opened 2024-01-31
279 bug in move assignment of matrices

### LINKED ISSUE #279: Bug in move assignment of matrices
Using move assignment on matrices doesn't do a cleanup of a previous id. This results in id collisions.

byjtew:
Closes #279


tests/unit/id.cpp
```diff
@@ -215,6 +215,17 @@ void grb_program2( const struct input &in, struct output &out ) {
 		return;
 	}
 
+	/*
+	Test for move assignement id cleanup.
+	
+	Creating and performing move assignment on multiple new objects to generate potential id collision.
+	*/
```
anyzelman: The commenting style should be of the form
```
/**
 * line 1
 *
 * line 2
 */
```

tests/unit/id.cpp
```diff
@@ -215,6 +215,17 @@ void grb_program2( const struct input &in, struct output &out ) {
...
+	for( int i = 0; i < 1000; i++ ) {
+		grb::Matrix< int > new_one( 10, 10, 10 );
+		grb::Matrix< int > new_two( 10, 10, 10 );
+		new_two = std::move(new_one);
```
anyzelman: I'm confused on how a clash is detected -- does it on collision trigger some internal assertion? If yes, is something like the below a correct way to make the test more robust?
```
	for( int i = 0; i < 1000; i++ ) {
		grb::Matrix< int > new_one( 10, 10, 10 );
		grb::Matrix< int > new_two( 10, 10, 10 );
                if( grb::getID( new_one ) == grb::getID( new_two ) { /* test failed */ }
                const size_t old_one_id = grb::getID( new_one );
                const size_t old_two_id = grb::getID( new_two );
		new_two = std::move(new_one);
                if( grb::getID( new_two ) == old_two_id ) { /* test fails */ }
                if( grb::getID( new_two ) != old_one_id ) { /* test fails */ }
```
(Note that I don't think it's correct to call getID(new_one) after it has been moved from, that would incur undefined behaviour. I also don't think it is correct to build an `std::set` of previous IDs, because in principle our spec allows re-use of IDs, right?)

anyzelman (changes requested):
Hi @byjtew @aleksamilisavljevic could you have a look at the rfc? The first is easy, the second is a little bit tricky to think about-- is the suggestion ok or am i missing something?

tests/unit/id.cpp
aleksamilisavljevic: Thanks, done

tests/unit/id.cpp
aleksamilisavljevic: I think that the issue here is that when a matrix is created, in the `initialize` function, the pointer `alloc[0]` is forwarded to the `reference_mapper`:

https://github.com/Algebraic-Programming/ALP/blob/dc0396cbbae05aaa4df8d90dafe50c938166bfa9/include/graphblas/reference/matrix.hpp#L1426-L1434

And then, the following assertion of the `reference_mapper` fails:

https://github.com/Algebraic-Programming/ALP/blob/dc0396cbbae05aaa4df8d90dafe50c938166bfa9/include/graphblas/utils/DMapper.hpp#L91-L98

I think that the problem is that, upon move assignment, the `old_two_id` isn't removed from the `reference_mapper`, which causes old unused pointers to remain stored in the map, in turn causing the assertion to fail on one of the subsequent matrix creations.

anyzelman: Yeah I agree but I mean that a unit test should not rely on assertions -- the error should ideally also cause an observable ill effect that the test can check for, such as those in the above suggested code block

aleksamilisavljevic: I'm not certain if the code after the failed assert causes any ill effect that is observable outside of the `DMapper.h` itself. 

The maps will have some outdated information, i.e. two deterministic IDs pointing to the same non-deterministic ID, but I don't think this could cause a problem, as the outdated deterministic ID shouldn't be retrievable from any `getID` call.

anyzelman: > I'm not certain if the code after the failed assert causes any ill effect that is observable outside of the DMapper.h itself.

I agree -- I meant a unit test should not rely on the implementation (here asserts) to detect bugs. I have resolved this one as suggested above: it will do a bunch of moves and checks that those moves are consistent in terms of observable behaviour. Running the unit tests with that change now and if it works, will merge. Many thanks for this one (and apologies on the big delay in getting these MRs done!)

anyzelman:
Concept release notes:

Aleksa uncovered a bug where move-assigning matrices could result in container IDs not explicitly being removed. Since IDs are created on the basis of unique keys and the unique keys are derived from pointers, the following could (and did) happen: on creating a new matrix, a new ID is requested based on the same pointer the retained ID was also created with. This was detected by the internal mechanism keeping track of IDs, and resulted in an exception.

This MR fixes the bug by removing the ID once the container with the original pointer is destroyed via a move. Even if a subsequent allocation then happens to return the same pointer, the internal state is now clean and a new ID is successfully derived. The MR also strengthens the `getID` unit test to have been able to detect this issue.

The MR also includes a hotfix for an issue detected by our internal CI: on defining full debug tracing (`-D_DEBUG`), various `BSP1D` code did not compile-- herewith fixed as well.

Finally, this MR and now again fully related with the issue at hand, improves on `_DEBUG` tracing in `bsp1d/matrix.hpp`, and fixes a similar issue for the `BSP1D` backend: there, prior to this MR, move-assignment would result in never-deleted IDs and a small memory leak. Because of the memory leak, the bug never materialised in the same way it did for the `reference` backend, though nevertheless with this MR now correctly removes the ID as part of a move-assign and prevents the memory leak.

Thanks to Aleksa Milisavljevic @aleksamilisavljevic for detecting and fixing this bug!

anyzelman (dismissed):

anyzelman (approved):

anyzelman:
GitHub CI OK, manual unit tests with LPF and with `_DEBUG` OK-- will merge. Thanks @aleksamilisavljevic for contributing this one!

PR #288, opened 2024-01-31
Issue #261: adding missing dense descriptor and delete unnecessary se…

### LINKED ISSUE #261: PageRank algorithm optimisations
The PageRank implementation does not pass the dense descriptor when it could, and includes the use of the expensive `invert_mask` in its performance-critical loop.

aristeidis-mastoras:
Closes issue #261 .

Closes issue #305 .

* [x] Adding `dense` descriptor
* [x] Spreading the `descriptor` template parameter in the `eWiseApply` and `eWiseLambda` functions 
* [x] work around invert_mask

### DIFF
--- a/include/graphblas/base/blas1.hpp
+++ b/include/graphblas/base/blas1.hpp
@@ -3741,11 +3741,13 @@ namespace grb {
 		typename Func,
 		typename DataType,
 		Backend backend,
-		typename Coords
+		typename Coords,
+		typename... Args
 	>
 	RC eWiseLambda(
 		const Func f,
-		const Vector< DataType, backend, Coords > &x
+		const Vector< DataType, backend, Coords > &x,
+		Args...
 	) {
 #ifndef NDEBUG
 		const bool should_not_call_base_vector_ewiselambda = false;
--- a/include/graphblas/base/blas1.hpp
+++ b/include/graphblas/base/blas1.hpp
@@ -3737,6 +3737,7 @@ namespace grb {
 	 * @see perfSemantics
 	 */
 	template<
+		Descriptor = descriptors::no_operation,
 		typename Func,
 		typename DataType,
 		Backend backend,
--- a/include/graphblas/base/blas1.hpp
+++ b/include/graphblas/base/blas1.hpp
@@ -3741,12 +3741,11 @@ namespace grb {
 		typename Func,
 		typename DataType,
 		Backend backend,
-		typename Coords,
-		typename... Args
+		typename Coords
 	>
 	RC eWiseLambda(
 		const Func f,
-		const Vector< DataType, backend, Coords > & x, Args...
+		const Vector< DataType, backend, Coords > &x
 	) {
 #ifndef NDEBUG
 		const bool should_not_call_base_vector_ewiselambda = false;
--- a/include/graphblas/hyperdags/blas2.hpp
+++ b/include/graphblas/hyperdags/blas2.hpp
@@ -444,6 +444,7 @@ namespace grb {
 
 	/** \internal Uses a direct implementation. */
 	template<
+		Descriptor = descriptors::no_operation,
 		typename Func, typename DataType,
 		typename RIT, typename CIT, typename NIT
 	>
@@ -470,6 +471,7 @@ namespace grb {
 
 		/** \internal This is the end recursion */
 		template<
+			Descriptor descr,
 			typename Func, typename DataType,
 			typename RIT, typename CIT, typename NIT
 		>
@@ -479,7 +481,7 @@ namespace grb {
 			std::vector< uintptr_t > &sources,
 			std::vector< uintptr_t > &destinations
 		) {
-			const RC ret = grb::eWiseLambda( f, internal::getMatrix(A) );
+			const RC ret = grb::eWiseLambda< descr >( f, internal::getMatrix(A) );
 			if( ret != SUCCESS ) { return ret; }
 			if( nrows( A ) == 0 || ncols( A ) == 0 ) { return ret; }
 			std::array< const void *, 0 > sourcesP{};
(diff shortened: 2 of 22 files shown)


include/graphblas/base/blas1.hpp
```diff
@@ -3737,15 +3737,15 @@ namespace grb {
...
 		typename Func,
 		typename DataType,
 		Backend backend,
-		typename Coords,
-		typename... Args
+		typename Coords
 	>
 	RC eWiseLambda(
 		const Func f,
-		const Vector< DataType, backend, Coords > & x, Args...
```
anyzelman, line 3748: Why are the varargs removed here?

include/graphblas/hyperdags/blas2.hpp
```diff
@@ -565,9 +569,13 @@ namespace grb {
 			return vxm< descr >( u, mask, v, A, ring, phase );
 		}
 		const RC ret = vxm< descr >(
-			internal::getVector(u), internal::getVector(mask),
-			internal::getVector(v), internal::getVector(v_mask), internal::getMatrix(A),
-			ring, phase
+			internal::getVector(u),
+			internal::getVector(mask),
```
anyzelman: I think it's clearer if these two are grouped (and similar further down for u and v_mask--- I agree the matrix could/should go on its own row)

include/graphblas/nonblocking/blas2.hpp
```diff
@@ -1518,11 +1519,12 @@ namespace grb {
 		internal::le.execution();
 
 		// second, delegate to the reference backend
-		return eWiseLambda< ActiveDistribution, Func, DataType, RIT, CIT, NIT >(
+		return eWiseLambda< descr, ActiveDistribution, Func, DataType, RIT, CIT, NIT >(
 			f, internal::getRefMatrix( A ), s, P );
 	}
 
 	template<
+		Descriptor descr = descriptors::no_operation,
```
anyzelman, line 1523: This descriptor is not passed to the subsequent call to eWuseLambda (line 1551 below)

include/graphblas/reference/blas1.hpp
```diff
@@ -9759,7 +9760,10 @@ namespace grb {
 	 * @see Vector::operator[]()
 	 * @see Vector::lambda_reference
 	 */
-	template< typename Func, typename DataType, typename Coords >
+	template<
+		Descriptor descr = descriptors::no_operation,
```
anyzelman: The function should return illegal if x was not dense but the dense descriptor was passed (maybe this check is also missing with the other backends)

include/graphblas/reference/blas2.hpp
```diff
@@ -2824,6 +2825,7 @@ namespace grb {
 	 * @see grb::eWiseLambda for the user-level specification.
 	 */
 	template<
+		Descriptor descr = descriptors::no_operation,
```
anyzelman, line 2854: (it seems forgotten here as well -- note that the dense descriptor is only defined to apply for vectors (x), and not for A)

anyzelman (changes requested):
Please see the review comments -- also one question: should allowing a dense descriptor to eWiseLambda not also lead to a change in a cpp file src/nonblocking/ somewhere?

anyzelman:
Note: I also added the todo of removing the invert_mask in the PR algorithm. If no objection I could do that quickly around now?

anyzelman:
> Note: I also added the todo of removing the invert_mask in the PR algorithm. If no objection I could do that quickly around now?

This is done. I did run into a new bug in nonblocking, I opened issue #305 for it.

include/graphblas/base/blas1.hpp
byjtew: Because they are never implemented

anyzelman:
@aristeidis-mastoras @byjtew -- this one is now conflicted but was the next on my list. I'll try resolve the conflicts now

anyzelman:
Rebase complete, smoke tests with LPF OK. Unit tests with LPF and GitHub CI running overnight

anyzelman:
(If anyone of @aristeidis-mastoras @byjtew wants to review how I handled the conflicts lemme know before tomorrow -- if no message by e.g. noon, I'll proceed with review/merge:)

anyzelman:
Crossref #305 which this MR should also fix (TBC)

anyzelman (dismissed):

anyzelman:
Review OK but found one issue in the bsp1d backend. Re-running smoke & unit tests with LPF.

anyzelman:
All tests OK, will merge

anyzelman:
Concept release notes:

Prior to this MR, the PageRank implementation does not pass the dense descriptor when it could, while it includes the use of the expensive `invert_mask` in its performance-critical loop. Mask inversion is expensive since it requires iterating over the full vector range, while, in this case, the inverted mask is constant. Hence, this MR introduces the following changes:
- the inverted mask is precomputed once, while in the performance-critical loop the precomputed inversion is reused as a structural mask; and
- the PR algorithm is reviewed to pass the dense descriptor whenever possible.

This improvement does come at the cost of one additional work-space vector that is now required for the PageRank implementation.

In implementing these fixes, the following feature has been added:
- the `grb::eWiseLambda` now takes a descriptor template argument. This was previously accepted in the nonblocking backend, but was missing from the specification as well as missing in the implementations of other backends.

The following bugs were encountered and fixed with this MR:
- mixed-domain vector-to-scalar reductions were bugged for the nonblocking backend: when the second (underlying) monoid domain type did not match its third (output) domain type, a compiler error would be returned. This MR fixes the behaviour for both left- and right-handed reductions;
- the descriptors for multiple-vector eWiseLambdas were not correctly propagated for the nonblocking backend.

anyzelman (approved):

anyzelman:
Found one more minor issue in nonblocking, now fixed -- re-running unit tests excluding the LPF ones.

PR #290, opened 2024-02-05
654 grb::set matrix to matrix broken in mixed domain

### LINKED ISSUE #241: grb::set matrix-to-matrix broken in mixed-domain
The primitive `grb::set( Matrix<D>&, const Matrix<T> )` contains a bug related to the cast of the values from T to D.

byjtew:
Closes #241 

- Removed redundancy in the `Compressed_Storage::copyFrom(...)` functions. The section in charge of copying the *row_index* and *col_start* arrays has been moved to an external local function called `copyCoordinatesFrom(...)`

- Merged the cast/no-cast `Compressed_Storage::copyFrom(...)` functions using `std::copy_n` rather than a for-loop/memcpy switch

### DIFF
--- a/include/graphblas/base/io.hpp
+++ b/include/graphblas/base/io.hpp
@@ -1279,10 +1279,6 @@ namespace grb {
 	 * is forbidden also. Implementations shall throw a static assertion failure
 	 * if the user nonetheless asks for structural mask inversion.
 	 *
-	 * \warning Mask-modifier descriptors, such as #grb::descriptors::invert_mask
-	 *          are not accepted while #grb::descriptors::structural is in fact
-	 *          implied by this specification.
-	 *
 	 * \parblock
 	 * \par Performance semantics
 	 * Each backend must define performance semantics for this primitive.
--- a/include/graphblas/base/io.hpp
+++ b/include/graphblas/base/io.hpp
@@ -1181,13 +1181,14 @@ namespace grb {
 	 */
 	template<
 		Descriptor descr = descriptors::no_operation,
-		typename OutputType, typename RIT, typename CIT, typename NIT,
-		typename InputType,
+		typename OutputType, typename InputType,
+		typename RIT1, typename CIT1, typename NIT1,
+		typename RIT2, typename CIT2, typename NIT2,
 		Backend backend
 	>
 	RC set(
-		Matrix< OutputType, backend, RIT, CIT, NIT  > &C,
-		const Matrix< InputType, backend, RIT, CIT, NIT > &A,
+		Matrix< OutputType, backend, RIT1, CIT1, NIT1 > &A,
+		const Matrix< InputType, backend, RIT2, CIT2, NIT2 > &C,
 		const Phase &phase = EXECUTE,
 		const typename std::enable_if<
 			!grb::is_object< OutputType >::value &&
@@ -1198,8 +1199,8 @@ namespace grb {
 		const bool should_not_call_base_matrix_set = false;
 		assert( should_not_call_base_matrix_set );
 #endif
-		(void) C;
 		(void) A;
+		(void) C;
 		(void) phase;
 		return UNSUPPORTED;
 	}
@@ -1292,15 +1293,15 @@ namespace grb {
 	 */
 	template<
 		Descriptor descr = descriptors::no_operation,
-		typename DataType, typename RIT, typename CIT, typename NIT,
-		typename MaskType,
-		typename ValueType,
+		typename OutputType, typename InputType, typename ValueType,
+		typename RIT1, typename CIT1, typename NIT1,
+		typename RIT2, typename CIT2, typename NIT2,
 		Backend backend
 	>
 	RC set(
-		Matrix< DataType, backend, RIT, CIT, NIT  > &C,
-		const Matrix< MaskType, backend, RIT, CIT, NIT > &mask,
-		const ValueType& val,
+		Matrix< OutputType, backend, RIT1, CIT1, NIT1 > &C,
+		const Matrix< InputType, backend, RIT2, CIT2, NIT2 > &A,
+		const ValueType &val,
 		const Phase &phase = EXECUTE,
 		const typename std::enable_if<
 			!grb::is_object< DataType >::value &&
@@ -1312,8 +1313,8 @@ namespace grb {
 		const bool should_not_call_base_matrix_masked_set = false;
 		assert( should_not_call_base_matrix_masked_set );
 #endif
+		(void) A;
 		(void) C;
-		(void) mask;
 		(void) val;
 		(void) phase;
 		return UNSUPPORTED;
--- a/include/graphblas/base/io.hpp
+++ b/include/graphblas/base/io.hpp
@@ -1223,10 +1223,10 @@ namespace grb {
 	 * mutually exclusive for this primitive.
 	 * \endparblock
 	 *
-	 * @tparam DataType  The type of each element in the given matrix.
-	 * @tparam MaskType  The type of each element in the given mask.
-	 * @tparam ValueType The type of the given value. Should be convertible
-	 *                   to \a DataType.
+	 * @tparam OutputType The type of each element in the given matrix.
+	 * @tparam MaskType   The type of each element in the given mask.
+	 * @tparam ValueType  The type of the given value. Should be convertible
+	 *                    to \a OutputType.
 	 * @tparam RIT       The integer type for encoding row indices.
 	 * @tparam CIT       The integer type for encoding column indices.
 	 * @tparam NIT       The integer type for encoding nonzero indices.
@@ -1265,7 +1265,7 @@ namespace grb {
 	 *
 	 * When \a descr includes #grb::descriptors::no_casting then code shall not
 	 * compile if one of the following conditions are met:
-	 *  -# \a ValueType does not match \a DataType; or
+	 *  -# \a ValueType does not match \a OutputType; or
 	 *  -# \a MaskType does not match <tt>bool</tt>.
 	 *
 	 * In these cases, the code shall not compile: implementations must throw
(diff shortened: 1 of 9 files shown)


include/graphblas/base/io.hpp
```diff
@@ -1078,6 +1078,120 @@ namespace grb {
...
+		typename RIT2, typename CIT2, typename NIT2,
+		Backend backend
+	>
+	RC set(
+		Matrix< OutputType, backend, RIT1, CIT1, NIT1 > &A,
+		const Matrix< InputType, backend, RIT2, CIT2, NIT2 > &C,
+		const Phase &phase = EXECUTE
+	) noexcept {
+#ifndef NDEBUG
+		const bool should_not_call_base_matrix_set_copy_masked = false;
```
anyzelman: `should_not_call_base_matrix_set_copy_masked` -> `should_not_call_base_matrix_set_copy` ?

include/graphblas/reference/compressed_storage.hpp
```diff
@@ -502,21 +580,17 @@ namespace grb {
...
 				/**
-				 * Copies contents from a given Compressed_Storage.
+				 * Copies coordinates from a given Compressed_Storage, then fills the
+				 * values with the given identity.
 				 *
 				 * Performs no safety checking. Performs no (re-)allocations.
 				 *
-				 * @tparam use_id If set to <tt>true</tt>, use \a id instead of values in
-				 *                \a other.
-				 *
```
anyzelman: Incorrect deletion?

include/graphblas/reference/compressed_storage.hpp
```diff
@@ -502,21 +580,17 @@ namespace grb {
...
 				 * @param[in] other The container to copy from.
 				 * @param[in] nz    The number of nonzeroes in the \a other container.
 				 * @param[in] m     The index dimension of the \a other container.
 				 * @param[in] start The start position to copy from (inclusive).
 				 * @param[in] end   The end position to copy to (exclusive).
 				 * @param[in] id    A pointer to a value overriding those in \a other.
-				 *                  Will only be used if and only if \a use_id is set
-				 *                  <tt>true</tt>.
```
anyzelman: incorrect deletion?

include/graphblas/reference/compressed_storage.hpp
```diff
@@ -660,26 +692,12 @@ namespace grb {
...
-							assert( col_start[ chk ] <= col_start[ chk + 1 ] );
-						}
-#endif
-					}
+
+					copyCoordinatesFrom( *this, other, nz, m, k, end );
 				}
 
+
+
```
anyzelman: Remove superfluous empty lines?

include/graphblas/reference/io.hpp
```diff
@@ -969,27 +969,50 @@ namespace grb {
...
-			const InputType2 * __restrict__ id = nullptr
+			const Matrix< InputType, reference, RIT2, CIT2, NIT2 > &A,
+			const ValueType * __restrict__ id = nullptr
 		) noexcept {
 #ifdef _DEBUG
 			std::cout << "Called grb::set (matrices, reference), execute phase\n";
 #endif
+			if( A_is_mask ) {
+				assert( id != nullptr );
+			}
```
anyzelman: Maybe put this block in an `#ifndef NDEBUG`  ?

anyzelman: Also I tend to order dynamic checks (including asserts) until after the static checks -- in reality it doesn't matter much, but as a kind of style standard.

tests/unit/copyMixedDomainsMatrices.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include <vector>
+#include <numeric>
+#include <algorithm>
+
+#include <graphblas.hpp>
+
+
+using namespace grb;
+
+
```
anyzelman, line 27: superfluous empty line?

tests/unit/copyMixedDomainsMatrices.cpp
```diff
@@ -0,0 +1,211 @@
...
+		) {
+			std::cerr << "\t initialisation of A_int FAILED: rc is "
+				<< grb::toString( rc ) << "\n";
+			rc = FAILED;
+			return;
+		}
+	}
+
+
+
```
anyzelman: also here

tests/unit/copyMixedDomainsMatrices.cpp
```diff
@@ -0,0 +1,211 @@
...
+	size_t in = 100;
+
+	// error checking
+	if( argc > 2 ) {
+		printUsage = true;
+	}
+	if( argc == 2 ) {
+		size_t read;
+		std::istringstream ss( argv[ 1 ] );
+		if( ! ( ss >> read ) ) {
```
anyzelman: no space after `!`

include/graphblas/reference/io.hpp
```diff
@@ -969,27 +969,50 @@ namespace grb {
...
+			NO_CAST_ASSERT(
+				( !( descr & descriptors::no_casting ) ||
+				( std::is_same< NIT1, NIT2 >::value ) ),
+				"internal::grb::set", "called with non-matching value index types"
+			);
 			NO_CAST_ASSERT( ( !( descr & descriptors::no_casting ) ||
-				( A_is_mask && std::is_same< InputType2, OutputType >::value ) ),
+				( A_is_mask && std::is_same< ValueType, OutputType >::value ) ),
 				"internal::grb::set", "Called with non-matching value types"
 			);
```
anyzelman: Add a static assert that checks that ValueType can be cast into OutputType (if A_is_mask is false)?

include/graphblas/reference/compressed_storage.hpp
```diff
@@ -1176,45 +1194,22 @@ namespace grb {
...
+					typename UnusedType = std::nullptr_t
 				>
 				void copyFrom(
-					const Compressed_Storage< InputType, IND, SIZE > &other,
+					const Compressed_Storage< InputType, InputIND, InputSIZE > &other,
 					const size_t nz, const size_t m, const size_t start, size_t end,
 					const UnusedType * __restrict__ = nullptr
 				) {
 					// the use_id template is meaningless in the case of pattern matrices, but
 					// is retained to keep the API the same as with the non-pattern case.
```
anyzelman: add `(void) unusedValue;` (to prevent static code analysis warnings) and update the flagged comment with use_id->unusedValue

include/graphblas/reference/compressed_storage.hpp
```diff
@@ -618,39 +678,11 @@ namespace grb {
...
-#endif
-						assert( k <= loop_end );
-						(void) std::memcpy(
-							row_index + k,
-							other.row_index + k,
-							(loop_end - k) * sizeof( IND )
-						);
+
+						GRB_UTIL_IGNORE_CLASS_MEMACCESS;
+						std::copy_n( other.values + k, loop_end - k, values + k );
```
anyzelman: Two issues:
- should go into an OMP guard and set start, end initially also based on the thread ID?
- I prefer an explicit for-loop instead of `copy_n` because of the cross-domain casting that spawned the issue (better to have it explicit)

include/graphblas/reference/compressed_storage.hpp
```diff
@@ -30,6 +30,84 @@ namespace grb {
...
+						input.col_start + k,
+						loop_end - k,
+						output.col_start + k
+					);
+	#ifndef NDEBUG
+					for( size_t chk = k; chk < loop_end - 1; ++chk ) {
+						assert( input.col_start[ chk ] <= input.col_start[ chk + 1 ] );
+						assert( output.col_start[ chk ] <= output.col_start[ chk + 1 ] );
+					}
+	#endif
```
anyzelman: This block should execute in parallel (for the `reference_omp` backend)-- if intercepting at the indicated block, the usual OMP function may be used to cut the range k to loop_end statically.

If I read the static asserts in this file correctly, `copy_n` may also here perform casing, in which case it may be nicer to either disallow it (is it ever necessary to copy between mis-matching coordinate classes? I remember seeing static asserts that explicitly disallowed it elsewhere in this MR), or, alternatively, to write it explicitly in a for-loop.

Note that for parallelisation it doesn't matter -- if we have the thread-local start and end coordinates, each thread could in principle call `copy_n` (or do a for-loop).

I'm fine with either -- I think retaining the possibility of casting is more future-proof, so a slight bias towards that. Copy_n I would only retain if disallowing casting.

anyzelman (changes requested):

include/graphblas/base/io.hpp
byjtew:
```suggestion
		const bool should_not_call_base_matrix_set_copy = false;
```

include/graphblas/reference/compressed_storage.hpp
byjtew: Well since this whole logic is called within an already opened OMP region I am not sure it's a good idea. But I can switch to an `simd` for-loop

include/graphblas/reference/compressed_storage.hpp
byjtew: Same comment as above:
> Well since this whole logic is called within an already opened OMP region I am not sure it's a good idea. But I can switch to an `simd` for-loop

anyzelman (dismissed):

anyzelman:
Here is a question though: currently the `grb::set( matrix, mask, value );` does not take mask descriptors right?

I think there's good reason for not supporting `invert_mask`, because that implies a quadratic workload (in the size of the matrices), and seems only useful for cases where the mask matrix that is to be inverted has a number of nonzeroes that approaches the same quadratic bound-- i.e., it turns more into a dense computation that is typically best avoided.

However, we could support non-structural masks that inspect the mask-matrix' actual value, and if it compares false, not copy its coordinate into the output matrix. Since we copy row-by-row this should not be hard to implement. The current semantics would then be activated once more when passing the `descriptors::structural' in there. Would that be hard to add in @byjtew ?

anyzelman:
I do note that in disagreement to what I wrote above, that the GraphBLAS C API does support mask inversion even in the matrix case.

anyzelman:
Actually now also partially disagreeing with myself: it is still useful to invert a non-structural mask. We could forbid (static_assert) inverting a structural mask specifically maybe?

anyzelman:
This one is ready to merge (after conflict resolution)

anyzelman:
Kind of feel bad for not merging this earlier - a bunch of conflicts emerged meanwhile, now (hopefully) fixed. Also did a cursory code review and fixes due to the rebase. Running unit tests now to make sure nothing broke, then will continue review. I saw at least one TODO:

- [x] while I understand the intent of the `omp simd` pragmas, I don't believe they actually trigger at the moment. Can be re-written using the standard way for "auto"-vectorising applied everywhere else in ALP

anyzelman (approved):

anyzelman:
Code review OK, GitHub CI OK. Waiting internal CI OK and manual unit test run with LPF OK before merge.

anyzelman:
Concept release notes:

This MR fixes the behaviour of `grb::set( matrix, * )` by 1) correctly interpreting output masks and 2) fixes the behaviour of casting values from the input matrix value type to that of the output matrix.

This MR also
- adds a unit test for both fixes to guard against regressions;
- fixes / updates the base specification of `grb::set( matrix, * )`;
- allows for `grb::set( matrix, matrix[, scalar] )` to work for two matrices with different index / offset types;
- improves debug tracing for the reference `Compressed_Storage` class;
- includes minor code style fixes.

Thanks to Benjamin Lozes for initially identifying the issue and providing the fix!

anyzelman:
Internal CI & internal unit tests (with LPF) OK; will merge

PR #299, opened 2024-02-08
Issue #293: fixing generation of installation infrastructure

### LINKED ISSUE #293: Wrong include paths for CMake installation
When issuing `make install`, CMake creates several CMake configuration files for CMake-based projects to easily import ALP, as from the [docs](https://github.com/Algebraic-Programming/ALP/blob/develop/docs/Use_ALPGraphBLAS_in_your_own_project.md#via-cmake).

Some include paths (for transition libraries) are wrong in the installation, due to wrong settings in `include/CMakeLists.txt`. They should easily be fixed.

alberto-scolari:
Closes #293

### DIFF
--- a/include/CMakeLists.txt
+++ b/include/CMakeLists.txt
@@ -100,7 +100,12 @@ install( TARGETS backend_headers_nodefs EXPORT GraphBLASTargets
 	INCLUDES DESTINATION "${INCLUDE_INSTALL_DIR}"
 )
 
-install( TARGETS alp_utils_headers EXPORT GraphBLASTargets )
+# the utils library currently depends heavily on other ALP functionalities,
+# but could be easily split out in future; in view of this, export an
+# independent include path (incidentally the same as ALP)
+install( TARGETS alp_utils_headers EXPORT GraphBLASTargets
+	INCLUDES DESTINATION "${INCLUDE_INSTALL_DIR}"
+)
 
 if( WITH_REFERENCE_BACKEND_HEADERS )
 	add_library( backend_reference_headers INTERFACE )
@@ -244,3 +249,4 @@ install( DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/transition/"
 install( TARGETS transition_headers EXPORT GraphBLASTargets
 	INCLUDES DESTINATION "${TRANSITION_INCLUDE_INSTALL_DIR}"
 )
+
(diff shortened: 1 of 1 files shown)


anyzelman:
"Wrong" means there is a bug right? I added the bug label.

Also likewise perhaps add @byjtew as additional (cmake;) reviewer?

include/CMakeLists.txt
```diff
@@ -228,21 +225,22 @@ configure_file( ${CMAKE_CURRENT_SOURCE_DIR}/transition/spblas.h.in
...
-	DESTINATION "${INCLUDE_INSTALL_DIR}/transition"
+	DESTINATION "${TRANSITION_INCLUDE_INSTALL_DIR}"
 	FILES_MATCHING REGEX "${HEADERS_REGEX}"
 )
 
-install( TARGETS transition_headers EXPORT GraphBLASTargets )
-
+install( TARGETS transition_headers EXPORT GraphBLASTargets
+	INCLUDES DESTINATION "${TRANSITION_INCLUDE_INSTALL_DIR}"
+)
```
anyzelman, line 251: Missing empty line at end of file

include/CMakeLists.txt
```diff
@@ -101,9 +100,7 @@ install( TARGETS backend_headers_nodefs EXPORT GraphBLASTargets
 	INCLUDES DESTINATION "${INCLUDE_INSTALL_DIR}"
 )
 
-install( TARGETS alp_utils_headers EXPORT GraphBLASTargets
-	INCLUDES DESTINATION "${INCLUDE_INSTALL_DIR}"
-)
+install( TARGETS alp_utils_headers EXPORT GraphBLASTargets )
```
anyzelman: Why's `INCLUDES DESTINATION` removed here but added for other targets later on (e.g., transition_headers?)

anyzelman (changes requested):

include/CMakeLists.txt
alberto-scolari: fixed in latest commit

include/CMakeLists.txt
alberto-scolari: This has changed in the last commit, let me explain the motivation behind both changes.

This line was removed because it is redundant: the utils library has the same include paths as ALP, already installed at line 100 of the same file.

The second commit adds it again because the utils library is ideally a separate library, so this way we have independently installed targets: a user could ideally link utils without linking ALP, or vice versa. A comment explains this.

anyzelman: Got it, many thanks!

anyzelman:
I'm ok with this one if @byjtew is=)

byjtew (approved):

anyzelman (approved):

anyzelman:
Review OK, CI OK, all smoke tests OK, will merge

PR #301, opened 2024-02-09
283 provide the implementation of the masked outer product

### LINKED ISSUE #283: Provide the implementation of the maskedOuter(Matrix result, Matrix mask, Vector u, Vector v)
Provide the implementation of the outer product of vectors u and v, masked to non-zero values of the matrix mask.

aleksamilisavljevic:
Closes #283 

Depends on #317


include/graphblas/reference/blas3.hpp
```diff
@@ -916,6 +916,206 @@ namespace grb {
...
+		for( size_t i = 0; i < nrows; ++i ) {
+			if( internal::getCoordinates( u ).assigned( i ) ) {
+				for( auto k = mask_raw.col_start[ i ]; k < mask_raw.col_start[ i + 1 ]; ++k ) {
+					const size_t k_col = mask_raw.row_index[ k ];
+					if( internal::getCoordinates( v ).assigned( k_col ) ) {
+						nzc++;
+					}
+				}
+			}
+		}
```
byjtew, line 2430: If the matrix contains many sparse rows, iterating over the assigned coordinates of `u` would be more efficient. 

```suggestion
		for( size_t i : internal::getCoordinates( u ) ) {
			for( auto k = mask_raw.col_start[ i ]; k < mask_raw.col_start[ i + 1 ]; ++k ) {
				const size_t k_col = mask_raw.row_index[ k ];
				if( internal::getCoordinates( v ).assigned( k_col ) ) {
					nzc++;
				}
			}
		}
```

On the other hand, you can no longer use openMP on the outer foreach-loop. This leads to the second suggestion:

```suggestion
#ifdef _H_GRB_REFERENCE_OMP_BLAS3
		#pragma omp parallel for reduction(+:nzc)
#endif
		for( size_t i = 0; i < nrows; ++i ) {
			if( internal::getCoordinates( u ).assigned( i ) ) {
				for( auto k = mask_raw.col_start[ i ]; k < mask_raw.col_start[ i + 1 ]; ++k ) {
					const size_t k_col = mask_raw.row_index[ k ];
					if( internal::getCoordinates( v ).assigned( k_col ) ) {
						nzc++;
					}
				}
			}
		}
```

include/graphblas/reference/blas3.hpp
```diff
@@ -916,6 +916,206 @@ namespace grb {
...
+				return PANIC;
+			} else {
+				return FAILED;
+			}
+		}
+
+		grb::Monoid<
+			grb::operators::left_assign< OutputType >,
+			grb::identities::zero
+		> monoid;
```
byjtew: Shouldn't it be a template parameter instead ?

include/graphblas/reference/blas3.hpp
```diff
@@ -916,6 +916,206 @@ namespace grb {
...
+			}
+			CRS_raw.col_start[ i + 1 ] = nzc;
+		}
+
+
+		A_col_index[ 0 ] = 0;
+		for( size_t j = 1; j < ncols; ++j ) {
+			CCS_raw.col_start[ j + 1 ] += CCS_raw.col_start[ j ];
+			A_col_index[ j ] = 0;
+		}
```
byjtew: Split this loop in two and assign `A_col_index` in a separate `#pragma omp parallel for simd`

include/graphblas/reference/blas3.hpp
```diff
@@ -916,6 +916,206 @@ namespace grb {
...
+
+
+		const size_t old_nzc = nzc;
+
+		// use previously computed CCS offset array to update CCS during the
+		// computational phase
+		nzc = 0;
+		for( size_t i = 0; i < nrows; ++i ) {
+			coors.clear();
+			if( internal::getCoordinates( u ).assigned( i ) ) {
```
byjtew:
```suggestion
			if( internal::getCoordinates( u ).assigned( i ) ) {
				coors.clear();
```

include/graphblas/reference/blas3.hpp
```diff
@@ -916,6 +916,206 @@ namespace grb {
...
+			assert( CCS_raw.col_start[ j + 1 ] - CCS_raw.col_start[ j ] ==
+				A_col_index[ j ] );
+		}
+		assert( nzc == old_nzc );
+		
+
+		internal::setCurrentNonzeroes( A, nzc );
+		
+		return ret;
+	}
```
byjtew: Comments on the whole function:
- Add openMP support
- Wrap any CCS logic into `if( !( descr & descriptors::force_row_major ) )`

tests/unit/CMakeLists.txt
```diff
@@ -246,6 +246,10 @@ add_grb_executables( outer outer.cpp
 	BACKENDS reference reference_omp hyperdags nonblocking
 )
 
+add_grb_executables( maskedOuter maskedOuter.cpp
+	BACKENDS reference
```
byjtew: I can do the implementation for the other backends than `reference_omp` if you want, that's pretty quick.

tests/unit/maskedOuter.cpp
```diff
@@ -0,0 +1,226 @@
...
+	> ring;
+
+	// initialise vec
+	const double * vec_iter = &(vec1_vals[ 0 ]);
+	grb::RC rc = grb::buildVector( u, vec_iter, vec_iter + 3, SEQUENTIAL );
+	if( rc != SUCCESS ) {
+		std::cerr << "\t initial buildVector FAILED\n";
+		error = 5;
+	}
+
```
byjtew: We recently added the `std::initializer_list<>` constructor for `grb::Vector` if you want 😄 

```suggestion
	grb::Vector< double > u = { 1, 2, 3 };
```

include/graphblas/reference/blas3.hpp
```diff
@@ -916,6 +916,206 @@ namespace grb {
...
+
+		assert( phase != TRY );
+		if( nrows != grb::nrows( A ) || nrows != m ) {
+			return MISMATCH;
+		}
+
+		if( ncols != grb::ncols( A ) || ncols != n ) {
+			return MISMATCH;
+		}
+
```
byjtew, line 2395: Another clause guard for *empty mask* or *empty u* or *empty v* would be nice

byjtew (changes requested):
Nice implementation, the openMP implementation is pretty quick and easy. 
Good job 👍 

Also, it would be nice to have a (maybe separate) unit test with a user-provided size of the matrix instead of a single predefined one.

include/graphblas/reference/blas3.hpp
aleksamilisavljevic: Shouldn't the matrix just be empty in that case?

tests/unit/maskedOuter.cpp
aleksamilisavljevic: Done, thanks!

include/graphblas/reference/blas3.hpp
byjtew: Yes it should be cleared

tests/unit/CMakeLists.txt
aleksamilisavljevic: Much obliged if you could complete the implementation. The dependencies necessary for the bsp1d seem to break something on my machine.

include/graphblas/reference/blas3.hpp
aleksamilisavljevic: Done, thanks!

include/graphblas/reference/blas3.hpp
aleksamilisavljevic: Done, thanks!

include/graphblas/reference/blas3.hpp
aleksamilisavljevic: Done, thanks!

tests/unit/CMakeLists.txt
byjtew: Sure, I'll push it tomorrow

include/graphblas/reference/blas3.hpp
aleksamilisavljevic: Thanks, I opted for the second suggestion here

include/graphblas/reference/blas3.hpp
aleksamilisavljevic: This monoid can be completely removed, thanks.

include/graphblas/reference/blas3.hpp
byjtew: No wait there is a catch, if you read extremely attentively the doc: "empty mask" means "no mask". Hence, you should rely on the non-masked variant.

aleksamilisavljevic: Wait, I interpreted "empty mask" as it has nnz = 0? Is this the correct interpretation?

byjtew: Yes, "empty mask" means `nnz(mask)==0`

aleksamilisavljevic: But then, a mask with all non-zeros would return the same result as the mask with no non-zeros, right?

byjtew:
Yes, that's not intuitive I agree.
-------- Original Message --------

include/graphblas/reference/blas3.hpp
aleksamilisavljevic: Ok, did that, thanks for this case!

byjtew: @anyzelman Can you confirm ?

anyzelman: Sorry, saw this late -- empty mask means `size(mask)==0`

byjtew: @aleksamilisavljevic Ok sorry for the confusion it's not `nnz(mask)==0` but `size(mask)==0`

tests/unit/CMakeLists.txt
byjtew: (the tomorrow of today)

include/graphblas/reference/blas3.hpp
aleksamilisavljevic: Resolved, thanks.

byjtew:
By the way, why do you call the function `maskedOuter` and not simply `outer`, since the arguments are different it wouldn't collide with the non-masked implementation

byjtew:
I also realised that you were only checking against the structure of the mask (coordinates), but never its values (which is tricky when allowing void & non-void)

tests/unit/maskedOuter.cpp
```diff
@@ -100,8 +100,8 @@ void grbProgram( const void *, const size_t in_size, int &error ) {
 	}
 
 	if( !error ) {
-		rc = grb::maskedOuter< descriptors::force_row_major >( Result2, Mask2, u, v, ring.getMultiplicativeOperator(), RESIZE );
-		rc = rc ? rc : grb::maskedOuter< descriptors::force_row_major >( Result2, Mask2, u, v, ring.getMultiplicativeOperator() );
+		rc = grb::outer< descriptors::force_row_major >( Result2, Mask2, u, v, ring.getMultiplicativeOperator(), RESIZE );
+		rc = rc ? rc : grb::outer< descriptors::force_row_major >( Result2, Mask2, u, v, ring.getMultiplicativeOperator() );
 		if( rc != grb::SUCCESS ) {
 			std::cerr << "Unexpected return code from grb::maskedOuter: "
```
byjtew:
```suggestion
			std::cerr << "Unexpected return code from grb::outer: "
```

tests/unit/maskedOuter.cpp
```diff
@@ -253,8 +253,8 @@ void grb_program_custom_size( const size_t &n, int &error ) {
 	}
 
 	if( !error ) {
-		rc = grb::maskedOuter( Result, Mask, u, v, ring.getMultiplicativeOperator(), RESIZE );
-		rc = rc ? rc : grb::maskedOuter( Result, Mask, u, v, ring.getMultiplicativeOperator() );
+		rc = grb::outer( Result, Mask, u, v, ring.getMultiplicativeOperator(), RESIZE );
+		rc = rc ? rc : grb::outer( Result, Mask, u, v, ring.getMultiplicativeOperator() );
 		if( rc != grb::SUCCESS ) {
 			std::cerr << "Unexpected return code from grb::maskedOuter: "
```
byjtew:
```suggestion
			std::cerr << "Unexpected return code from grb::outer: "
```

byjtew (changes requested):

aleksamilisavljevic:
> I also realised that you were only checking against the structure of the mask (coordinates), but never its values (which is tricky when allowing void & non-void)

Yes, I was actually implementing the structural mask. @byjtew, if the structural descriptor isn't set, is it intended that values evaluate to true iff `mask_raw.getValue(...)` returns something that evaluates to true?

byjtew:
> > I also realised that you were only checking against the structure of the mask (coordinates), but never its values (which is tricky when allowing void & non-void)
> 
> Yes, I was actually implementing the structural mask. @byjtew, if the structural descriptor isn't set, is it intended that values evaluate to true iff `mask_raw.getValue(...)` returns something that evaluates to true?

Not that simple, you should use `utils::interpretMask<descr, T>` instead, it will interpret the mask value to true or false and handle the void case for you.

tests/unit/maskedOuter.cpp
aleksamilisavljevic: Done, thanks

aleksamilisavljevic:
> > > I also realised that you were only checking against the structure of the mask (coordinates), but never its values (which is tricky when allowing void & non-void)
> > 
> > 
> > Yes, I was actually implementing the structural mask. @byjtew, if the structural descriptor isn't set, is it intended that values evaluate to true iff `mask_raw.getValue(...)` returns something that evaluates to true?
> 
> Not that simple, you should use `utils::interpretMask<descr, T>` instead, it will interpret the mask value to true or false and handle the void case for you.

The descriptors should now be properly handled.

include/graphblas/base/blas3.hpp
```diff
@@ -442,6 +442,100 @@ namespace grb {
...
 
+	/**
+	 * Masked outer product of two vectors. Assuming vectors \a u and \a v are
+	 * oriented column-wise, the result matrix \a A will contain \f$ uv^T \f$,
+<<<<<<< HEAD
+	 * masked to non-zero values from \a mask. This is an out-of-place
+=======
+	 * masked to non-zero values from \a mask.. This is an out-of-place
+>>>>>>> 8b4813514f3b87c9a0fe12d0808fd7a6bef80618
+	 * function and will be updated soon to be in-place instead.
```
byjtew: ????

aleksamilisavljevic: Many thanks, don't know how I missed that.

byjtew:
Feel free to mark the MR as "ready" if the tests pass.

aleksamilisavljevic:
> Feel free to mark the MR as "ready" if the tests pass.

Ok, but now I'm wondering if the structural_complement descriptor should be supported at all. I guess that the similar case could be made for forbidding it as in #242

byjtew:
Running the test in the internal GitLab
@aleksamilisavljevic 

---

**FAILING TESTS** for all backends
* `tests/unit/maskedOuter_ndebug_<backend>`
* `tests/unit/maskedOuter_debug_<backend>`

anyzelman:
> > Feel free to mark the MR as "ready" if the tests pass.
> 
> Ok, but now I'm wondering if the structural_complement descriptor should be supported at all. I guess that the similar case could be made for forbidding it as in #242

I would agree-- structurally inverting any sparse matrix seems never a good idea

aleksamilisavljevic:
> Running the test in the internal GitLab @aleksamilisavljevic
> 
> **FAILING TESTS** for all backends
> 
>     * `tests/unit/maskedOuter_ndebug_<backend>`
> 
>     * `tests/unit/maskedOuter_debug_<backend>`

There was a bug in the way the test status was reported. It should be fine now, but I don't have access to the internal GitLab to check it.

Thanks to @aristeidis-mastoras for helping with the debug.

anyzelman:
This actually seems ready, so I provisionally slot it for v0.8

GiovaGa:
Currently `maskedOuter` unit test is failing. I will look into it.

I noticed that this should be merged only after #317 !
Changes in `include/graphblas/utils.hpp` are due to the implementation of some features from there. Some manual context fixes may be needed

PR #309, opened 2024-02-19
Add preconditioning to the ALP CG solvers

### LINKED ISSUE #308: Add preconditioned CG...
...to all of:
1. include/algorithms/conjugate_gradient.hpp;
2. include/transition/solver.h; and
3. include/transition/kml_iss.h.

anyzelman:
Initial jab at adding standard preconditioning to the ALP CG. The following TODOs remain:

- [x] make sure this passes the regular tests for non-preconditioned CG;
- [x] add a smoke test for the preconditioned CG (e.g., simple Jacobi preconditioning);
- [x] expose the mechanism to the solver and ksolver transition paths.

(Note that the latter point corresponds to the actual feature request)

Bugfixes in this MR:

- the vector-to-vector foldr was not implemented in the BSP1D and hybrid backends, herewith fixed.
- the eWiseLambda for BSP1D and hybrid was using local indexing, while it should have been global.
- this also fixes a kind of meta-bug in the distributed implementation of grb::set(BSP1D,hybrid) that erroneously relied on local indexing

Tag @alberto-scolari


anyzelman:
Optional feature: split preconditioners for use with ILU, I think @djelovina's use case was for that particular one, or does it make use of the standard preconditioning now provided (modulo testing)?

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -78,12 +85,26 @@ namespace grb {
...
 		 *
+		 * The preconditioner action \f$ M^{-1}r \f$ is supplied as a standard
+		 * <tt>std::function</tt> that outputs a #grb::RC and takes two arguments:
+		 *  -# the output vector of the preconditioning action as a reference to a
+		 *     #grb::Vector with element type <tt>IOType</tt>, and
+		 *  -# the input vector to the preconditioner as a const-reference to a
+		 *     #grb::Vector with the same element type.
+		 * Both vectors are guaranteed to be dense when given to \a M. A good
+		 * preconditioner action \a Minv is both efficient to apply as well as
+		 * reducing the number of CG iterations required drastically.
```
djelovina: ... i.e. condition number of matrix M^{-1}A is much smaller than the condition number of A.

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -323,110 +375,211 @@ namespace grb {
...
 
 				// temp = A * u;
 				ret = ret ? ret : grb::mxv< descr_dense >( temp, A, u, ring );
-				assert( ret == SUCCESS );
+				assert( ret == grb::SUCCESS );
 
-				// beta = u' * temp
+				// beta = temp' * u
 				beta = zero;
 				if( grb::utils::is_complex< IOType >::value ) {
```
djelovina: You could avoid `is_complex` check by passing   _conjugate_right_mul_ operator to `dot()` call, i.e. take a look  at [this](https://github.com/Algebraic-Programming/ALP/blob/68b6256e2ac0672d648980487399c9d18693fe59/include/graphblas/algorithms/gmres.hpp#L556C1-L561C9) section.

anyzelman: Very elegant indeed -- I applied it

include/graphblas/algorithms/conjugate_gradient.hpp
anyzelman: Thanks! Applied also

anyzelman:
Current status: all passes, except preconditioned CG with bsp1d and hybrid backends due to a segfault. Given it only appears for the preconditioned case, the error should be rather isolated (perhaps it's in the build of the diagonal vector?)

anyzelman:
> Current status: all passes, except preconditioned CG with bsp1d and hybrid backends due to a segfault. Given it only appears for the preconditioned case, the error should be rather isolated (perhaps it's in the build of the diagonal vector?)

I suspect it's a global/local index translation issue -- the eWiseLambda on matrices gives back global indices (at least that's what's specified), while the bracket-operator may expect local indices (which would be an error). If this is indeed the case, then a second bugfix that would be piggy-backing on this MR is that eWiseLambda on vectors then apparently does not use global indices (for bsp1d and hybrid).

anyzelman:
Retarget to v0.8 due to the bugfixes this MR contains

anyzelman:
Running unit and smoke tests, then will move to the last TODO

anyzelman:
All tests OK on x86 and ARM at this point

anyzelman:
This one seems ready, I'll try to test the KML entry point manually (and will run automated tests simultaneously)

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -215,218 +249,330 @@ namespace grb {
 			const IOType zero = ring.template getZero< IOType >();
 			const size_t n = grb::ncols( A );
 
+			// retrieve conditional buffers
+			grb::Vector< IOType > &z = preconditioned ? temp_precond : r;
```
alberto-scolari: I'm not sure why we need the `backend` template param, but if we use it for the input vector we should also use it for the type of `z`, otherwise we risk an assignment between incompatible types; if, instead, we are sure that `backend` is always the default type, then we don't need the type param.

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -215,218 +249,330 @@ namespace grb {
...
+				"Require floating-point residual type."
+			);
+
+			// create a dummy preconditioner and buffer that will never be used
+			std::function<
+				grb::RC( grb::Vector< IOType >&, const grb::Vector< IOType >& )
+			> dummy_preconditioner =
+				[](grb::Vector< IOType >&, const grb::Vector< IOType >&) -> grb::RC {
+					return grb::FAILED;
+				};
```
alberto-scolari: why defining a callback that will never be called? `std::function` has a [default constructor](https://en.cppreference.com/w/cpp/utility/functional/function/function)

anyzelman:
Manual test of entry point also passes. Waiting review/approve from @alberto-scolari before merge~

alberto-scolari (dismissed):

include/graphblas/algorithms/conjugate_gradient.hpp
anyzelman: Oh sorry, only now saw this. Yes this is a good point. I remember me suggesting to indeed remove the `backend` parameter, but then reconsidered. I think the reason was the possibility for dispatching on an algorithm level. The decision was to keep it, so let's keep it here too. But then this should be fixed indeed; will do so now

include/graphblas/algorithms/conjugate_gradient.hpp
anyzelman: Error code? (If at some point the code flow changes so that it actually does call the preconditioner even though it should not, then the FAILED error code will propagate)

anyzelman:
I'm happy again:) Concept release notes:

This MR resolves a feature request for a preconditioned CG accessible via a transition path. To enable this feature, this MR adds the following:
- standard left-sided preconditioning to the ALP CG has been added. There are now two entry points `grb::algorithms::conjugate_gradient` for non-preconditioned CG, and `grb::algorithms::preconditioned_conjugate_gradient` for PCG. They are kept separate since PCG requires one more work-space vector. The preconditioner is given as an `std::function` that returns a `grb::RC`, takes an output `grb::Vector` and an input `const grb::Vector`. Given it's an `std::function`, any state required by the preconditioner can be embedded into the function (there is no additional argument required for this).
- the sparse solver transition path API now can take preconditioners also. Since this is a C API, the preconditioner is given as a function pointer. The return type is an `int` where `0` indicates successful application of the preconditioner. The first argument is a raw output vector, the second a const input vector, and the third a `void *` argument to any state required by the preconditioner.
- the kml_iss.h transition path API now also takes preconditioners, given as a function pointer with signature according to the current KML solver specification.

The MR also expands and improves documentation regarding the transition path libraries.

This MR adds the following smoke test:
- preconditioned CG with simple Jacobi preconditioning, both for real (`double`) and the complex case. Note that for the complex case, Jacobi preconditioning may not be a good choice; it is not encouraged to use outside of smoke testing.

To enable the smoke tests, this MR additionally includes the following bugfixes:
- the vector-to-vector foldr was not implemented in the BSP1D and hybrid backends, herewith fixed;
- the eWiseLambda(vector) for BSP1D and hybrid was using local indexing, while it should have been global;
- a meta-bug in the distributed implementation of `grb::set(BSP1D,hybrid)` erroneously relied on local indexing of eWiseLambda.

As always and additionally, the MR include code style fixes in both related and unrelated files.

anyzelman:
CI and tests OK

alberto-scolari (approved):

PR #314, opened 2024-02-21
268 Provide grb::set(Matrix out, Matrix mask, Matrix in)

### LINKED ISSUE #268: Provide grb::set(Matrix out, Matrix mask, Matrix in)
Same as [this variant](http://albert-jan.yzelman.net/alp/user/group__IO.html#gaa0fc695dbdc8494058d636cb03a9443d) but for Matrix.

aleksamilisavljevic:
Closes #268


include/graphblas/nonblocking/io.hpp
```diff
@@ -1217,6 +1217,41 @@ namespace grb {
...
+		}
+
+		// nonblocking execution is not supported
+		// first, execute any computation that is not completed
+		internal::le.execution();
+
+		// second, delegate to the reference backend
+		return set< descr >(
+			internal::getRefMatrix( C ),
+			internal::getRefMatrix( Mask ),
```
byjtew: Variable name should be `mask` or `M`, not `Mask` otherwise it looks like a type-name

include/graphblas/reference/io.hpp
```diff
@@ -1156,6 +1156,206 @@ namespace grb {
...
+		OutputType * mask_valbuf = nullptr;
+		internal::getMatrixBuffers( mask_arr, mask_buf, mask_valbuf, 1, Mask );
+
+		internal::Coordinates< reference > mask_coors;
+		mask_coors.set( mask_arr, false, mask_buf, ncols );
+
+		for( size_t i = 0; i < nrows; ++i ) {
+			mask_coors.clear();
+			for( auto k = mask_raw.col_start[ i ]; k < mask_raw.col_start[ i + 1 ]; ++k ) {
+				const size_t k_col = mask_raw.row_index[ k ];
```
byjtew:
```suggestion
				const auto k_col = mask_raw.row_index[ k ];
```
Since we use `RTI`, `CIT`, etc.

include/graphblas/reference/io.hpp
```diff
@@ -1156,6 +1156,206 @@ namespace grb {
...
+				if( utils::interpretMask< descr, MaskType >( true, mask_raw.values, k ) ) {
+					mask_coors.assign( k_col );
+				}
+			}
+			for( auto k = A_raw.col_start[ i ]; k < A_raw.col_start[ i + 1 ]; ++k ) {
+				const size_t k_col = A_raw.row_index[ k ];
+				if( mask_coors.assigned( k_col ) ) {
+					nzc++;
+				}
+			}
```
byjtew: openMP reduction possible easily

include/graphblas/reference/io.hpp
```diff
@@ -1156,6 +1156,206 @@ namespace grb {
...
+
+		internal::Coordinates< reference > mask_coors;
+		mask_coors.set( mask_arr, false, mask_buf, ncols );
+
+		for( size_t i = 0; i < nrows; ++i ) {
+			mask_coors.clear();
+			for( auto k = mask_raw.col_start[ i ]; k < mask_raw.col_start[ i + 1 ]; ++k ) {
+				const size_t k_col = mask_raw.row_index[ k ];
+				if( utils::interpretMask< descr, MaskType >( true, mask_raw.values, k ) ) {
+					mask_coors.assign( k_col );
```
byjtew: ~There's an openMP version of the assign, you can find it in other places in the code using the function `assignAsync` if you're in a parallel region~

include/graphblas/reference/io.hpp
```diff
@@ -1156,6 +1156,206 @@ namespace grb {
...
+		auto &CRS_raw = internal::getCRS( C );
+		auto &CCS_raw = internal::getCCS( C );
+
+		config::NonzeroIndexType * C_col_index = internal::template
+			getReferenceBuffer< typename config::NonzeroIndexType >( ncols + 1 );
+
+		CRS_raw.col_start[ 0 ] = 0;
+
+#ifdef _H_GRB_REFERENCE_OMP_BLAS3
+			#pragma omp parallel for simd
```
byjtew:
```suggestion
		#pragma omp parallel for simd
```

include/graphblas/reference/io.hpp
```diff
@@ -1156,6 +1156,206 @@ namespace grb {
...
+					mask_coors.assign( k_col );
+				}
+			}
+			for( auto k = A_raw.col_start[ i ]; k < A_raw.col_start[ i + 1 ]; ++k ) {
+				const size_t k_col = A_raw.row_index[ k ];
+				if( mask_coors.assigned( k_col ) ) {
+					OutputType val = A_raw.getValue( k, *( (OutputType*) nullptr ) );
+					CRS_raw.row_index[ nzc ] = k_col;
+					CRS_raw.setValue( nzc, val );
+					const size_t CCS_index = C_col_index[ k_col ]++ + CCS_raw.col_start[ k_col ];
```
byjtew:
```suggestion
					const size_t CCS_index = C_col_index[ k_col ] + CCS_raw.col_start[ k_col ];
					C_col_index[ k_col ]++;
```

include/graphblas/reference/io.hpp
```diff
@@ -1156,6 +1156,206 @@ namespace grb {
...
+					nzc++;
+				}
+			}
+		}
+
+		for( size_t j = 0; j < ncols; ++j ) {
+			assert( CCS_raw.col_start[ j + 1 ] - CCS_raw.col_start[ j ] ==
+				C_col_index[ j ] );
+		}
+		assert( old_nzc == nzc );
```
byjtew:
```suggestion
#ifndef NDEBUG
		for( size_t j = 0; j < ncols; ++j ) {
			assert( CCS_raw.col_start[ j + 1 ] - CCS_raw.col_start[ j ] ==
				C_col_index[ j ] );
		}
		assert( old_nzc == nzc );
#endif
```

byjtew (dismissed):
Nice job, as for the unit-test, you could rely on the matrix factories that we have.
e.g: L masked by I should return I
L being a Lower triangular matrix 
I being an identity matrix

include/graphblas/reference/io.hpp
```diff
@@ -1156,6 +1156,206 @@ namespace grb {
...
+		internal::getMatrixBuffers( mask_arr, mask_buf, mask_valbuf, 1, Mask );
+
+		internal::Coordinates< reference > mask_coors;
+		mask_coors.set( mask_arr, false, mask_buf, ncols );
+
+		for( size_t i = 0; i < nrows; ++i ) {
+			mask_coors.clear();
+			for( auto k = mask_raw.col_start[ i ]; k < mask_raw.col_start[ i + 1 ]; ++k ) {
+				const size_t k_col = mask_raw.row_index[ k ];
+				if( utils::interpretMask< descr, MaskType >( true, mask_raw.values, k ) ) {
```
byjtew: Will not compile if you test with void matrices, I am currently developing a version of interpretMask for matrices because this one only works for vectors

byjtew:
```c++
/** Specialisation for void-valued matrice's masks */
		template< Descriptor descriptor, typename MatrixDataType, typename ValuesType >
		static bool interpretMatrixMask(
			const bool &assigned,
			const ValuesType * const values,
			const size_t k,
			typename std::enable_if<
				!std::is_void< MatrixDataType >::value
			>::type * = nullptr
		) {
			return interpretMask< descriptor, ValuesType >( assigned, values, k );
		}

		/** Specialisation for void-valued matrice's masks */
		template< Descriptor descriptor, typename MatrixDataType, typename ValuesType >
		static bool interpretMatrixMask(
			const bool &assigned,
			const ValuesType * const,
			const size_t,
			typename std::enable_if<
				std::is_void< MatrixDataType >::value
			>::type * = nullptr
		) {
			// set default mask to false
			bool ret = assigned;
			// check whether we should return the inverted value
			if( descriptor & descriptors::invert_mask ) {
				return !ret;
			} else {
				return ret;
			}
		}
```

Add these two functions in `utils.hpp` and call them like this `utils::interpretMatrixMask< descr, MaskType >( true, mask_raw.getValues(), k_col )`

aleksamilisavljevic: Done, thanks!

include/graphblas/reference/io.hpp
aleksamilisavljevic: Done, thanks!

include/graphblas/reference/io.hpp
aleksamilisavljevic: Done, thanks!

anyzelman:
@aleksamilisavljevic @byjtew -- apologies, I thought this one was merged ages ago but found it still open just now. I've rebased it (and weirdly enough indeed a component from this MR was already in utils?) and will look at it more shortly: at the moment it builds but not sure if it passes unit tests, plus I see a spec for the missing primitive should still be provided). I'll take those remainder, hopefully small, steps over the coming days:)

anyzelman:
It's back up with unit tests passing, still a few TODOs:

- [x] check if it makes sense to unit test non-void to void, no-casting vs. casting, and void-to-void. At present, the unit test goes for non-casting and non-void to non-void only
- [x] unit test should check passing empty masks
- [x] BSP1D and hybrid implementations are missing
- [x] reference_omp variant is not yet fully parallel

anyzelman:
Upgraded unit test -- running the full suite before continuing some other time

anyzelman:
All tests pass at this point, by the way (test results from over the weekend and including for LPF). Only one TODO remaining, which is the parallelisation of this function in the `reference_omp` backend.

anyzelman:
Since it's almost done, some draft release notes:

A matrix copy operation with output mask (`grb::set( matrix, matrix, matrix )`) is added. This includes a specification as well as implementations for all current backends, plus an extension to make the unit test for all variants of `grb::set` that operate on matrices exhaustive in coverage (including for masked matrix copies, with and without casting, etc.).

Additionally, this MR:
 - moves the SPA buffer management class from `blas3.hpp` to `matrix.hpp`;
 - fixes incomplete handling of the `no_casting` descriptor of `grb::set` with matrix outputs;
 - fixes a missing include in `bsp1d/init.hpp`;
 - cleans up minor inaccuracies and typos to the specifications in `io.hpp`;
 - expands on the specification of mask inversion in the case of matrix outputs;
 - improves debug tracing; and,
 - as always, includes minor code style fixes.

Big thanks to @aleksamilisavljevic for providing the implementation for the reference backend, now finally merged!

anyzelman:
All tests indeed pass. Now factoring in the crs_only descriptor and then this should be ready for merge=)

anyzelman (dismissed):

anyzelman:
@byjtew Hi Benjamin - you had requested changes here a long long time ago -- just making sure if dismissing your review is ok for you, or whether you want to have a look (and have the time for it--ofc totally ok if not!)

(It's just that github would block the merge in the current situation given your past review:)

byjtew:
> @byjtew Hi Benjamin - you had requested changes here a long long time ago -- just making sure if dismissing your review is ok for you, or whether you want to have a look (and have the time for it--ofc totally ok if not!)
> 
> 
> 
> (It's just that github would block the merge in the current situation given your past review:)

Yes sorry, you can dismiss my review indeed ahah @anyzelman

anyzelman (approved):

PR #315, opened 2024-02-23
More testing found a bug in kml_iss implementation, herewith fixed

anyzelman:
Hotfix for the previous commit -- more usage of the KML transition path found that certain getters and setters seemed to have no effect, which is fixed with this MR. The expected impact of this bug is expected minimal, as there has been minimal time passed while the bug was present in the stable branch.


anyzelman:
Review OK, manual testing OK. Waiting approval from @alberto-scolari and CI

alberto-scolari (approved):

PR #336, opened 2024-04-09
335 wrong fma perf test

### LINKED ISSUE #335: Wrong fma perf test
The `fma` test is wrong, reporting

```
Benchmark label: grb::eWiseApply (axpy) of size 10
1 (expected) does not equal 2 (template optimised) at position 0.
Outer iteration #0 timings (io, preamble, useful, postamble, time since epoch): 0.000000, 0.010252, 0.000185, 0.000000, 1712665968227
1 (expected) does not equal 2 (template optimised) at position 0.
Outer iteration #1 timings (io, preamble, useful, postamble, time since epoch): 0.000000, 0.013413, 0.000264, 0.000000, 1712665969227
1 (expected) does not equal 2 (template optimised) at position 0.
Outer iteration #2 timings (io, preamble, useful, postamble, time since epoch): 0.000000, 0.020564, 0.000199, 0.000000, 1712665970227
...
```

The problem is probably due to the change about the in-place semantics.

The very problem is that the error went long time unnoticed.

alberto-scolari:
Merge fix

### DIFF
--- a/tests/performance/CMakeLists.txt
+++ b/tests/performance/CMakeLists.txt
@@ -26,6 +26,7 @@ set( TEST_CATEGORY "performance" )
 add_library( bench_kernels OBJECT bench_kernels.c bench_kernels.h )
 add_library( bench_kernels_omp OBJECT bench_kernels.c bench_kernels.h )
 target_compile_definitions( bench_kernels_omp PRIVATE BENCH_KERNELS_OPENMP )
+target_link_libraries( bench_kernels_omp PRIVATE test_performance_flags OpenMP::OpenMP_C )
 
 add_grb_executables( fma fma.cpp $<TARGET_OBJECTS:bench_kernels>
 	BACKENDS reference NO_BACKEND_NAME
--- a/tests/performance/CMakeLists.txt
+++ b/tests/performance/CMakeLists.txt
@@ -26,6 +26,7 @@ set( TEST_CATEGORY "performance" )
 add_library( bench_kernels OBJECT bench_kernels.c bench_kernels.h )
 add_library( bench_kernels_omp OBJECT bench_kernels.c bench_kernels.h )
 target_compile_definitions( bench_kernels_omp PRIVATE BENCH_KERNELS_OPENMP )
+target_link_libraries( bench_kernels PRIVATE test_performance_flags )
 target_link_libraries( bench_kernels_omp PRIVATE test_performance_flags OpenMP::OpenMP_C )
 
 add_grb_executables( fma fma.cpp $<TARGET_OBJECTS:bench_kernels>
--- a/tests/performance/CMakeLists.txt
+++ b/tests/performance/CMakeLists.txt
@@ -33,8 +33,13 @@ add_grb_executables( fma fma.cpp $<TARGET_OBJECTS:bench_kernels>
 	ADDITIONAL_LINK_LIBRARIES "rt"
 )
 
-add_grb_executables( fma-openmp fma.cpp $<TARGET_OBJECTS:bench_kernels_omp>
+add_grb_executables( fma-blocking fma.cpp $<TARGET_OBJECTS:bench_kernels_omp>
 	BACKENDS reference_omp NO_BACKEND_NAME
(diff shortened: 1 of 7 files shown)


anyzelman:
This one is about done except it ran into #340 . Must fix that issue first before merging this one. If everything is all "right", the CIs should fail on the latest push (because of #340).

anyzelman:
> This one is about done except it ran into #340 . Must fix that issue first before merging this one. If everything is all "right", the CIs should fail on the latest push (because of #340).

actually all CIs are happy. This is because the fma test is a perftest, and the CI doesn't run those on every branch and push. This emphasizes further the importance of adding a unit test that covers this case (and similar cases) as part of #340 / #341 . That issue blocks resolving this MR. (NB.: that issue/MR already includes a confirmed fix for this branch.)

anyzelman:
Concept release notes:

The fma (fused multiply-add) performance test aims to replicate a simple triad benchmark, computing `z=alpha.*x+y` out-of-place.

The matching ALP/GraphBLAS primitive for that, once upon a time, was `grb::eWiseMulAdd`. Since that time, however, that primitive has undergone two changes: 1) it became in-place rather than out-of-place (i.e., computing `z+=alpha.*x+y`) while 2) since the release of the nonblocking backend this function has become deprecated (as the same could be computed using `grb::eWiseMul` followed by `grb::foldl` or `grb::foldr`, at negligible performance overheads compared to the one-shot `grb::eWiseMulAdd`).

Neither change was ever reflected in the `fma` test itself, however; and, at present, the original out-of-place `fma` computation `z=alpha.*x+y` similarly has no single matching (deprecated or not) ALP/GraphBLAS primitive. The nonblocking backend, however, could implement an fma by first calling `grb::set(z,y)` followed by the in-place `grb::eWiseMul`. 

Additionally, since the incorporation of the CMake build infra, the compiler-optimised performance tests were erroneously not compiled with performance flags, causing a further apple-to-oranges type comparison in this performance test.

The issues were detected by Alberto Scolari, who additionally indicated that adding performance tests to the CI and regularly reviewing those results would have caught this issue earlier (crossref #343 , visualisation only - the performance test now actually regularly trigger for `develop` on our internal CI). He also produced an initial set of fixes and we thank Alberto for both.

This MR fixes these issues regarding the fma performance test by:
 1. introducing the fma-blocking and fma-nonblocking shared-memory parallel performance test instead of a single fma-omp one. Only for fma-nonblocking is performance expected to match that with a compiler-optimised or lambda-generated fma;
 2. this distinction is now also printed to screen as part of the test run; and
 3. improved test initialisation checks and improved error detection.

This MR also introduced some changes to the `dot` and `reduce` performance tests, in tandem with changes to the `fma` performance test. Those "joint" changes are:
 4. the compiler-optimised shared-memory parallel fma (as well as the reduce and dot) performance tests now avoid critical sections (which is in line with standard HPC practices), and e.g. for reductions (reduce and dot) prefer the OpenMP `reduction`-clause;
 5. the compiler-optimised shared-memory parallel fma, reduce, and dot now use the same distribution of work across threads (essentially a static block-wise schedule);
 6. both sequential and shared-memory parallel compiler-optimised benchmarks are now given the proper performance flags during compilation;
 7. avoid mixing `fprintf` and `stdcerr/stdcout` in the fma, dot, and reduce performance tests as well as other more minor code style fixes;
 8. test output and test summary now distinguish between sequential and shared-memory parallel tests;
 9. use the `dense` descriptor where appropriate (potentially enhancing performance of ALP code);
 10. if any test auto-tunes the number of inner repetitions, the number is chosen so that a single experiment is expected to take no less than 100 milliseconds (down from 1 second);
 11. these tests now employ vectors of 100 million elements (up from 10 million); and, finally,
 12. it is double-checked that the numbers the new tests produce are on par with - or exceeding the performance of - the numbers published by Yzelman et al. (2020).

anyzelman (approved):
Code review OK, awaiting internal CI OK as well as manual perftest run ok.

anyzelman:
Manual perftests OK and fma, reduce, and dot performance double-checked with '20 paper. Internal CI also OK. Will merge

PR #339, opened 2024-10-09
338 request for fuselets

### LINKED ISSUE #338: Request for fuselets
A request came in for accelerating process-local parts of existing HPC solvers using the nonblocking backend. This should be easy to do, and this issue proposes to demo an easily-extensible set of basic level-1, level-2, and mixed level-{1,2} sequences that can be fused. We will call elements of these sets a _fuselet_, and expose the set as part of the transition path as _fuselets_.

anyzelman:
MR that provides so-called _fuselets_.

Fuselets are sequences of operations that may occur in existing code bases, and could be accelerated using the nonblocking backend. A fuselet represents such a sequence of operations, which are exposed as a single plain-C function call. ALP during its standard build will generate a library of those fuselets using its nonblocking backend, thus resulting in high-performance, automatically fused kernels.

The fuselets furthermore, at run-time and again automatically, tunes performance parameters such as tile sizes and the number of threads to be used.

Finally, the example fuselets provided in `include/transition/fuselets.h` and `src/transition/fuselets.cpp` demonstrate how easy it is to use ALP to provide arbitrary fuselets that existing code bases may require, and are easily modified or extended.

This MR also includes:
 - smoke and performance tests for the new fuselets;
 - standard semiring and monoid definitions in the new `grb::monoids` and `grb::semirings` namespaces;
 - tests for those standard semirings and monoids (insofar possible);
 - a bugfix in the stdout of the spmspm unit test;
 - a bugfix for missing cstdint includes in various headers; and
 - as always, minor code style fixes.

### DIFF
--- a/include/graphblas.hpp
+++ b/include/graphblas.hpp
@@ -170,6 +170,12 @@
  *  -# #grb::Monoid, and
  *  -# #grb::Semiring.
  *
+ * A list of standard operators, monoids, and semirings are provided in their
+ * respective name spaces:
+ *  -# #grb::operators,
+ *  -# #grb::monoids, and
+ *  -# #grb::semirings.
+ *
  * Binary operators are parametrised in two input domains and one output domain,
  * \f$ D_1 \times D_2 \to D_3 \f$. The \f$ D_i \f$ are given as template
  * arguments to the operator. A #grb::Monoid is composed from a binary operator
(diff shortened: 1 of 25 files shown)


anyzelman:
Code review nearly complete. Could remove some element-wise scalar divisions, but need to double-check (lack of) impact on performance. After that, should be ready for merge.

anyzelman:
Code review complete. Waiting CI OKs before merge.

anyzelman:
All CIs pass, manual full test suite runs OK, manual checks of fuselets speedup (vs. blocking) also pass. Will merge.

PR #350, opened 2025-03-17
Add early exit in preconditioned CG when initial residual is below tolerance

### LINKED ISSUE #349: Early exit in preconditioned CG when initial residual is below tolerance
In the `preconditioned_conjugate_gradient`  implementation, the algorithm applies the preconditioner before checking if the current residual already meets the required tolerance. This change would allow the function to return immediately when the initial residual is small enough, avoiding the unnecessary cost of a preconditioner call.

Details:

- After computing the residual (i.e., after the initial computation of `r = b - A*x` and before applying the preconditioner), we can check if `sqrt(residual) < tol`.
- If true, the function can immediately return with success, setting iterations to 0.
- This change can improve performance in cases where the initial guess is already close to the solution.

ChristosMatzoros:
Closes #349

### DIFF
--- a/include/graphblas/algorithms/conjugate_gradient.hpp
+++ b/include/graphblas/algorithms/conjugate_gradient.hpp
@@ -358,9 +358,12 @@ namespace grb {
 				);
 			assert( ret == grb::SUCCESS );
 
-			// get residual. In the preconditioned case, the resulting scalar is *not*
-			// used for subsequent operations. Therefore, we first compute the residual
-			// using alpha as a temporary scalar
+			// get effective tolerance
+			if( ret == grb::SUCCESS ) {
+				tol *= std::sqrt( grb::utils::is_complex< IOType >::modulus( bnorm ) );
+			}
+
+			// get residual
 			alpha = zero;
 			ret = ret ? ret : grb::dot< descr_dense >(
 					alpha,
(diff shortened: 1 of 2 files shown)


include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -348,6 +348,33 @@ namespace grb {
 			ret = ret ? ret : grb::foldl< descr_dense >( r, temp, minus );
 			assert( ret == grb::SUCCESS );
 
+			// bnorm = b' * b;
+			bnorm = zero;
+			ret = ret ? ret : grb::dot< descr_dense >(
+				bnorm,
+				b, b,
+				ring.getAdditiveMonoid(),
+				grb::operators::conjugate_left_mul< IOType >() );
```
anyzelman: I know it's cut-and-pasted, but could you bring the code style here in line with 364-369?

anyzelman (changes requested):
Also please add your name to the `NOTICE` file under this MR. After that and the minor flagged issue are done, I think this one can be merged immediately -- many thanks for contributing!

include/graphblas/algorithms/conjugate_gradient.hpp
ChristosMatzoros: Hi @anyzelman, I made the change with my next commit

anyzelman:
Rebased prior to merge

anyzelman (approved):

anyzelman:
Running all smoke tests with the changes from this MR, which should be enough to detect any issues with the proposed changes. (Running them with LPF.)

Concept release notes:

Previously, the Conjugate Gradients (CG) algorithm applies the preconditioner before checking if the current residual already meets the required tolerance. This MR instead allows CG to return immediately when the initial residual is small enough, avoiding the unnecessary cost of a preconditioner call. This change can improve performance in cases where the initial guess is already close to the solution. Thanks to @ChristosMatzoros for proposing this enhancement!

anyzelman:
All smoke tests passed, will merge. Thank you @ChristosMatzoros !

PR #353, opened 2025-04-07
352 introduce a numa aware sptrsv

### LINKED ISSUE #352: Introduce a NUMA-aware sptrsv
This sptrsv will use the schedule provided by OneStopParallel, as sketched eg. here:

https://arxiv.org/abs/2503.05408

Ideally, the sptrsv will use the type structures introduced by Spampinato et al. here:

https://dl.acm.org/doi/abs/10.1145/3589246.3595373

anyzelman:
Closes issue #352 .

### DIFF
--- a/include/graphblas/base/blas2.hpp
+++ b/include/graphblas/base/blas2.hpp
@@ -340,6 +340,8 @@ namespace grb {
 	/**
 	 * Computes \f$ x \f$ from \f$ Tx=b \f$.
 	 *
+	 * \warning This is an experimental feature.
+	 *
 	 * Here, \f$ T, b \f$ are given while \f$ T \f$ additionally must be either
 	 * lower- or upper-triangular. The output \f$ x \f$ may furthermore be masked.
 	 *
@@ -432,6 +434,8 @@ namespace grb {
 
 	/**
 	 * Computes \f$ x \f$ from \f$ Tx = b \f$, unmasked variant.
+	 *
+	 * \warning This is an experimental feature.
 	 *
 	 * \todo Extend documentation
 	 */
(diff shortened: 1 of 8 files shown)


anyzelman:
At this point all unit tests and smoke tests pass

anyzelman:
At this point also all unit & smoke tests are OK

anyzelman:
At this point, all smoke and unit tests pass. Also using OSP's simulator, the sptrsv code introduced here is thoroughly verified, for all three sorting modes.

Some next items remain:

- [ ] adapt build infra to take OSP as an optional module;
- [ ] spec out a tune step (suggest introducing `grb::PHASE::TUNE`);
- [ ] when OSP is enabled, allow the sptrsv tuning step to call OSP;
- [ ] optionally (or alternatively) allow reading-in OSP schedules from some external source;
- [ ] expose a transition path for the unmasked sptrsv with dense input/output vectors;
- [ ] create a performance test driver for (tuned) sptrsv with dense input/output vectors;
- [ ] implement and unit-test combinations of masked sptrsv and sptrsv with sparse I/O vectors;
- [ ] relatedly, expand the current-quite-basic unit test for the unmasked sptrsv with dense I/O vectors - ideally use a golden output from an external software package like Octave on a small (e.g. 550x550) synthetic system.

No specific milestone will be set for this new feature, but please raise it if this functionality is urgent and we can attempt take that into account. This MR will not address the triangular type structure from Spampinato et al.- ideally first the dense backend (with dense trsv support) is merged after which a new MR can enable sptrsv support through the functionalities provided by this MR.

anyzelman:
Target set to v0.9

PR #361, opened 2025-06-05
360 transition path allocates workspace vector data for sparse information

### LINKED ISSUE #360: Transition path allocates workspace vector data for sparse information...
...while e.g. in the case of PCG's transition path, those vectors are guaranteed to be always dense. In some cases this results in a noticeable memory overhead that should be eliminated.

anyzelman:
Resolves and closes #360 

TODOs:

- [x] test current version (@ChristosMatzoros :)
- [x] add transition path option to default-allocate the workspace without the workspace required for the preconditioner
- [x] add transition path option for the user to pass the workspace explicitly, either with or without preconditioner workspace
- [x] add transition path option to default-allocate using libnuma for cross-NUMA deployments

Note when not passing a preconditioner workspace or when not asking for it: in this case, when the user still calls set_preconditioning, the current/past logic should still kick in: we allocate a vector using the standard ALP allocation mechanism, initialise the preconditioner workspace, and still allow (therefore) the use of a preconditioned solve. We can opt to print a warning, but that's optional (and i don't think required/useful)

Also tag @djelovina for possible interest

### DIFF
--- a/include/graphblas/algorithms/conjugate_gradient.hpp
+++ b/include/graphblas/algorithms/conjugate_gradient.hpp
@@ -314,43 +314,77 @@ namespace grb {
 					std::cerr << "Error: at least one CG iteration must be requested\n";
 					return grb::ILLEGAL;
 				}
+
+				// dense descriptor
+				if( descr & grb::descriptors::dense ) {
+					if( grb::nnz( x ) != grb::size( x ) ) {
+						std::cerr << "Error: x was sparse while the dense descriptor was given\n";
+						return grb::ILLEGAL;
+					}
+					if( grb::nnz( b ) != grb::size( b ) ) {
+						std::cerr << "Error: b was sparse while the dense descriptor was given\n";
+						return grb::ILLEGAL;
+					}
+					if( grb::nnz( r ) != grb::size( r ) ) {
+						std::cerr << "Error: r was sparse while the dense descriptor was given\n";
+						return grb::ILLEGAL;
+					}
+					if( grb::nnz( temp ) != grb::size( temp ) ) {
+						std::cerr << "Error: temp was sparse while the dense descriptor was "
+							<< "given\n";
+						return grb::ILLEGAL;
+					}
+					if( preconditioned &&
+						(grb::nnz( temp_precond ) != grb::size( temp_precond ))
+					) {
+						std::cerr << "Error: temp_precond was sparse while the dense descriptor "
+							<< "was given\n";
+						return grb::ILLEGAL;
+					}
+				}
 			}
 
 			// set pure output fields to neutral defaults
 			iterations = 0;
 			residual = std::numeric_limits< double >::infinity();
 
-			// make x and b structurally dense (if not already) so that the remainder
+			// declare internal scalars
+			IOType sigma, bnorm, alpha, beta;
+
+			// make x structurally dense (if not already) so that the remainder
 			// algorithm can safely use the dense descriptor for faster operations
-			{
-				RC rc = grb::SUCCESS;
+			grb::RC ret = grb::SUCCESS;
+			if( !(descr & grb::descriptors::dense) ) {
 				if( nnz( x ) != n ) {
-					rc = grb::set< descriptors::invert_mask | descriptors::structural >(
+					ret = grb::set< descriptors::invert_mask | descriptors::structural >(
 						x, x, zero
 					);
+					assert( ret == grb::SUCCESS );
 				}
-				if( rc != grb::SUCCESS ) { return rc; }
 				assert( nnz( x ) == n );
 			}
 
-			IOType sigma, bnorm, alpha, beta;
-
 			// r = b - temp;
-			grb::RC ret = grb::set( temp, 0 ); assert( ret == grb::SUCCESS );
+			ret = ret ? ret : grb::set< descr >( temp, 0 );
+			assert( ret == grb::SUCCESS );
 			ret = ret ? ret : grb::mxv< descr_dense >( temp, A, x, ring );
 			assert( ret == grb::SUCCESS );
-			ret = ret ? ret : grb::set( r, zero ); assert( ret == grb::SUCCESS );
-			// note: no dense descriptor since we actually allow sparse b
-			ret = ret ? ret : grb::foldl( r, b, ring.getAdditiveMonoid() );
-			// from here onwards, r, temp, x are dense and will remain so
-			assert( nnz( r ) == n );
-			assert( nnz( temp ) == n );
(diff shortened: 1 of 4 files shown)


ChristosMatzoros:
Dear @anyzelman, In the pipeline where I test the transition path of cg, I do the normal initialization (sparse_cg_init_dii ) and set my own preconditioner method using sparse_cg_set_preconditioner_dii. The segmentation fault is happening when executing the solve method sparse_cg_solve_dii. Here is the segmentation fault (I put ... wherever I should not disclose info):

Thread 2 "..." received SIGSEV, Segmentation fault.
grb::internal:: Coordinates<(grb:: Backend)3>:: asyncSubsetInit (upper_bound=<optimized out>, lower_bound <optimized out>, this <optimized out>) at ... /ALP/include/graphblas/nonblocking/coordinates.hpp:509
509  config::VectorIndexType *local_stack = local_buffer[ tile_id ] + 1;

the backtrace goes like that:
Segmentation fault in line 509 of ALP/include/graphblas/nonblocking/coordinates.hpp
    --> Called by line 943 of ALP/src/graphblas/nonblocking/pipeline.cpp

ChristosMatzoros:
Dear @anyzelman 

Here is a backtrace from Thread 1:

0 grb::internal:: Coordinates<(grb:: Backend)3>:: asyncSubsetInit (upper_bound=, lower_bound , this ) at ... /ALP/include/graphblas/nonblocking/coordinates.hpp:509
1 (omp realated)
2 in GOMP_parallel
3 in internal::Pipelineexecution at ALP/src/graphblas/nonblockingh/pipeline.cpp
4 in grb::internal::LazyEvaluation::addStage(..).. at ALP/src/graphblas/nonblocking/lazy_evaluation.cpp:367
5 in grb::internal::dot_generic<..> .. at ALP/include/graphblas/nonblocking/blas1.hpp:10728
6. grb::dot<...> at  ALP/include/graphblas/nonblocking/blas1.hpp:10835
7. in grb::dot<...> ... at ALP/include/graphblas/nonblocking/blas1.hpp:10821
8. grb::algorithms::preconditioned_conjugate_gradient<...> .. at ALP/include/graphblas/algorithms/conjugate_gradient.hpp:399
9. CG_Data<double, int, int> at ALP/src/transition/solver.cpp:333
10. sparse_cg_solve_impl<double, int, int> at ALP/src/transition/solver.cpp:868

anyzelman:
Thanks - the trace indicates sparse data is still being used, which should not happen since we pass in the `dense` descriptor. While actually the details of the trace are confusing (a MWE might help) - I did find a possible culprit: there are some statements in the PCG algorithm in `algorithms/conjugate_gradient.hpp` that ignore any descriptors that the user may pass in, including our hint for dense vectors. I'm fixing that up now and maybe that already resolves the bug on your end:)

anyzelman:
Pushed a fix

anyzelman:
Note to self, I forgot about this statement:

> Note when not passing a preconditioner workspace [but asking for a preconditioned solve anyway,] the current/past logic should still kick in: we allocate a vector using the standard ALP allocation mechanism

I instead let it raise an error if this is done. Not sure what is more sensible to do -- maybe @ChristosMatzoros , do you have a preference?

anyzelman:
FWIW, I'm now leaning to follow that comment and simplify the API a bit -- taking the numa and the no-preconditioner-workspace flags as hints that can be overridden by the transition path when required.

ChristosMatzoros:
I attach an MWA but I need to attach it as .txt (you can change to .cpp)
[example.txt](https://github.com/user-attachments/files/21228852/example.txt)

anyzelman:
Note - rebased on latest master

anyzelman:
Note - rebased again; waiting for @ChristosMatzoros' feedback before finishing the last couple of TODOs

ChristosMatzoros:
@anyzelman  just tested and its working as expected, without loss of performance

include/graphblas/algorithms/conjugate_gradient.hpp
```diff
@@ -314,43 +314,77 @@ namespace grb {
 					std::cerr << "Error: at least one CG iteration must be requested\n";
 					return grb::ILLEGAL;
 				}
+
+				// dense descriptor
+				if( descr & grb::descriptors::dense ) {
```
djelovina, line 319: I would be in favor of having `if constexpr` here, which would require `C++17`. Alternatively I propose to move these block into a function calls templated to descriptors. I mention this here because sometimes compilation fails on unused if blocks. @anyzelman 

Even more critical is resolve similar problems in blas*hpp files.

anyzelman: Actually I tested a very long time ago that in C++11, both GCC and LLVM still erase this if-statement (there is no assembly corresponding to this if). I tested it specifically back in '16 or '17 because I too was worried about this. The C++17 formalisation that turns this into a requirement is great, but I'm very conservative about bumping the C version requirements. We can check internally what the product folks accept as the "maximum" C++ version (also the MLIR / JIT work actually required C++17 as I recall, making makes bumping the C++ version also a useful feature requirement--which is more compelling to me)

anyzelman:
> FWIW, I'm now leaning to follow that comment and simplify the API a bit -- taking the numa and the no-preconditioner-workspace flags as hints that can be overridden by the transition path when required.

Made these changes in the latest commits. Will now move on to the last TODO and then hopefully close this MR=)

include/graphblas/algorithms/conjugate_gradient.hpp
anyzelman: > sometimes compilation fails on unused if blocks

Sorry I missed this part -- compilation shouldn´t fail of course, if it does, could you make an issue with such an instance?

djelovina: Sure, here https://github.com/Algebraic-Programming/ALP/issues/371#issue-3259043347

anyzelman: Thanks! Resolving here, and following up in #371

anyzelman:
@ChristosMatzoros - the last feature request is now also implemented: transition path users can now also manually provide their own workspace buffers. I've reviewed myself (usually not the best idea) and to me everything looks OK to merge -- could you review this MR as well, and also test it in your use case once more to triple-check nothing gets broken (and the SPA memory is indeed now not allocated in that use case) ?

ChristosMatzoros (approved):
Dear @anyzelman,

I tested the changes on my use case, and everything works as expected, both in terms of performance and correctness. The code looks clean, and the comments are very descriptive. Approved!

PR #362, opened 2025-06-10
Thread number configuration during runtime

### LINKED ISSUE #351: Runtime thread configuration for ALP's transition path
Currently, when using ALP's transition path (e.g., with the CG method), the maximum number of threads is determined by the `OMP_NUM_THREADS ` environment variable. However, attempts to dynamically set the thread count via `omp_set_num_threads()` in the calling code (e.g., in `main()`) have no effect on ALP's behavior.

This occurs because ALP initializes the thread count via a **static global variable**, which is resolved before `main()` executes. Consequently, runtime adjustments using OpenMP's API are ignored.

While one potential fix would be to query the active thread count dynamically (e.g., via `omp_get_max_threads()` in `numThreads()` (see `init.hpp` in nonblocking)), this could introduce **work imbalance** as there are assumptions that was already made based on the original value of the number of threads before the execution of `main()`. Existing logic may rely on the initial static thread count, and altering it mid-execution might lead to unintended side effects.

Requested Feature:
A mechanism to safely configure ALP's thread count **at runtime** (e.g., through an explicit ALP API call or controlled re-initialization) without disrupting internal assumptions.

ChristosMatzoros:
This PR introduces a temporary fix to issue #351 

The idea is to get the current number of threads during runtime, every time we call the analytic model instead of getting this value once (statically before the start of the execution).

All smoke tests have passed.

Note: This is a temporary solution based on specific assumptions. Further investigation is required to ensure long-term stability and to identify a more robust fix.

### DIFF
--- a/src/graphblas/nonblocking/init.cpp
+++ b/src/graphblas/nonblocking/init.cpp
@@ -38,7 +38,7 @@ bool grb::internal::NONBLOCKING::warn_if_not_native = true;
 bool grb::internal::NONBLOCKING::manual_tile_size = false;
 size_t grb::internal::NONBLOCKING::manual_fixed_tile_size =
 	grb::config::ANALYTIC_MODEL::MIN_TILE_SIZE;
-size_t grb::internal::NONBLOCKING::num_threads = grb::config::OMP::threads();
+size_t grb::internal::NONBLOCKING::num_threads; // = grb::config::OMP::threads();
 
 template<>
 grb::RC grb::init< grb::nonblocking >(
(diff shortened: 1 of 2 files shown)


src/graphblas/nonblocking/init.cpp
```diff
@@ -38,7 +38,7 @@ bool grb::internal::NONBLOCKING::warn_if_not_native = true;
 bool grb::internal::NONBLOCKING::manual_tile_size = false;
 size_t grb::internal::NONBLOCKING::manual_fixed_tile_size =
 	grb::config::ANALYTIC_MODEL::MIN_TILE_SIZE;
-size_t grb::internal::NONBLOCKING::num_threads = grb::config::OMP::threads();
+size_t grb::internal::NONBLOCKING::num_threads; // = grb::config::OMP::threads();
```
anyzelman, line 41: What's happening here btw, was this a circular dependence? (If yes, that's a straight up bug right??)

anyzelman: @ChristosMatzoros

PR #364, opened 2025-06-18
363 missing grb wait for collecting execution times

aristeidis-mastoras:
adding missing `grb::wait` for the correct benchmarking in multiple tests

### DIFF
--- a/include/graphblas/interfaces/pregel.hpp
+++ b/include/graphblas/interfaces/pregel.hpp
@@ -413,6 +413,9 @@ namespace grb {
 					) {
 						throw std::runtime_error( "Could not compute vertex IDs" );
 					}
+					if( grb::wait( IDs, indegrees, outdegrees, ones ) != SUCCESS ) {
+						throw std::runtime_error( "Error during initialisation" );
+					}
 				}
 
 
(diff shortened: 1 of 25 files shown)


anyzelman:
Concept release notes:

This MR adds calls to `grb::wait` for performance tests where its omission may result in false performance results for backends that perform nonblocking execution. This MR also fixes the following meta-bugs:
- the return code of the label propagation algorithm was ignored in the performance benchmark driver,
- same for the label propagation smoke test,
- some initialisation operations were not error-checked in the BiCGstab smoke test,
- same for the CG smoke tests,
- re-enabled `sleep(1)` in the HPCG benchmark to detect system variabilities,
- the Pregel API constructor did not account for nonblocking behaviour, and should wait before returning to adhere to RAII.

This MR also improves, on occasion, error reporting, and, as always, includes code style fixes. Thank you to @aristeidis-mastoras for flagging these issues and providing the initial fixes!

anyzelman:
Code review about 25% done (I thought this was an easy one, but alas...)

anyzelman (approved):

anyzelman:
Full test suite started, with LPF, to make sure all changes here do not break anything. When confirmed they run OK, will merge. Thanks Aristeidis for contributing!

anyzelman:
2467 tests pass, will merge

PR #372, opened 2025-07-25
Fixes the reported issue #371

### LINKED ISSUE #371: Descriptor resolutions on compile time
Code like this one 
```cpp
	rc = rc ? rc : grb::eWiseApply< (descr_dense) >(
		y_comp, static_cast<IOType>(0),
		mask, 
		grb::operators::left_assign_if<IOType, bool, IOType>()
	);
```
will fail to compile due to the static assertions in inactive part of the code 
```bash 
static_assert( !internal::maybe_noop< OP >::value, "Warning: you may be "
```
in this example from 	https://github.com/Algebraic-Programming/ALP/blob/136ac7022a0b4993364413ec7e25dafe8493c027/include/graphblas/reference/blas1.hpp#L5000
...

https://github.com/Algebraic-Programming/ALP/blob/136ac7022a0b4993364413ec7e25dafe8493c027/include/graphblas/reference/blas1.hpp#L5054-L5075

```cpp
		// check for dense variant
		if( (descr & descriptors::dense) ||
			internal::getCoordinates( y ).nonzeroes() == n
		) {
			internal::getCoordinates( z ).assignAll();
			const internal::Coordinates< reference > * const no_coordinates = nullptr;
			 // **<<------ active code**
			return internal::dense_apply_generic< true, false, false, false, descr >(  
				 internal::getRaw( z ), &alpha, no_coordinates,
				 internal::getRaw( y ), no_coordinates, op, n
			);
		}
		
		// **<<------ inactive code**
		// we are in the sparse variant
		internal::getCoordinates( z ).clear();
		const bool * const null_mask = nullptr;
		const Coords * const null_coors = nullptr;
		//**<<------ static assert inside**
		return internal::sparse_apply_generic< false, false, true, false, descr >(    
			internal::getRaw( z ), internal::getCoordinates( z ), null_mask, null_coors,
			&alpha, null_coors,
			internal::getRaw( y ), &( internal::getCoordinates( y ) ),
			op, n
		);
```

In order to avoid these problems the blocks of the code should be properly guarded by static checks.
related to  #370 where some statements could be better implemented with this examples

anyzelman:
Closes #371

### DIFF
--- a/include/graphblas/ops.hpp
+++ b/include/graphblas/ops.hpp
@@ -76,7 +76,8 @@ namespace grb {
 		 *
 		 * \warning Therefore, this operator may propagate the use of uninitialised
 		 *          values if not used with care. Ensuring its use with in-place
-		 *          primitives is recommended.
+		 *          primitives is recommended; it should \em not be used with
+		 *          #grb::eWiseApply.
 		 */
 		template<
 			typename D1, typename D2 = D1, typename D3 = D2,
@@ -132,7 +133,8 @@ namespace grb {
 		 *
 		 * \warning Therefore, this operator may propagate the use of uninitialised
 		 *          values if not used with care. Ensuring its use with in-place
-		 *          primitives is recommended.
+		 *          primitives is recommended; it should \em not be used with
+		 *          #grb::eWiseApply.
 		 */
 		template<
 			typename D1, typename D2 = D1, typename D3 = D2,
--- a/include/graphblas/reference/blas1.hpp
+++ b/include/graphblas/reference/blas1.hpp
@@ -3132,9 +3132,11 @@ namespace grb {
 			(void) n;
 #endif
 #ifndef GRB_NO_NOOP_CHECKS
-			static_assert( !internal::maybe_noop< OP >::value, "Warning: you may be "
-				"generating an output vector with uninitialised values. Define "
-				"the GRB_NO_NOOP_CHECKS macro to disable this check.\n" );
+			static_assert(
+				(descr & descriptors::dense) || !internal::maybe_noop< OP >::value,
+				"Warning: you may be generating an output vector with uninitialised values."
+				" (Define the GRB_NO_NOOP_CHECKS macro to disable this check.)\n"
+			);
 #endif
 			// assertions
 			assert( !masked || mask_coors != nullptr );
--- a/include/graphblas/reference/blas1.hpp
+++ b/include/graphblas/reference/blas1.hpp
@@ -2975,6 +2975,13 @@ namespace grb {
 #ifdef _DEBUG
 			std::cout << "\t internal::dense_apply_generic called\n";
 #endif
+#ifndef GRB_NO_NOOP_CHECKS
+			static_assert(
+				!internal::maybe_noop< OP >::value,
+				"Warning: you may be generating an output vector with uninitialised values."
+				" (Define the GRB_NO_NOOP_CHECKS macro to disable this check.)\n"
+			);
+#endif
 			static_assert( !(left_scalar && left_sparse),
 				"The left-hand side must be scalar OR sparse, but cannot be both!" );
 			static_assert( !(right_scalar && right_sparse),


anyzelman:
Currently proposed C++11-style fix running CI. Suggest to run full unit test suite prior to merge. Issue is resolved as by the following manual test:

```
$ /tmp/grbinstall/bin/grbcxx test.cpp 
$ cat test.cpp 
#include <graphblas.hpp>

int main() {
	constexpr size_t n = 100;
	grb::Vector< double > mask( n ), y_comp( n );
	grb::RC ret = grb::set( y_comp, (double)3.14 );
	ret = ret ? ret : grb::set( mask, (double)1 );
	ret = ret ? ret : grb::eWiseApply< grb::descriptors::dense >(
		y_comp, mask, (double)0, grb::operators::right_assign< double >() );
	return ret;
}

$ ./a.out 
$ echo $?
0
```

anyzelman:
Did a quick scan of the same issue in other places -- no such case. If CI and tests run, ready for merge.

anyzelman:
@djelovina could you confirm this addresses your issue? I applied a C++11 style fix, the bumping to C++17 and use of `if constexpr` which would have prevented this issue from appearing also seems a lot of work and better first considered more deeply. Opened #373 to not forget and track it. Some more thoughts are tersely added there, and welcome to add more of course=)

anyzelman:
CI OK. Unit tests (with LPF) are running

djelovina (dismissed):

anyzelman:
All unit tests OK; will merge

anyzelman:
Actually no, I think the change in this MR is correct, but the code example, which was wrong in the above but now fixed (see below), should in fact trigger that assertion failure(!)

```
$ cat test.cpp 
#include <graphblas.hpp>

int main() {
	constexpr size_t n = 100;
	grb::Vector< double > mask( n ), y_comp( n );
	grb::RC ret = grb::set( y_comp, (double)3.14 );
	ret = ret ? ret : grb::set( mask, (double)1 );
	ret = ret ? ret : grb::eWiseApply< grb::descriptors::dense >(
		y_comp, (double)0, mask, grb::operators::left_assign_if< double >() );
	return ret;
}
$ /tmp/grbinstall/bin/grbcxx test.cpp
$ ./a.out 
$ echo $?
0
```
The reason this code should fail, is because even if `y_comp` is dense on input, under the combined semantics of that of `eWiseApply` (which semantically clears the output vector and then applies an operator on two dense inputs, thus having to produce a dense output also) and the `left_assign_if` operator (which may produce nothing on output when the if-value evaluates false), the call should indeed generate a dense output but the output element values for those entries where the mask evaluates `false`, are indeed undefined. It now indeed retain the old values -- but that's actually undefined behaviour. So in fact a static_assert is missing in the dense code path(!!!)

anyzelman:
Now the behaviour is correct, same `test.cpp` as above:

```
$ /tmp/grbinstall/bin/grbcxx test.cpp 
In file included from /tmp/grbinstall/include/graphblas/blas1.hpp:29,
                 from /tmp/grbinstall/include/graphblas.hpp:538,
                 from test.cpp:1:
/tmp/grbinstall/include/graphblas/reference/blas1.hpp: In instantiation of ‘grb::RC grb::internal::dense_apply_generic(OutputType*, const InputType1*, const grb::internal::Coordinates<grb::reference>*, const InputType2*, const grb::internal::Coordinates<grb::reference>*, const OP&, size_t) [with bool left_scalar = true; bool right_scalar = false; bool left_sparse = false; bool right_sparse = false; unsigned int descr = 16; OP = grb::operators::left_assign_if<double>; OutputType = double; InputType1 = double; InputType2 = double; size_t = long unsigned int]’:
/tmp/grbinstall/include/graphblas/reference/blas1.hpp:5069:76:   required from ‘grb::RC grb::eWiseApply(grb::Vector<DataType, grb::reference, Coords>&, InputType1, const grb::Vector<InputType, grb::reference, Coords>&, const OP&, const grb::Phase&, const typename std::enable_if<((((! grb::is_object<InputType1>::value) && (! grb::is_object<InputType2>::value)) && (! grb::is_object<OutputType>::value)) && grb::is_operator<OP>::value), void>::type*) [with unsigned int descr = 16; OP = grb::operators::left_assign_if<double>; OutputType = double; InputType1 = double; InputType2 = double; Coords = grb::internal::Coordinates<grb::reference>; typename std::enable_if<((((! grb::is_object<InputType1>::value) && (! grb::is_object<InputType2>::value)) && (! grb::is_object<OutputType>::value)) && grb::is_operator<OP>::value), void>::type = void]’
test.cpp:8:62:   required from here
/tmp/grbinstall/include/graphblas/reference/blas1.hpp:2980:33: error: static assertion failed: Warning: you may be generating an output vector with uninitialised values. (Define the GRB_NO_NOOP_CHECKS macro to disable this check.)

 2980 |                                 !internal::maybe_noop< OP >::value,
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/grbinstall/include/graphblas/reference/blas1.hpp:2980:33: note: ‘!(bool)grb::internal::maybe_noop<grb::operators::left_assign_if<double> >::value’ evaluates to false
```

I also added a warning about this to the operators for which this might happen -- actually a warning was already there, but I just made it more explicit with this MR.

I am re-running the CI and unit tests. Meanwhile, @djelovina could you double-check what I wrote above here is correct and then also review the other changes? Many thanks!

djelovina (approved):

PR #379, opened 2025-10-07
345 Compilation warnings suppression

### LINKED ISSUE #345: Subdue some compiler warnings when compiling in debug mode
List of known warnings:
```
graphblas/src/transition/fuselets.cpp:341:30: warning: comparison of integer expressions of different signedness
graphblas/src/transition/fuselets.cpp:342:20: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/utils/unordered_memmove.hpp:306:39: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/utils/unordered_memmove.hpp:310:43: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/utils/unordered_memmove.hpp:340:23: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/utils/unordered_memmove.hpp:341:21: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/utils/unordered_memmove.hpp:353:20: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/utils/unordered_memmove.hpp:354:20: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/reference/blas3.hpp:1036:25: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/reference/blas3.hpp:1037:11: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/reference/blas3.hpp:1082:25: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/reference/blas3.hpp:1083:11: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/reference/blas3.hpp:1118:25: warning: comparison of integer expressions of different signedness
graphblas/include/graphblas/reference/blas3.hpp:1119:11: warning: comparison of integer expressions of different signedness
```

Some new warnings that should likely be suppressed are the following:
```
graphblas/include/graphblas/reference/blas1.hpp:3744:9: warning: 'z_b[7]' may be used uninitialized in this function [-Wmaybe-uninitialized]
 3744 |         *( z_p + index ) = z_b[ k ];
      |         ^
```

GiovaGa:
The warnings are no more (as far as I can tell)

Resolves #345

### DIFF
--- a/include/graphblas/reference/blas1.hpp
+++ b/include/graphblas/reference/blas1.hpp
@@ -3751,9 +3751,7 @@ namespace grb {
 								(void) z_coors.assign( index );
 #endif
 								GRB_UTIL_IGNORE_MAYBE_UNINITIALIZED
-								// z_b[ k ] has been initialized in te loop just before
-								// Some compilers do not see it, others do.
-								// See issue #345 and related PR
+								// z_b[ k ] has been initialized in the loop just before
 								*( z_p + index ) = z_b[ k ];
 								GRB_UTIL_RESTORE_WARNINGS
 							}
--- a/include/graphblas/reference/blas1.hpp
+++ b/include/graphblas/reference/blas1.hpp
@@ -3735,14 +3735,10 @@ namespace grb {
 							}
 						}
 						for( size_t k = 0; k < block_size; ++k ) {
-							if( mask_b[ k ] ) {
-								apply( z_b[ k ], x_b[ k ], y_b[ k ], op );
-							}
-						}
-						for( size_t k = 0; k < block_size; ++k ) {
 							const size_t index = i + k;
 							assert( index < n );
 							if( mask_b[ k ] ) {
+								apply( z_b[ k ], x_b[ k ], y_b[ k ], op );
 #ifdef _H_GRB_REFERENCE_OMP_BLAS1
 								if( !z_coors.asyncAssign( index, update ) ) {
 									(void) ++asyncAssigns;
--- a/include/graphblas/reference/blas1.hpp
+++ b/include/graphblas/reference/blas1.hpp
@@ -3735,10 +3735,14 @@ namespace grb {
 							}
 						}
 						for( size_t k = 0; k < block_size; ++k ) {
+							if( mask_b[ k ] ) {
+								apply( z_b[ k ], x_b[ k ], y_b[ k ], op );
+							}
+						}
+						for( size_t k = 0; k < block_size; ++k ) {
 							const size_t index = i + k;
 							assert( index < n );
 							if( mask_b[ k ] ) {
-								apply( z_b[ k ], x_b[ k ], y_b[ k ], op );
 #ifdef _H_GRB_REFERENCE_OMP_BLAS1
 								if( !z_coors.asyncAssign( index, update ) ) {
 									(void) ++asyncAssigns;
--- a/include/graphblas/reference/blas1.hpp
+++ b/include/graphblas/reference/blas1.hpp
@@ -3750,7 +3750,12 @@ namespace grb {
 #else
 								(void) z_coors.assign( index );
 #endif
+								GRB_UTIL_IGNORE_MAYBE_UNINITIALIZED
+								// z_b[ k ] has been initialized in te loop just before
+								// Some compilers do not see it, others do.
+								// See issue #345 and related PR
 								*( z_p + index ) = z_b[ k ];
+								GRB_UTIL_RESTORE_WARNINGS
 							}
 						}
 #ifdef _H_GRB_REFERENCE_OMP_BLAS1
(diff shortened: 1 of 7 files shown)


GiovaGa:
Some warnings appear when compiling unit tests. I will fix also those before review

anyzelman:
Also be aware that to truly compile everything, LPF should be installed and the bootstrap should point to the LPF installation. Doing so enables the `bsp1d` and `hybrid` backends, and all tests that depend on them (both generic as well as some extra specific ones).

To build all tests (also smoke and performance, in addition to unit), the target `build_tests_all` is your friend=)

anyzelman:
I also saw that I forgot to note the compiler I got those warnings with...not very nice of myself hahah. I think for this MR, we should at least note which compilers we use in our checking, and in any case the warnings collected in the issue should be fixed anyways? (They all seem to be valid issues)

src/transition/fuselets.cpp
```diff
@@ -338,8 +338,8 @@ static int update_spmv_dot(
 	// compiled in debug mode:
 	for( size_t i = 0; i < n; ++i ) {
 		assert( ia[ i + 1 ] >= ia[ i ] );
-		for( size_t k = ia[ i ]; k < ia[ i + 1 ]; ++k ) {
-			assert( ij[ k ] < n );
+		for( OffsetT k = ia[ i ]; k < ia[ i + 1 ]; ++k ) {
+			assert( ij[ k ] <  static_cast<OffsetT>( n ) );
```
anyzelman, line 343: Should there be an assertion before the for loops to check that n does not exceed `std::numeric_limits< OffsetT >::max()` ?

anyzelman (changes requested):
Just one minor comment thus far, all other fixes are perfect=)

anyzelman:
Finally, before I forget, could you add yourself to the NOTICE file as part of this MR, assuming it'll be your first contrib?

src/transition/fuselets.cpp
GiovaGa: Yes, I must agree that such a check is probably better to have. Will add

GiovaGa:
Some warnings still appear, but shouldn't (it seems to be 2. in [Suppressions.md](https://github.com/Algebraic-Programming/ALP/blob/develop/docs/Suppressions.md)).

For future reference: I am compiling with g++ 11.4.0

GiovaGa:
Those warnings are not shown with gcc 9.4.0
I really have no idea what's happening :disappointed:

anyzelman:
> Those warnings are not shown with gcc 9.4.0 I really have no idea what's happening 😞

This is "normal", by the way: different compiler versions may trigger different warnings, while different issues may be left unflagged. LLVM / Clang seems a bit more stable in emitting warnings than GCC.

Quick question: OK to merge this after the ID test one right?

include/graphblas/reference/blas1.hpp
```diff
@@ -3734,15 +3734,11 @@ namespace grb {
...
-						for( size_t k = 0; k < block_size; ++k ) {
-							if( mask_b[ k ] ) {
-								apply( z_b[ k ], x_b[ k ], y_b[ k ], op );
-							}
-						}
 						for( size_t k = 0; k < block_size; ++k ) {
 							const size_t index = i + k;
 							assert( index < n );
 							if( mask_b[ k ] ) {
+								apply( z_b[ k ], x_b[ k ], y_b[ k ], op );
```
anyzelman: Could you revert this change? The rationale of the original code is that lines 3737-3741 corresponds to a masked vectorised addition instruction (if op, just as an example, addition). This works because `block_size` is a `constexpr size_t`, which both GCC and LLVM are OK with substituting to SIMD instructions. If instead interleaving the `apply` with a bunch of other code, I'm quite certain compilers won't vectorise it properly-- and even if some do, I'd be worried other compilers or other compiler versions will not. In other words, the current coding style "tricks" compilers into vectorising, while doing it in this way furthermore is transportable between architectures (diff. AVX versions, NEON, SVE are all tested to work with this trick)

anyzelman (changes requested):
All other changes look OK to me, just one issue

include/graphblas/reference/blas1.hpp
GiovaGa: I see. The change suppresses a `maybe-uninitialized` warning, so I guess we'll have to live with this warning.
Should I just suppress it? it is clear that the warning is not a true danger of any kind.

anyzelman:
Yes, please suppress it instead (and note it in the related markdown document:)

Many thanks!!

anyzelman:
Rebased and letting the CI do its thing first

anyzelman (dismissed):

anyzelman:
I piggy-backed on to this MR a hotfix for a metabug that somehow got triggered by the previous commit to develop, re-testing before merge

anyzelman (approved):

PR #382, opened 2025-10-07
168 Added copy-assignment vector ID unit tests

### LINKED ISSUE #168: Assignment-operator from temporaries overwrites ID
Consider the following, in which the final assertion is assumed to hold:

```
grb::Vector< T > a( n ), b( n );
size_t a_id = grb::getID( a );
// ...
a = b;
assert( grb::getID( a ) == a );
```

Two items:
1. this is currently guaranteed in all backends for copy-assignment, but is never tested;
2. this seems currently *not* guaranteed for move-assignment.

This issue is to introduce such a test, and confirm whether the latter issue indeed exists (and if yes, fix it in some good way)

GiovaGa:
Move assignment id tests were already present from commit 1cdf5833f .

Resolves #168

### DIFF
--- a/tests/unit/id.cpp
+++ b/tests/unit/id.cpp
@@ -124,7 +124,7 @@ void grb_program1( const struct input &in, struct output &out ) {
 
 
 	/**
-	 * Test for copy assignement id
+	 * Test for copy assignment id
 	 *
 	 * Creating and performing copy assignment on multiple new objects and check
 	 * for collisions.
@@ -146,18 +146,27 @@ void grb_program1( const struct input &in, struct output &out ) {
 			rc = grb::FAILED;
 			return;
 		}
-		if( (grb::getID( new_two ) != old_two_id)
-         || (grb::getID( new_one ) == grb::getID( new_two ) ) ) {
+		if( grb::getID( new_one ) != old_one_id ) {
+			std::cerr << "\t detected ID change in source object after copty\n";
+			rc = grb::FAILED;
+			return;
+		}
+		if( grb::getID( new_two ) != old_two_id) {
 			std::cerr << "\t detected new ID was different from original after assignment\n";
+			rc = grb::FAILED;
+			return;
+		}
+        if( grb::getID( new_one ) == grb::getID( new_two ) ) {
+			std::cerr << "\t detected duplicated ID after copy-assignment\n";
 			rc = grb::FAILED;
 			return;
 		}
 	}
 
 	/**
-	 * Test for move assignement id
+	 * Test for move assignment id
 	 *
-	 * Creating and performing copy assignment on multiple new objects and check
+	 * Creating and performing move assignment on multiple new objects and check
 	 * for collisions.
 	 */
 	for( int i = 0; i < 100; i++ ) {
@@ -172,14 +181,14 @@ void grb_program1( const struct input &in, struct output &out ) {
 		const size_t old_two_id = grb::getID( new_two );
 		new_two = std::move( new_one );
 
-		if( grb::getID( new_two ) == old_one_id ) {
-			std::cerr << "\t detected duplicated ID transferred with copy-assignment\n";
+		if( grb::getID( new_two ) == old_two_id ) {
+			std::cerr << "\t detected old ID was retained after a call to std::move\n";
 			rc = grb::FAILED;
 			return;
 		}
-		if( (grb::getID( new_two ) != old_two_id)
-         || (grb::getID( new_one ) == grb::getID( new_two ) ) ) {
-			std::cerr << "\t detected new ID was different from original after assignment\n";
+		if( grb::getID( new_two ) != old_one_id ) {
+			std::cerr << "\t detected a new ID after std::move, expected the moved-from "
+				<< "ID to have been retained\n";
 			rc = grb::FAILED;
 			return;
 		}
@@ -278,7 +287,7 @@ void grb_program2( const struct input &in, struct output &out ) {
 	}
 
 	/**
-	 * Test for move assignement id cleanup.
+	 * Test for move assignment id cleanup.
 	 *
 	 * Creating and performing move assignment on multiple new objects and check
 	 * for collisions.
--- a/tests/unit/id.cpp
+++ b/tests/unit/id.cpp
@@ -122,6 +122,37 @@ void grb_program1( const struct input &in, struct output &out ) {
 		return;
 	}
 
+
+	/**
+	 * Test for copy assignement id
+	 *
+	 * Creating and performing copy assignment on multiple new objects and check
+	 * for collisions.
+	 */
+	for( int i = 0; i < 100; i++ ) {
+		grb::Vector< int > new_one( 10 );
+		grb::Vector< int > new_two( 10 );
+		if( grb::getID( new_one ) == grb::getID( new_two ) ) {
+			std::cerr << "\t two calls to getID on two new containers return same ID\n";
+			rc = grb::FAILED;
+			return;
+		}
+		const size_t old_one_id = grb::getID( new_one );
+		const size_t old_two_id = grb::getID( new_two );
+		new_two = new_one;
+
+		if( grb::getID( new_two ) == old_one_id ) {
+			std::cerr << "\t detected duplicated ID transferred with copy-assignment\n";
+			rc = grb::FAILED;
+			return;
+		}
+		if( (grb::getID( new_two ) != old_two_id)
+         || (grb::getID( new_one ) == grb::getID( new_two ) ) ) {
+			std::cerr << "\t detected new ID was different from original after assignment\n";
+			rc = grb::FAILED;
+			return;
+		}
+	}
 }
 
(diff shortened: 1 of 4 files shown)


tests/unit/id.cpp
```diff
@@ -122,6 +122,68 @@ void grb_program1( const struct input &in, struct output &out ) {
 		return;
 	}
 
+
```
anyzelman, line 125: superfluous empty line

tests/unit/id.cpp
```diff
@@ -122,6 +122,68 @@ void grb_program1( const struct input &in, struct output &out ) {
...
+	/**
+	 * Test for copy assignement id
```
anyzelman: typo (assignement->assignment) -- maybe also elsewhere in the file ?

tests/unit/id.cpp
```diff
@@ -122,6 +122,68 @@ void grb_program1( const struct input &in, struct output &out ) {
...
+		if( (grb::getID( new_two ) != old_two_id)
+         || (grb::getID( new_one ) == grb::getID( new_two ) ) ) {
+			std::cerr << "\t detected new ID was different from original after assignment\n";
+			rc = grb::FAILED;
+			return;
+		}
+	}
+
+	/**
+	 * Test for move assignement id
```
anyzelman: yes, here the typo also appears (i.e., looks like it was already there)

tests/unit/id.cpp
```diff
@@ -122,6 +122,68 @@ void grb_program1( const struct input &in, struct output &out ) {
...
+		const size_t old_two_id = grb::getID( new_two );
+		new_two = new_one;
+
+		if( grb::getID( new_two ) == old_one_id ) {
+			std::cerr << "\t detected duplicated ID transferred with copy-assignment\n";
+			rc = grb::FAILED;
+			return;
+		}
+		if( (grb::getID( new_two ) != old_two_id)
+         || (grb::getID( new_one ) == grb::getID( new_two ) ) ) {
```
anyzelman: This or condition detects a separate issue right, namely that two different containers must never have the same ID? (This makes the error message on the next line incorrect, i.e., better to put this as a separate if-statement with its own error message)

tests/unit/id.cpp
```diff
@@ -122,6 +122,68 @@ void grb_program1( const struct input &in, struct output &out ) {
...
+		const size_t old_one_id = grb::getID( new_one );
+		const size_t old_two_id = grb::getID( new_two );
+		new_two = new_one;
+
+		if( grb::getID( new_two ) == old_one_id ) {
+			std::cerr << "\t detected duplicated ID transferred with copy-assignment\n";
+			rc = grb::FAILED;
+			return;
+		}
+		if( (grb::getID( new_two ) != old_two_id)
```
anyzelman: maybe for good measure, the ID of new_one should also be tested to remain the same?

tests/unit/id.cpp
anyzelman: (ie., the same typo is also elsewhere, could you help find/replace all of them?)

tests/unit/id.cpp
```diff
@@ -122,6 +122,68 @@ void grb_program1( const struct input &in, struct output &out ) {
...
+		if( grb::getID( new_one ) == grb::getID( new_two ) ) {
+			std::cerr << "\t two calls to getID on two new containers return same ID\n";
+			rc = grb::FAILED;
+			return;
+		}
+		const size_t old_one_id = grb::getID( new_one );
+		const size_t old_two_id = grb::getID( new_two );
+		new_two = std::move( new_one );
+
+		if( grb::getID( new_two ) == old_one_id ) {
```
anyzelman: I think `old_one_id` should be `old_two_id`. Rationale: the `std::move` is to pass all resources underlying `new_two` to `new_one`. Thus both practically and semantically, `new_two` "becomes" `new_one` (while `new_one` enters some undefined unusable state). Therefore, if `new_two` still has the same ID as before, then something went wrong -- while the expectation in fact is that `new_two` has the former ID of `new_one`. This also matches the logic for the test later on in this file, for matrices rather than vectors.

tests/unit/id.cpp
```diff
@@ -122,6 +122,68 @@ void grb_program1( const struct input &in, struct output &out ) {
...
+			std::cerr << "\t detected duplicated ID transferred with copy-assignment\n";
+			rc = grb::FAILED;
+			return;
+		}
+		if( (grb::getID( new_two ) != old_two_id)
+         || (grb::getID( new_one ) == grb::getID( new_two ) ) ) {
```
anyzelman: this clause is invalid (we can't refer to `new_one` after having moved away its resources)

tests/unit/id.cpp
```diff
@@ -122,6 +122,68 @@ void grb_program1( const struct input &in, struct output &out ) {
...
+		const size_t old_one_id = grb::getID( new_one );
+		const size_t old_two_id = grb::getID( new_two );
+		new_two = std::move( new_one );
+
+		if( grb::getID( new_two ) == old_one_id ) {
+			std::cerr << "\t detected duplicated ID transferred with copy-assignment\n";
+			rc = grb::FAILED;
+			return;
+		}
+		if( (grb::getID( new_two ) != old_two_id)
```
anyzelman: The latter should be `old_one_id`, in line with the preceding comment

anyzelman (changes requested):

tests/unit/id.cpp
GiovaGa: This was most definitely reckless copy-paste on my side

GiovaGa:
The issues highlighted should be resolved with these changes

anyzelman (dismissed):

anyzelman:
Code review complete -- lemme run the unit test suite to make sure all is OK before merging:)

anyzelman:
I think actually this change request is working perfectly -- it found a bug. Tracing manually, we have for `id_debug_reference` the following issue:

```
#10 0x000000000040307f in grb_program1 (in=..., out=...) at /home/yzelman/Documents/alp/tests/unit/id.cpp:173
173			grb::Vector< int > new_one( 10 );
```
creates a new vector where `assigned` has the following pointer:
```
#7  0x000000000040a17e in grb::Vector<int, (grb::Backend)0, grb::internal::Coordinates<(grb::Backend)0> >::initialize (this=0x7fffffffd4f0, id_in=0x0, raw_in=0x0, assigned_in=0x0, assigned_initialized=false, buffer_in=0x0, 
    cap_in=10, nz=10) at /home/yzelman/Documents/alp/include/graphblas/reference/vector.hpp:433
433					_id = internal::reference_mapper.insert(
(gdb) list
428				}
429	
430				// assign _id
431				assert( assigned != nullptr );
432				if( id_in == nullptr ) {
433					_id = internal::reference_mapper.insert(
434						reinterpret_cast< uintptr_t >( assigned )
435					);
436				}
437	
(gdb) print assigned
$3 = 0x444dc0 "D\004"
```
The following assertion, given that pointer, however, fails:
```
#6  0x00000000004096ac in grb::utils::DMapper<unsigned long>::insert (this=0x4264e0 <grb::internal::reference_mapper>, in=4476352) at /home/yzelman/Documents/alp/include/graphblas/utils/DMapper.hpp:97
97						assert( it == mapper.end() );
```
I.e., the pointer was already registered, thus indicating that the old pointer was not removed correctly from the mapper. This example uses `std::move`, so this (move assignment) is the likely culprit to have a look at.

This btw happens at the thirtieth time we're running this, which appears to be when Linux decided to give us the same pointer back after having previously freed it:
```
#10 0x000000000040307f in grb_program1 (in=..., out=...) at /home/yzelman/Documents/alp/tests/unit/id.cpp:173
173			grb::Vector< int > new_one( 10 );
(gdb) list
168		 *
169		 * Creating and performing move assignment on multiple new objects and check
170		 * for collisions.
171		 */
172		for( int i = 0; i < 100; i++ ) {
173			grb::Vector< int > new_one( 10 );
174			grb::Vector< int > new_two( 10 );
175			if( grb::getID( new_one ) == grb::getID( new_two ) ) {
176				std::cerr << "\t two calls to getID on two new containers return same ID\n";
177				rc = grb::FAILED;
(gdb) print i
$1 = 30
```

GiovaGa:
It seems that the issue is that with `std::move` the destructor of the recieving object is never called (thus the reference mapper never deletes its address). I will now try to solve this

anyzelman:
Perfect, sounds good!

anyzelman:
Thanks @GiovaGa ! I checked the fix, and it's good - I also had to apply a similar fix to the BSP1D backend (implicitly also the hybrid backend), and I did some minor touch-ups on code style-- you can check out the last three commits for details:)

anyzelman (approved):
Just waiting for a final check of the unit tests with LPF, as well as on the internal CI. If both pass, will merge=) Thanks @GiovaGa for the contribution!!

anyzelman:
All tests OK, will merge

PR #383, opened 2025-10-09
255 refactor `alloc.hpp`

### LINKED ISSUE #255: Refactor `alloc.hpp`
The `alloc.hpp` started out as a utility, but currently the code base requires that each backend provides its own allocator functions. Therefore, `alloc.hpp` should be moved from the `utils` directory to the base directory, and follow the same pattern as e.g. the collectives.

GiovaGa:
Resolves #255

Compiles and unit tests pass.


include/graphblas/alloc.hpp
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+
+/*
+ * @author A. N. Yzelman
+ * @date 13th of September, 2017
+ */
+
+#ifndef _H_GRB_UTILS_ALLOC
+#define _H_GRB_UTILS_ALLOC
```
anyzelman: `UTILS_` (or `_UTILS`:) should be removed here

include/graphblas/alloc.hpp
```diff
@@ -0,0 +1,66 @@
...
+namespace grb {
+	template< enum Backend implementation = config::default_backend >
+	class AllocatorFunctions ;
+
+	template< enum Backend implementation = config::default_backend >
+	class Allocator;
+}
+#endif
+
+#endif
```
anyzelman: maybe append
```
 // end `_H_GRB_ALLOC`
```
after the endif on line 65

include/graphblas/utils/alloc.hpp
```diff
@@ -1,160 +0,0 @@
...
-
-	namespace utils {
-
-		/**
-		 * Allocates a single memory area. This is the trivial version of the other
-		 * #grb::utils::alloc function which is designed to allocate multiple memory
-		 * areas in one go; please see that function for extended documentation.
-		 *
-		 * @see grb::utils::alloc.
-		 */
```
anyzelman, line 80: I see dead code was being documented here...apologies for that. Did you double-check if this type of comment is now in `base/alloc.hpp` perhaps?

include/graphblas/utils/alloc.hpp
```diff
@@ -1,160 +0,0 @@
...
-		 *
-		 * @returns SUCCESS    When the function has completed as expected. When any
-		 *                     other code is returned, then the following fields shall
-		 *                     remain untouched: \a pointer, \a deleter, plus all
-		 *                     instances of those two fields in \a args.
-		 * @returns OUTOFMEM   When there was not sufficient memory available to
-		 *                     allocate the requested memory areas.
-		 * @returns PANIC      When any other non-mitigable failure was found during
-		 *                     allocation.
-		 */
```
anyzelman, line 147: same here

anyzelman (changes requested):
Again only minor things -- will slot this after the other two MRs that I commented on just now. Looking good!

include/graphblas/utils/alloc.hpp
GiovaGa: I did not check. I am looking now and this is definitely all in `base/alloc.hpp`

anyzelman:
Rebased and letting the CI do its thing first

anyzelman (dismissed):

anyzelman (approved):

anyzelman:
Code review, smoke tests, and CI are all happy -- will merge

PR #384, opened 2025-10-10
151 benchmarker base does not check for errors in collective calls

### LINKED ISSUE #151: BenchmarkerBase does not check for errors in collective calls
Instead, it ignores any error codes coming from collective calls, while on benchmark failures due to interrupted calls to `sleep` it calls `abort` instead of returning `FAILED`. This issue was, in part, also a case of under-specification, but that part is fixed in issue #6 .

GiovaGa:
This should be just a simple change.
Upstream of the abort return codes seem to be correctly passed. After the condition in the loop means that the function finishes soon, so it should be fine like this.

Resolves #151 

Note: exchanged arguments in the if to avoid waiting unnecessarily if there was an error before then.

### DIFF
--- a/include/graphblas/base/benchmark.hpp
+++ b/include/graphblas/base/benchmark.hpp
@@ -277,12 +277,12 @@ namespace grb {
 #endif
 
 						// pause for next outer loop
-						if( sleep( 1 ) != 0 && ret == grb::SUCCESS ) {
+						if( ret == grb::SUCCESS && sleep( 1 ) != 0 ) {
 #ifndef _GRB_NO_STDIO
 							std::cerr << "Sleep interrupted, assume benchmark is unreliable; "
 								<< "exiting.\n";
 #endif
-							abort();
+							ret = grb::FAILED;
 						}
 					}
 
--- a/include/graphblas/base/benchmark.hpp
+++ b/include/graphblas/base/benchmark.hpp
@@ -283,7 +283,7 @@ namespace grb {
 							std::cerr << "Sleep interrupted, assume benchmark is unreliable; "
 								<< "exiting.\n";
 #endif
-							ret = grb::FAILED;
+							ret = grb::PANIC;
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


include/graphblas/base/benchmark.hpp
```diff
@@ -31,6 +31,7 @@
 #include <cmath>  // for sqrt
 #include <limits>
 #include <vector> // warning: normally should not be used in ALP backends(!)
+#include <unistd.h> // for sleep
```
anyzelman, line 34: Nice find!

include/graphblas/base/benchmark.hpp
```diff
@@ -277,12 +278,12 @@ namespace grb {
...
 
 						// pause for next outer loop
-						if( sleep( 1 ) != 0 && ret == grb::SUCCESS ) {
+						if( ret == grb::SUCCESS && sleep( 1 ) != 0 ) {
 #ifndef _GRB_NO_STDIO
 							std::cerr << "Sleep interrupted, assume benchmark is unreliable; "
 								<< "exiting.\n";
 #endif
-							abort();
+							ret = grb::FAILED;
```
anyzelman: I would suggest to return `PANIC`, since this must be some external effect outside the control of ALP that would cause a sleep to get interrupted

anyzelman (changes requested):
Just a minor change request as well as a question: is it correct that the original issue in fact was already handled? (The collectives being called actually do already have their error codes checked right?)

GiovaGa:
It is quite possible that the issue in its original form was mostly solved previously: it seems that 84212ed71 did most of the work.

I would guess that abort was just forgotten, because everything else seems to have been changed with that commit :smile:

anyzelman:
Rebased

anyzelman (approved):

anyzelman:
@GiovaGa I looked again and I think the logic should be that if there was an error, both the sleep and the printing of the error should not occur. I pushed a minor fix to that just now.

anyzelman:
Aside from CI, will also wait on a full test suite run since it's weekend anyway (i.e., will do a run of the performance tests as well, as the minor change with this MR mostly affects those-- as well as some smoke tests which will also run). These tests are with LPF and should be finished by Monday - if all OK will merge

anyzelman:
Things were still running but we got 2876 passed tests and coverage of all backends, so merging:)

PR #386, opened 2025-10-17
Added const qualifier when copy-assigning grb::Vector with the bsp1d backend 

### LINKED ISSUE #385: Copy-assignment of `grb::Vector` with the bsp1d backend has not const parameter
In file `include/graphblas/bsp1d/vector.hpp` at line 2536, I think the parameter should be `const`. 
``` 
                 /**
		 * Copy-assignment.
		 *
		 * Same performance semantics as #grb::set.
		 *
		 * \warning Errors will be thrown as standard C++ exceptions. Users who rather
		 *          not deal with exceptions are encouraged to use #grb::set directly.
		 *
		 * \internal Dispatches to #grb::set.
		 */
		Vector< D, BSP1D, C > & operator=( const Vector< D, BSP1D, C > &x ) { // <------
			const auto rc = set( *this, x );
			if( rc != SUCCESS ) {
				throw std::runtime_error( "grb::set inside copy-constructor: "
					+ toString( rc )
				);
			}

```
Otherwise copy-assigning a `grb::Vector` with a `const grb::Vector` would throw a cryptic compile error when compiling with the bsp1d backend.
Other backends don't seem to have this issue.

GiovaGa:
Solves #385

### DIFF
--- a/include/graphblas/bsp1d/vector.hpp
+++ b/include/graphblas/bsp1d/vector.hpp
@@ -2536,7 +2536,7 @@ namespace grb {
 		 *
 		 * \internal Dispatches to #grb::set.
 		 */
-		Vector< D, BSP1D, C > & operator=( Vector< D, BSP1D, C > &x ) {
+		Vector< D, BSP1D, C > & operator=( const Vector< D, BSP1D, C > &x ) {
 			const auto rc = set( *this, x );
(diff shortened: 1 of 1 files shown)


anyzelman:
You're right, that's a bug. Will handle after #384

anyzelman (approved):

PR #389, opened 2025-10-21
Replaced occurrences of grb::FAILED with grb::ILLEGAL 

### LINKED ISSUE #374: Some functions in when faced with not enough capacity...
...during execution return `FAILED` instead of `ILLEGAL`.

GiovaGa:
Solves #374

It seems that the issue was only in the documentation. Now `GRB::FAILED` appears only in tests, benchmark and algorithms, as should be the case

### DIFF
--- a/include/graphblas/bsp1d/blas1.hpp
+++ b/include/graphblas/bsp1d/blas1.hpp
@@ -71,7 +71,8 @@ namespace grb {
 		template< bool isgd, typename T >
 		void handle_try_execute(
 			grb::Vector< T > &x,
-			const grb::Phase &phase, grb::RC &ret
+			const grb::Phase &phase,
+			grb::RC &ret
 		) {
 			// handle try and execute
 			if( phase != RESIZE ) {
@@ -442,18 +443,7 @@ namespace grb {
 		}
 
 		// handle try and execute phases
-		if( phase != RESIZE ) {
-			if( ret == SUCCESS ) {
-				internal::setDense( y );
-			} else if( !config::IMPLEMENTATION< BSP1D >::fixedVectorCapacities() &&
-				ret == FAILED
-			) {
-				// handle failed TRY
-				const RC subrc = internal::updateNnz( y );
-				if( subrc != SUCCESS ) { ret = PANIC; }
-				// otherwise, still return FAILED (as required)
-			}
-		}
+		grb::internal::handle_try_execute< true >( y, phase, ret );
 
 		// done
 		return ret;
@@ -1807,22 +1797,7 @@ namespace grb {
 		}
 
 		// catch execute
-		if( phase != RESIZE ) {
-			if( !config::IMPLEMENTATION< BSP1D >::fixedVectorCapacities() && 
-				ret == FAILED
-			) {
-				// handle failed TRY
-				const RC subrc = internal::updateNnz( z );
-				if( subrc != SUCCESS ) { ret = PANIC; }
-				// take care to propagate FAILED error code
-			} else if( ret == SUCCESS ) {
-				if( !(descr & descriptors::dense) ) {
-					ret = internal::updateNnz( z );
-				} else {
-					internal::setDense( z );
-				}
-			}
-		}
+		grb::internal::handle_try_execute< descriptors::dense == (descr & descriptors::dense) >( z, phase, ret );
 
 		// done
 		return ret;
--- a/include/graphblas/bsp1d/blas1.hpp
+++ b/include/graphblas/bsp1d/blas1.hpp
@@ -1798,7 +1798,7 @@ namespace grb {
 
 		// catch execute
 		internal::template handle_try_execute<
-				(descr & descriptors::dense)
+				((descr & descriptors::dense) > 0)
 			>( z, phase, ret );
 
 		// done
--- a/include/graphblas/bsp1d/blas1.hpp
+++ b/include/graphblas/bsp1d/blas1.hpp
@@ -53,6 +53,34 @@
 
 namespace grb {
 
+	namespace internal {
+
+		/*
+		 * Handle try and execute phases return code
+		 */
+		template< typename T >
+		grb::RC handle_try_execute( grb::Vector< T > &x, const grb::Phase phase, const grb::RC ret ){
+			// handle try and execute
+			if( phase != RESIZE ) {
+				if( ret == SUCCESS ) {
+					// in this case, the number of nonzeroes in the output vector may have
+					// changed (recall that the dense case is not handled here)
+					return internal::updateNnz( x );
+				} else if( !config::IMPLEMENTATION< BSP1D >::fixedVectorCapacities() &&
+					ret == FAILED
+				) {
+					// in this case, the full computation has not completed but the contents of
+					// x do contain a subset of results. Therefore, the number of nonzeroes may
+					// have changed, but we need to take care to still propagate FAILED
+					const RC subrc = internal::updateNnz( x );
+					if( subrc != SUCCESS ) { return grb::PANIC; }
+				}
+			}
+			return ret;
+		}
+
+	}
+
 	/**
 	 * \defgroup BLAS1_REF The Level-1 ALP/GraphBLAS routines -- BSP1D backend
 	 *
@@ -684,18 +712,7 @@ namespace grb {
 		}
 
 		// handle try and execute
-		if( phase != RESIZE ) {
-			if( ret == SUCCESS ) {
-				internal::setDense( x );
-			} else if( !config::IMPLEMENTATION< BSP1D >::fixedVectorCapacities() &&
-				ret == FAILED
-			) {
-				// handle failed try
-				const RC subrc = internal::updateNnz( x );
-				if( subrc != SUCCESS ) { ret = PANIC; }
-				// ensure we still propagate the FAILED error code
-			}
-		}
+		ret = grb::internal::handle_try_execute( x, phase, ret );
 
 		// done
 		return ret;
@@ -1019,20 +1036,7 @@ namespace grb {
 		}
 
 		// handle try and execute
-		if( phase != RESIZE ) {
-			if( ret == SUCCESS ) {
-				// x may have a new global number of nonzeroes that needs to be synced
-				// (recall that the dense case is not handled here)
-				ret = internal::updateNnz( x );
-			} else if( ret == FAILED ) {
-				// handle failed TRY
-				// x may contain useful results that are a subset of the requested
-				// computation. Therefore the nnz may have changed, but we should
-				// take care to continue propagate FAILED
-				const RC subrc = internal::updateNnz( x );
-				if( subrc != SUCCESS ) { ret = PANIC; }
-			}
-		}
+		ret = handle_try_execute( x, phase, ret );
 
 		// done
 		return ret;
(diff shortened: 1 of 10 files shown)


include/graphblas/phase.hpp
```diff
@@ -196,7 +196,7 @@ namespace grb {
 		 * \em must complete as specified-- unless #grb::PANIC is returned.
 		 *
 		 * If, nevertheless, capacity was not sufficient then the result of the
-		 * computation is incomplete and the primitive shall return #grb::FAILED.
+		 * computation is incomplete and the primitive shall return #grb::ILLEGAL.
```
anyzelman: For the try, this was actually intentionally FAILED - I'll update now

include/graphblas/bsp1d/blas1.hpp
```diff
@@ -1249,7 +1267,12 @@ namespace grb {
 			return ILLEGAL;
 		}
 		if( capacity( z ) < n && phase == EXECUTE ) {
-			return FAILED;
```
anyzelman, line 1252: This was an erroneous `return FAILED;`

include/graphblas/bsp1d/blas1.hpp
```diff
@@ -1326,8 +1349,13 @@ namespace grb {
 		if( size( mask ) != n ) {
 			return MISMATCH;
 		}
-		if( capacity( z ) < n && phase == EXECUTE ) {
-			return FAILED;
```
anyzelman, line 1330: here too

include/graphblas/bsp1d/blas1.hpp
```diff
@@ -1393,8 +1421,13 @@ namespace grb {
 		if( (descr & descriptors::dense) && nnz( z ) != n ) {
 			return ILLEGAL;
 		}
-		if( capacity( z ) < n && phase == EXECUTE ) {
-			return FAILED;
```
anyzelman, line 1397: same

include/graphblas/bsp1d/blas1.hpp
```diff
@@ -1471,8 +1504,13 @@ namespace grb {
 		if( size( mask ) != n ) {
 			return MISMATCH;
 		}
-		if( capacity( z ) < n && phase == EXECUTE ) {
-			return FAILED;
```
anyzelman, line 1475: same

include/graphblas/bsp1d/blas1.hpp
```diff
@@ -1572,8 +1610,8 @@ namespace grb {
 		} else if( phase == EXECUTE ) {
 			if( ret == SUCCESS ) {
 				internal::setDense( z );
-			} else if( ret == FAILED ) {
-				const RC subrc = internal::updateNnz( z );
```
anyzelman, line 1576: This pattern was a bit more weird -- pretty sure it can only be checking for `ILLEGAL`, since this code block is for the `EXECUTE` phase explicitly. Thus rather than syncing the nnzs, it should be ensuring that all subvectors on all user processes are empty. I fixed it as thus

include/graphblas/bsp1d/blas1.hpp
```diff
@@ -1662,8 +1700,8 @@ namespace grb {
...
```
anyzelman, line 1666: same

include/graphblas/bsp1d/blas1.hpp
```diff
@@ -1770,10 +1808,13 @@ namespace grb {
 
 		// catch execute
 		if( phase != RESIZE ) {
-			assert( phase == EXECUTE );
-			if( ret == FAILED ) {
```
anyzelman, line 1774: This here is the most common pattern though. The code block pertains to either EXECUTE or TRY, so FAILED is a possible valid return code (from the TRY). In those cases, the global number of nonzeroes needs to be synced while the FAILED error code is still returned. This is not necessary if the underlying implementation guarantees that vector outputs always have the same full capacity -- in that case, FAILED would never be returned.

This pattern repeats many times in the code that follows, sometimes with slight variations

include/graphblas/reference/blas3.hpp
```diff
@@ -1700,12 +1700,12 @@ namespace grb {
...
 			if( clear_at_exit ) {
 #ifdef _DEBUG_REFERENCE_BLAS3
-				std::cout << "\t mxm_generic is about to exit with FAILED "
+				std::cout << "\t mxm_generic is about to exit with ILLEGAL "
 					<< "(and will clear the output matrix)\n";
 #endif
 				const RC clear_rc = clear( C );
 				if( clear_rc != SUCCESS ) { return PANIC; }
-				return FAILED;
+				return ILLEGAL;
```
anyzelman, line 1708: This one is another confusion between FAILED and ILLEGAL

include/graphblas/reference/blas3.hpp
```diff
@@ -1888,7 +1888,7 @@ namespace grb {
 					<< "complete the requested computation\n";
 #endif
 				if( clear_rc == SUCCESS ) {
-					return FAILED;
+					return ILLEGAL;
```
anyzelman, line 1891: here also

include/graphblas/reference/blas3.hpp
```diff
@@ -2043,7 +2043,12 @@ namespace grb {
 			// step 4, check nonzero capacity
 			assert( crs_offsets[ nrows ] == ccs_offsets[ ncols ] );
 			if( internal::getNonzeroCapacity( A ) < crs_offsets[ nrows ] ) {
-				return FAILED;
```
anyzelman, line 2046: here too

include/graphblas/reference/blas3.hpp
```diff
@@ -2252,7 +2257,10 @@ namespace grb {
 		if( ncols != grb::ncols( A ) ) {
 			return MISMATCH;
 		}
-
+		if( std::numeric_limits< size_t >::max() / nnz( u ) < nnz( v ) ||
+			std::numeric_limits< size_t >::max() / nnz( v ) < nnz( u ) ) {
+			return ILLEGAL;
+		}
```
anyzelman, line 2263: I added here a guard against overflow (unrelated to this MR though)

include/graphblas/reference/blas3.hpp
```diff
@@ -2267,7 +2275,7 @@ namespace grb {
 			if( clear_rc != SUCCESS ) {
 				return PANIC;
 			} else {
-				return FAILED;
```
anyzelman, line 2270: here too an actual confusion between FAILED and ILLEGAL

include/graphblas/reference/blas3.hpp
```diff
@@ -2835,7 +2843,7 @@ namespace grb {
 					if( clear_rc != SUCCESS ) {
 						return PANIC;
 					} else {
-						return FAILED;
+						return ILLEGAL;
```
anyzelman, line 2846: same

anyzelman:
@GiovaGa I found a couple more instances in the code -- mostly in the distributed-memory backend (`BSP1D`). I'll run some tests now but perhaps before merging, you could have a look at the changes I added and make sure they make sense?

include/graphblas/phase.hpp
anyzelman: @GiovaGa for clarity, the TRY phase attempts to EXECUTE, but if the capacity of the output container during execution is found to be too small, then the computation will prematurely exit while retaining partially correct entries in the output container. That is, if the true output is Y, then the try phase may compute Y', where Y' is a structural subset of the values of Y, while for that subset the nonzero values match. It's a phase that we once specced out but have never properly implemented.

(Curiously, perhaps, the only backend that seems structurally OK with the use of TRY is the BSP1D backend.)

Anyway just added for explanation, if this makes sense to you, feel free to press resolve (same for other comments below). Once all resolved, you're happy with the changes, and the CI / tests are OK, will merge~

anyzelman:
All smoke & unit tests OK, will merge after automatic CI and @GiovaGa 's OK

Concept release notes:

Before this MR, the `mxm` and `zip` level-3 `reference`-engine primitives, as well as some level-1 and level-3 `bsp1d`-engine primitives could return `FAILED` instead of `ILLEGAL` when output containers did not have enough capacity reserved before calling an ALP primitive in `EXECUTE` mode. Additionally, some of the specification still erroneously noted that in such cases `FAILED` rather than `ILLEGAL` should be returned.

Additionally, this MR fixes up some minor issues and bugs that were found by GCC 14.3.1:
 - missing include in the bsp1d `init.hpp` 
 - mismatching deleter to `unique_ptr`  in bsp1d `io.hpp`
 - erroneous move-constructor signature in reference `compressed_storage.hpp`
 - implicit conversion between incompatible types (`bool` and `enum`) in `matrixFileReaderBase.hpp` utility

All above issues are addressed with this MR, which, finally, also include minor code style fixes and improved readability of the various options  `grb::Phase` specifies.

anyzelman (dismissed):

GiovaGa:
It seems that I completely missed all the occurrences of `FAILED` without the namespace. What is the code style suggestion about including the `grb::` namespace in such cases? I saw that sometimes it appears and sometimes it does not

GiovaGa (commented):
Looked around and everything seems fine, only thing: maybe it makes sense to put some repeating error logic handling in a separate function to avoid repetitions of the same code over and over

include/graphblas/bsp1d/blas1.hpp
GiovaGa: This makes sense. Maybe it is better to put this logic in a separate function to avoid repetitions?

GiovaGa:
Another thing, @anyzelman maybe it makes sense to also merge #386 just before or just after this one

anyzelman:
I think that would be great yes =) (same for the other similar remark).

From:Giovanni Gaio ***@***.******@***.***>>
To:Algebraic-Programming/ALP ***@***.******@***.***>>
Cc:Albert Jan YZELMAN ***@***.******@***.***>>;Review requested ***@***.******@***.***>>
Date:2025-11-04 09:42:51
Subject:Re: [Algebraic-Programming/ALP] Replaced occurrences of grb::FAILED with grb::ILLEGAL (PR #389)


@GiovaGa commented on this pull request.

________________________________

In include/graphblas/bsp1d/blas1.hpp<https://github.com/Algebraic-Programming/ALP/pull/389#discussion_r2489276191>:

> @@ -1770,10 +1808,13 @@ namespace grb {               // catch execute                if( phase != RESIZE ) {-                        assert( phase == EXECUTE );-                    if( ret == FAILED ) {

This makes sense. Maybe it is better to put this logic in a separate function to avoid repetitions?

anyzelman:
I like to be explicit about namespaces and avoid "using namespaces" whenever possible, but I don't think I've been enforcing that. Searching for FAILED should work (that's what I grepped for:)

From:Giovanni Gaio ***@***.******@***.***>>
To:Algebraic-Programming/ALP ***@***.******@***.***>>
Cc:Albert Jan YZELMAN ***@***.******@***.***>>;Review requested ***@***.******@***.***>>
Date:2025-11-04 09:31:06
Subject:Re: [Algebraic-Programming/ALP] Replaced occurrences of grb::FAILED with grb::ILLEGAL (PR #389)

[https://avatars.githubusercontent.com/u/48856010?s=20&v=4]GiovaGa left a comment(Algebraic-Programming/ALP#389)<https://github.com/Algebraic-Programming/ALP/pull/389#issuecomment-3484556648>

It seems that I completely missed all the occurrences of FAILED without the namespace. What is the code style suggestion about including thegrb:: namespace in such cases? I saw that sometimes it appears and sometimes it does not

anyzelman:
Great @GiovaGa - just one minor issue is that in (I think) two cases, the output vector is guaranteed dense (if the call was successful). This means that instead of calling updateNnz, a call to setDense is more effective as that avoids calling a collective. I did a minor edit of your helper function to reflect that. Please have a look -- I'll try to review more in depth and test the MR again later today. (Also feel free to resolve some of the discussion items above that are still open -- that would indicate you had a look and agree-- or, of course, feel free to raise it if it looks like I made a mistake in some of those=)

GiovaGa:
With your change the function applies in two more places that last time I left out because they were a bit different (I apparently missed some). It looks better now!

anyzelman (approved):

anyzelman:
Rebased, and code review OK. CI OK. Running unit tests and smoke tests with LPF now -- if all ok, will merge

PR #378, opened 2025-09-29
377 simulated annealing replica exchange

djelovina:
Issue #377

PR #398, opened 2025-10-30
Resolving bug in `grb::dot` with nonblocking backend

### LINKED ISSUE #397: Different behavior of dot with nonblocking backend vs reference
The following code snippet:
```
grb::Vector< int > x( n ), y( n );

for(size_t i = n; i > 0 ; --i ){
	grb::setElement( y, i-1, i-1 );
}	

grb::setElement( x, 2, 1 );
grb::setElement( x, 1, 4 );

grb::wait();
int res = 0;
for(size_t i = 0; i < 3; ++i ){
	grb::dot( res, x, y, ring );
	grb::wait();
	std::cout << res << std::endl;
}
grb::wait();
```
There is an abundance of wait to hopefully avoid unexpected effects of the lazy evaluation.

Compiling and running with the reference backend we get, as expected:
```
6
12
18
```
With the nonblocking backend I get:
```
2
4
6
```
Is this a bug or am I missing something?

GiovaGa:
For now the branch contains just a test that reproduces the bug

Solves #397

### DIFF
--- a/include/graphblas/nonblocking/blas1.hpp
+++ b/include/graphblas/nonblocking/blas1.hpp
@@ -10398,6 +10398,9 @@ namespace grb {
 	// internal namespace for implementation of grb::dot
 	namespace internal {
 
+		/**
+		 * Invariant: x should have fewer nonzeroes than y
+		 */
 		template<
 			Descriptor descr,
 #ifdef GRB_BOOLEAN_DISPATCHER
@@ -10432,6 +10435,7 @@ namespace grb {
 #else
 			(void) upper_bound;
 #endif
+			assert( local_x.nonzeroes() <= local_y.nonzeroes() );
 
 			// get raw alias
 			const InputType1 * __restrict__ a = internal::getRaw( x );
@@ -10450,8 +10454,7 @@ namespace grb {
 
 					// prepare registers
 					for( size_t k = 0; k < AnyOp::blocksize; ++k, ++i ) {
-						mask[ k ] = already_dense_input_x ||
-							local_x.assigned( already_dense_input_y ? i : local_y.index( i ) );
+						mask[ k ] = already_dense_input_x || local_y.assigned(local_x.index( i ));
 					}
 
 					// rewind
@@ -10463,7 +10466,7 @@ namespace grb {
 							xx[ k ] = static_cast< typename AnyOp::D1 >(
 								a[ ( already_dense_input_x ? i : local_x.index( i ) ) + lower_bound ] );
 							yy[ k ] = static_cast< typename AnyOp::D2 >(
-								b[ ( already_dense_input_y ? i : local_y.index( i ) ) + lower_bound ] );
+								b[ ( already_dense_input_x ? i : local_x.index( i ) ) + lower_bound ] );
 						}
 					}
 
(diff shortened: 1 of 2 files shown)


GiovaGa:
Tests seem to pass without issue (we'll do full test suite after review).
This fix may depend on some assumptions about the number of nonzeros of the two vectors and the parameters supplied.
It may be good to add some assertions and/or comments about the assumptions made, but I am not sure what are the right things to check

anyzelman:
Hi @GiovaGa , I had a quick look -- assuming the fix is correct (haven't dived deep yet), isn't this line here also in error?

https://github.com/Algebraic-Programming/ALP/pull/398/files#diff-7936a0a79fe58f59fa9da39efc2a7c15f7ae9e49710647c69ba507e51cfa8e3dR10464

GiovaGa:
The line you mention surely looks wrong, because it is the same index as the one below. But it may not be strictly related to the other fix

GiovaGa:
I have now had a look again at this code, and I propose a new fix that is more convincing to me. I also fixed the line that you indicated. I will run tests to check everything works as expected

GiovaGa:
And in fact tests failed. I will look into it

GiovaGa:
After this fix tests pass on internal CI, except for `dot_debug_hyperdags` and `dot_ndebug_hyperdags`, I will check locally

GiovaGa:
Locally test pass. @anyzelman I think you can check that the change makes sense
You can find my rationale in a comment of the issue

anyzelman:
Note to self: ready for review

anyzelman:
Hi @GiovaGa -- could you check my latest commit? I think the issue was half-resolved, and should now be fully resolved. Can also discuss in person, perhaps easier

anyzelman (approved):

anyzelman:
All tests succeed (with LPF). CI running-- if pass, will merge. Concept release notes:

The implementation of the dot-product for the nonblocking backend could iterate over different nonzero indices in its input vectors. This bug did not materialise always, and did not materialise if the dot-product was called with the dense descriptor. This MR extends the unit test of the dot product to test more challenging sparse nonzero patterns, especially patterns that differ amongst the inputs, and add tests with two vectors with partial overlap as well as with zero overlap. The thus-extended unit test was able to trigger the pre-existing bug, which this MR also addresses. In addressing the bug, furthermore, issue #408 was uncovered.

Thanks to @GiovaGa for finding the bug, and for proposing an initial fix and unit test extension!

GiovaGa:
@anyzelman Smoke test for simulated annealing-Replica Exchange (#378) fails with nonblocking backend (see [Actions](https://github.com/Algebraic-Programming/ALP/actions/runs/20781509108) )
Running locally I see that the following assertion fails:
``` 
include/graphblas/nonblocking/blas1.hpp:10438: grb::RC grb::internal::sparse_dot_generic(typename AddMonoid::D3&, size_t, size_t, const Coords&, const Coords&, const grb::Vector<MaskType, grb::nonblocking, Coords>&, const grb::Vector<MaskType, grb::nonblocking, Coords>&, size_t, const AddMonoid&, const AnyOp&) [with unsigned int descr = 0; bool already_dense_input_x = false; bool already_dense_input_y = true; AddMonoid = grb::Monoid<grb::operators::add<float>, grb::identities::zero>; AnyOp = grb::operators::mul<float>; InputType1 = float; InputType2 = float; Coords = grb::internal::Coordinates<grb::nonblocking>; typename AddMonoid::D3 = float; size_t = long unsigned int]: Assertion `local_x.nonzeroes() <= local_y.nonzeroes()' failed.
```

PR #401, opened 2025-11-12
Added checks on output vector in mxv

### LINKED ISSUE #400: `grb::mxv<dense>` with sparse output returns `grb::SUCCESS`
When calling `grb::mxv()` with a dense descriptor but output vector not structurally dense, the return code is `grb::SUCCESS` instead of the expected `grb::ILLEGAL`.

For example:
```
	grb::Vector<float> x( 5 ), y( 5 );
	grb::Matrix<float> A( 5, 5 );
	
	// Build the vectors and the matrix ...
	
	auto plusTimes = grb::semirings::plusTimes< float >();

	grb::clear( x );
	// grb::set( x, 0 ); // without this line x is not initialized and thus not dense
	grb::set( y, 1 ); // without this line x is not initialized and thus not dense

	std::cerr << "grb::nnz(x) = " << grb::nnz(x) << std::endl; // this prints 0

	rc = grb::mxv< grb::descriptors::dense >(x, A, y, plusTimes);
	assert( rc == grb::SUCCESS );
```
Runs successfully even though it should throw an error.

This behavior has been observed with the following backends: `reference`, `reference_omp`, `nonblocking`, `bsp1d`,  `hyperdags`, `hybrid`

GiovaGa:
Resolves #400

### DIFF
--- a/include/graphblas/base/distribution.hpp
+++ b/include/graphblas/base/distribution.hpp
@@ -67,14 +67,11 @@ namespace grb {
 				 * The argument \a P must be strictly larger than one.
 				 */
 				static constexpr size_t global_index_to_process_id(
-// need to employ an ugly hack to make sure all of compilation without warnings
-// and doxygen work
-#ifdef __DOXYGEN__
 					const size_t i, const size_t n, const size_t P
-#else
-					const size_t, const size_t, const size_t
-#endif
 				) {
+					(void) i;
+					(void) n;
+					(void) P;
 					return 0;
 				}
 
@@ -133,15 +130,12 @@ namespace grb {
 				 * @returns The number of elements preceding \a s.
 				 */
 				static constexpr size_t local_offset(
-// need to employ an ugly hack to make sure all of compilation without warnings
-// and doxygen work
-#ifdef __DOXYGEN__
 					const size_t global_size,
 					const size_t s, const size_t P
-#else
-					const size_t, const size_t, const size_t
-#endif
 				) {
+					(void) global_size;
+					(void) P;
+					(void) s;
 					return 0;
 				}
 
@@ -163,6 +157,50 @@ namespace grb {
 					(void) global_size;
 					(void) P;
 					return 0;
+				}
+
+				/**
+				 * For a given global length, how many elements or rows shall be stored at
+				 * the given process \a s.
+				 *
+				 * @param[in] global The global size of the vector or of the matrix
+				 *                   dimension.
+				 * @param[in]   s    Request the local length at this process.
+				 * @param[in]   P    The global number of active user processes. Must be
+				 *                   larger than \a s.
+				 *
+				 * @returns The number of vector elements or matrix rows to store at the
+				 *          given process \a s, given the \a global size of the vector or
+				 *          matrix.
+				 *
+				 * This function completes in \f$ \Theta(1) \f$ time.
+				 */
+				static inline size_t global_length_to_local(
+					const size_t global,
+					const size_t s, const size_t P
+				) {
+					(void) s;
+					(void) P;
+					return global;
+				}
+
+				/**
+				 * For a given global length, the maximum number of elements, rows, or
+				 * columns that any process stores locally.
+				 *
+				 * @param[in] global The global size of the vector or of the matrix
+				 *                   dimension (row- or column-wise).
+				 * @param[in] P      The global number of active user processes.
+				 *
+				 * @returns The requested number of elements, rows, or columns.
+				 *
+				 * This function completes in \f$ \Theta(1) \f$ time.
+				 */
+				static inline size_t global_length_to_max_local(
+					const size_t global, const size_t P
+				) {
+					(void) P;
+					return global;
 				}
 
 		};
--- a/include/graphblas/bsp/collectives.hpp
+++ b/include/graphblas/bsp/collectives.hpp
@@ -291,7 +291,8 @@ namespace grb {
 				if( all ) {
 #ifdef _DEBUG_BSP_COLLECTIVES
 					std::cout << "\t\t collectives< BSP >::reduce_allreduce_generic, calls "
-						<< "lpf_allreduce with size " << sizeof(OPIOT) << std::endl;
+						<< "lpf_allreduce with size " << sizeof(OPIOT) << " and buffer content "
+						<< (*buffer) << std::endl;
 #endif
 					(void) root;
 					lpf_rc = lpf_allreduce(
@@ -334,7 +335,7 @@ namespace grb {
 				}
 
 				// done
-				return SUCCESS;
+				return rc;
 			}
 
 
@@ -356,6 +357,8 @@ namespace grb {
 			 *
 			 * This function may place an alloc of \f$ P\mathit{sizeof}(IOType) \f$ bytes
 			 * if the internal buffer was not sufficiently large.
+			 *
+			 * @tparam Operator A binary operator that must never generate no-ops.
 			 */
 			template<
 				Descriptor descr = descriptors::no_operation,
@@ -374,6 +377,8 @@ namespace grb {
 #endif
 
 				// static sanity checks
+				static_assert( !grb::internal::maybe_noop< Operator >::value,
+					"Operators passed to allreduce must never be able to generate no-ops" );
 				static_assert( !grb::is_object< IOType >::value,
 					"grb::collectives::allreduce cannot have another ALP object as its scalar "
 					"type!" );
@@ -445,6 +450,8 @@ namespace grb {
 					<< std::endl;
 #endif
 				// static sanity checks
+				static_assert( !grb::internal::maybe_noop< Monoid >::value, "Monoids "
+					"passed to allreduce must never be able to generate no-ops" );
 				static_assert( !grb::is_object< IOType >::value,
 					"grb::collectives::allreduce cannot have another ALP object as its scalar "
 					"type!" );
@@ -513,6 +520,8 @@ namespace grb {
 					<< inout << " and op = " << &op << std::endl;
 #endif
 				// static sanity checks
+				static_assert( !grb::internal::maybe_noop< Operator >::value,
+					"Operators passed to reduce must never be able to generate no-ops" );
 				static_assert( !grb::is_object< IOType >::value,
 					"grb::collectives::allreduce cannot have another ALP object as its scalar "
 					"type!" );
@@ -582,6 +591,8 @@ namespace grb {
 					<< &monoid << std::endl;
 #endif
 				// static sanity checks
+				static_assert( !grb::internal::maybe_noop< Monoid >::value, "Monoids passed to "
+					"reduce must never be able to generate no-ops" );
 				static_assert( !grb::is_object< IOType >::value,
 					"grb::collectives::allreduce cannot have another ALP object as its scalar "
 					"type!" );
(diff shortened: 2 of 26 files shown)


anyzelman:
The MR fixes it for reference and reference_omp, but I thought you mentioned all other backends also run into this issue @GiovaGa ?

GiovaGa:
You are definitely right. I have now fixed also for the nonblocking backend.
The other backends call the same internal function of the reference backend, so this should do it for them

anyzelman (dismissed):

anyzelman:
Running CI, running all unit & smoke tests with LPF. Looks ready to merge if these are both OK. Concept release notes:

Prior to this MR, calling `mxv` or `vxm` with a dense descriptor while the input/output vector was not dense, would not yield `ILLEGAL` for all supported backends. This MR fixes that, and also ensures that `ILLEGAL` is returned when one of the vector masks is non-empty (size larger than 0) and sparse. Furthermore, this MR adds a unit tests to guard against regressions.

Thanks to @GiovaGa for spotting the bug and providing the fixes for the `reference`, `reference_omp`, and `nonblocking` backends!

GiovaGa (commented):
Seems good to me.
My only observation is that it may make sense to add an assert in `exec_tests` to guarantee that indeed at least one vector is not dense, as going quickly through the test, this doesn't seem obvious (and the function `exec_tests` does not specify such precondition)

GiovaGa:
Right now, the test fails when running with 16 processes (and only in that case)

GiovaGa:
It seems that the test fails only if `n` is small enough. Using `n = 1000` the test runs correctly on my machine. This is probably a bug with the bsp1d backend

anyzelman:
Example of a failed run:

```
$ /scratch/workspace/alp-build/install/bin/grbrun -b bsp1d -np 16 tests/unit/illegal_spmv_debug_bsp1d 148
This is functional test tests/unit/illegal_spmv_debug_bsp1d
Info: grb::init (BSP1D) called using 16 user processes.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Error: unexpected error code in grb::setElement (BSP1D): Uninterpretable error code detected, please notify the developers.. Please submit a bug report.
Error: unexpected error code in grb::setElement (BSP1D): Uninterpretable error code detected, please notify the developers.. Please submit a bug report.
Test batch 5-8: initialisation FAILED
Test batch 5-8: initialisation FAILED
Info: grb::finalize (bsp1d) called.
Launching test FAILED
```

Any larger problem size no longer fails.

anyzelman:
The following similarly fails (P=11, n=102 -- all larger n are OK):

```
/scratch/workspace/alp-build/install/bin/grbrun -b bsp1d -np 11 tests/unit/illegal_spmv_debug_bsp1d 102
```

anyzelman:
More minimal one that fails:
```
void grb_program( const size_t &n, grb::RC &rc ) {
        grb::Vector< bool > out( n ) , out2( 2 * n );
        if( grb::setElement( out, true, 0 ) != grb::SUCCESS ) {
                std::cout << "FAILED\n";
                rc = grb::FAILED;
        } else {
                std::cout << "OK\n";
                rc = grb::SUCCESS;
        }
        std::cout << std::endl;
}
```
Failure here occurs "already" for P=7 and n=64 (this is the smallest P, n that causes failure). What is particularly interesting is that without declaring `out2` is needed while the 2x size is also "mandatory"-- otherwise there would be no failure. This points to a shared buffer corruption or a more general memory corruption problem.

anyzelman:
Workstation died but is now back up with a newer Linux. GCC 14.3.1, MPICH 4.1.2 has the same issue persist:

```
$ mpirun -np 11 tests/unit/illegal_spmv_debug_bsp1d 
[...]
		 collectives< BSP >::reduce_allreduce_generic, calls lpf_allreduce with size 4 and buffer content 0
		 before update: 0, update[ 0 ] = 0
		 after update: 0
		 before update: 0, update[ 1 ] = 64
		 after update: 1
		 before update: 1, update[ 2 ] = 0
		 after update: 1
		 before update: 1, update[ 3 ] = 64
		 after update: 1
		 before update: 1, update[ 4 ] = 0
		 after update: 1
		 before update: 1, update[ 5 ] = 8
		 after update: 1
		 before update: 1, update[ 6 ] = 0
		 after update: 1
		 before update: 1, update[ 7 ] = 0
		 after update: 1
		 before update: 1, update[ 8 ] = 0
		 after update: 1
		 before update: 1, update[ 9 ] = 0
		 after update: 1
		 writing back 1
illegal_spmv_debug_bsp1d: /home/yzelman/Documents/alp/tests/unit/illegal_spmv.cpp:85: void grb_program(const size_t&, grb::RC&): Assertion `rc == grb::SUCCESS' failed.
[...]
```

anyzelman:
I'm at this point pretty certain there's a skew in number of calls to lpf_sync at different processes that causes this bug. This occurs somewhere before io.hpp:590 (setElement, BSP1D). At the entry-point to this setElement there is no way one of the processes could exit early before others, and this is verified. Thus the error is earlier.

Before this primitive is called and the program crashes, the test allocates vectors-- one of size n, and another of size 2n. Shifting focus to those constructors instead.

anyzelman:
Found it: this test, for certain combinations of n and P, triggers a situation where the global buffer of some processes must be reallocated while for others the current size suffices. This also causes a mismatch in memory registrations, that by virtue of all of our particular LPF implementations happened to not crash out immediately. It also causes the observed skew in syncs that caused the much later and rather obscure crash documented above.

The error is one or the other of the following:
1. the Vector< BSP1D >::initialize function makes is the following -- the requested buffer size depends on `_local_n`, which for some processes, depending on the global size of the requested vector, may be 0; OR
2. the ensureBufferSize function should not implement early-exit on sufficient memory.

The fixes are the following:
1. call ensureBufferSize with the maximum _local_n across all processes (hence all processes will be guaranteed to make the same resize decision); OR
2. have ensureBufferSize first collectively check whether any process requires reallocation.

The problem is option 2 is expensive, especially given that ensureBufferSize is called on all non-trivial container constructors and is also called for some primitives in its RESIZE phase-- each such occurrance would induce an (extra) full synchronisation, even if not needed. Therefore fix no. 1 is preferred, and I will add that on top of this MR.

anyzelman:
Full test suite with LPF running, and code review underway. Concept release notes (WIP):

@GiovaGa noted that sometimes an `mxv` or `vxm` with a dense descriptor would execute even if one of the given vectors was sparse. This happened across various backends. This MR fixes that, and also ensures that `ILLEGAL` is returned when one of the vector masks is non-empty (size larger than 0) and sparse. Furthermore, this MR adds a unit tests to guard against regressions.

While resolving this issue, two notable bugs were encountered:
1. for the BSP1D backend, certain combinations of n1, n2, and P, allocating two vectors of size n1 and n2 after one another could cause a superstep skew between processes, which almost surely causes critical errors during any subsequent computation. This could only occur for relatively small n1, n2 when compared to P.
2. for reduce and allreduce operations, the use of monoids could lead to no-ops when preparing the payload-to-be-reduced for outgoing communication. This means that it could be possible uninitialised values were communicated and reduced. The implementation of (all)reduce was reviewed and modified to no longer have this possibility (by using identities to initialise otherwise possibly uninitialised values, and using fold/foldr when casts are required).

The former bug is unrelated to the issue this MR addresses, as it was traced to within the BSP1D vector constructor.  Given the critical nature of the error, however, and also given that the new unit test for this MR triggered this bug, a fix is applied within this MR. The latter bug was not triggered anywhere in this MR (or in any known code), however was found while diagnosing the former bug. Given the potential criticality --use and propagation of uninitialised values-- a fix is applied on top of this MR immediately.

While resolving this issue, the following additional bugs were found:
 - not all functions a Distribution instance should implement were specified in the base class, herewith fixed;
 - the `grb::Collectives< BSP1D >::{all,}reduce` primitive could return SUCCESS even on failure, herewith fixed;
 - some traces were sent to stderr instead of stdout, herewith fixed.

While resolving all of the above, the following features were added:
 - the Distribution class now specifies a function to infer the maximum local vector length across all processes in O(1) time;
 - the no-op type trait now also is defined for monoids, which helped debugging the issue;
 - the broadcast collective for BSP1D has been optimised, similar to (all)reduce, to not require on-the-fly memory slot registrations.

This is accompanied with improved debug tracing of in particular the BSP1D backend, as well as minor code style fixes throughout the files touched. Many thanks to Giovanni for finding the issue and providing an initial fix!

anyzelman:
Code review: actually the change from any_or to logical_or is not what we want. It seems better to either revise the reduce/allreduce code such that it can deal with no-ops correctly. This would also reduce the size of this MR significantly. That this is not wanted is also picked up on by the unit tests -- aborting those and addressing this problem first.

anyzelman (dismissed):

anyzelman (dismissed):

anyzelman:
Unit tests found some issues with the new allreduce implementation -- indeed it was incompatible with non-monoid (all)reduces while for monoids there was a missing identity application. Now fixed -- CIs are running again, plus full test suite + LPF. If OK, will merge

anyzelman (approved):

anyzelman:
One last test was caught tripping-- now all should pass

anyzelman:
(Btw, just recording here post-merge, smoke and unit tests with LPF were all OK on CentOS Stream 9 & 10. On CentOS 10, furthermore, all performance tests are also OK- a total number of 2440 tests ran.)

PR #406, opened 2025-12-04
Fixing creation of `grb::Vector<reference>` with bsp1d backend

### LINKED ISSUE #403: Creating a `grb::Vector< T, grb::reference >` when running with `bsp1d` backend gives runtime error
Consider the following:
```cpp
#include <graphblas.hpp>

void grb_program( const int &in, int &out ) {
	const size_t s = grb::spmd<>::pid();
	const size_t n = 10;

	std::cerr << "Hello0 from process " << s << std::endl;

	grb::Vector< int > x( n );
	grb::RC rc = grb::SUCCESS;

	grb::Vector< int, grb::reference > mask ( n );
	std::cerr << "Hello1 from process " << s << std::endl;
	assert( rc == grb::SUCCESS );
}

int main( int argc, char ** argv ) {
	int n = 42;
	int error_code = 100;

	grb::Launcher< grb::AUTOMATIC > launcher;
	assert( launcher.exec( &grb_program, n, error_code, true )
		== grb::SUCCESS );
	return error_code;
}
```

The code compiles without warnings or errors. When running this program terminates with signal 11 (segfault). The output is something like the following: 
```
Info: grb::init (BSP1D) called using 2 user processes.
Hello0 from process 0
Hello0 from process 1
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Info: process mask is all-one, we therefore assume a single user process is present on this node and thus shall use aligned mode for memory allocations that are potentially touched by multiple threads.
Hello1 from process 0
Info: grb::finalize (bsp1d) called.
	 process 0 is finalising

a.out:25405 terminated with signal 11 at PC=706556ac658e SP=7ffc2f206538.  Backtrace:
[...]
```
It is notable that if the template argument of launcher instead of `grb::AUTOMATIC` is `grb::FROM_MPI` (with the suitable mpi initializations), the same program runs fine.

GiovaGa:
Resolves #403 

The issue seems to be that reference_mapper was not correctly initialized. Its constructor was called only for process 0. 

@anyzelman should I also add a test for to check that there is no issue like this in the future?

### DIFF
--- a/src/graphblas/reference/init.cpp
+++ b/src/graphblas/reference/init.cpp
@@ -62,6 +62,7 @@ grb::RC grb::init< grb::reference >(
 	// set memory policy
 	numa_set_localalloc();
 #endif
+	grb::internal::reference_mapper.clear();
 	// done
 	return grb::SUCCESS;
 }


GiovaGa:
This fix is clearly not working (weirdly fails only a single test in smoke+unit).
But the problem indeed seems that the constructor of `reference_mapper` is called only once, but should be called once per process

PR #407, opened 2025-12-09
Provide put & get message passing primitives for the bsp1d backend

### LINKED ISSUE #405: Provide put & get message passing primitives for the bsp1d backend
Provide an interface for point-to-point communication between processes run by the bsp1d backend

GiovaGa:
Resolves #405

### DIFF
--- a/include/graphblas/base/rdma.hpp
+++ b/include/graphblas/base/rdma.hpp
@@ -51,13 +51,15 @@ namespace grb {
 	class rdma {
 		public:
 		template< typename T >
-		static grb::RC register_global( T &buf );
+		static inline grb::RC register_global( T &buf );
 
 		template< typename T >
-		static grb::RC register_global( grb::Vector< T, grb::reference > &buf );
+		static inline grb::RC register_global( grb::Vector< T, grb::reference > &buf );
+
+		static inline grb::RC localRegisterSize( const size_t size );
 
 		template< typename T >
-		static grb::RC get( const size_t src_pid, T &src, T &dst );
+		static inline grb::RC get( const size_t src_pid, T &src, T &dst );
 
 		template<
 			grb::Descriptor descr,
@@ -65,10 +67,10 @@ namespace grb {
 			typename T,
 			typename Coords
 			>
-		static grb::RC get( const size_t src_pid, const grb::Vector< T, backend, Coords > &src, grb::Vector< T, backend, Coords > &dst );
+		static inline grb::RC get( const size_t src_pid, const grb::Vector< T, backend, Coords > &src, grb::Vector< T, backend, Coords > &dst );
 
 		template< typename T >
-		static grb::RC put( const T &src, const size_t dst_pid, T &dst );
+		static inline grb::RC put( const T &src, const size_t dst_pid, T &dst );
 
 		template<
 			grb::Descriptor descr,
@@ -76,7 +78,7 @@ namespace grb {
 			typename T,
 			typename Coords
 			>
-		static grb::RC put( const grb::Vector< T, backend, Coords > &src, const size_t dst_pid, grb::Vector< T, backend, Coords > &dst );
+		static inline grb::RC put( const grb::Vector< T, backend, Coords > &src, const size_t dst_pid, grb::Vector< T, backend, Coords > &dst );
 	}; // end class ``rdma''
 
 } // namespace grb
--- a/include/graphblas/base/rdma.hpp
+++ b/include/graphblas/base/rdma.hpp
@@ -51,7 +51,7 @@ namespace grb {
 	class rdma {
 		public:
 		template< typename T >
-		static grb::RC register_global( T &buf);
+		static grb::RC register_global( T &buf );
 
 		template< typename T >
 		static grb::RC register_global( grb::Vector< T, grb::reference > &buf );
@@ -60,8 +60,8 @@ namespace grb {
 		static grb::RC get( const size_t src_pid, T &src, T &dst );
 
 		template<
-			grb::Descriptor descr = descriptors::no_operation,
-			grb::Backend backend = grb::reference,
+			grb::Descriptor descr,
+			grb::Backend backend,
 			typename T,
 			typename Coords
 			>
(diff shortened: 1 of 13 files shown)


include/graphblas/base/rdma.hpp
```diff
@@ -0,0 +1,85 @@
...
+	 *
+	 * All backends must implement this interface, including backends that do not
+	 * support multiple user processes. The interface herein defined hence ensures
+	 * to allow for trivial implementations for single user process backends.
+	 */
+	template< Backend implementation >
+	class rdma {
+		public:
+		template< typename T >
+		static grb::RC register_global( T &buf);
```
anyzelman: missing space after `buf`

include/graphblas/base/rdma.hpp
```diff
@@ -0,0 +1,85 @@
...
+		template< typename T >
+		static grb::RC put( const T &src, const size_t dst_pid, T &dst );
+
+		template<
+			grb::Descriptor descr = descriptors::no_operation,
+			grb::Backend backend = grb::reference,
+			typename T,
+			typename Coords
+			>
+		static grb::RC put( const grb::Vector< T, backend, Coords > &src, const size_t dst_pid, grb::Vector< T, backend, Coords > &dst);
```
anyzelman: same here (missing space)

include/graphblas/base/rdma.hpp
```diff
@@ -0,0 +1,85 @@
...
+			typename Coords
+			>
+		static grb::RC get( const size_t src_pid, const grb::Vector< T, backend, Coords > &src, grb::Vector< T, backend, Coords > &dst );
+
+		template< typename T >
+		static grb::RC put( const T &src, const size_t dst_pid, T &dst );
+
+		template<
+			grb::Descriptor descr = descriptors::no_operation,
+			grb::Backend backend = grb::reference,
```
anyzelman: here it is better to not use a default template arg (we should let C++ infer it from the argument)

include/graphblas/base/rdma.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+		template< typename T >
+		static grb::RC register_global( grb::Vector< T, grb::reference > &buf );
+
+		template< typename T >
+		static grb::RC get( const size_t src_pid, T &src, T &dst );
+
+		template<
+			grb::Descriptor descr = descriptors::no_operation,
+			grb::Backend backend = grb::reference,
```
anyzelman: same here (no default template argument)

include/graphblas/bsp/rdma.hpp
```diff
@@ -0,0 +1,309 @@
...
+			 */
+			template< typename T >
+			static grb::RC register_global( const T* buf, const size_t size ) {
+				grb::internal::BSP1D_Data & data = grb::internal::grb_BSP1D.load();
+				lpf_err_t lpf_rc = LPF_SUCCESS;
+				lpf_memslot_t memslot = LPF_INVALID_MEMSLOT;
+				const void* buf_void = reinterpret_cast< const void* >( buf );
+
+				data.ensureMemslotAvailable( 1 ); 
+				data.signalMemslotTaken();
```
anyzelman: this one should come after line 62 (and only if that call returned SUCCESS probably?)

include/graphblas/bsp/rdma.hpp
```diff
@@ -0,0 +1,309 @@
...
+			template< typename T >
+			static grb::RC deregister( const T &buf ) {
+#ifdef _DEBUG
+				std::cout << "deregister: memslot " << memslot << std::endl;
+#endif
+				grb::internal::BSP1D_Data & data = grb::internal::grb_BSP1D.load();
+				lpf_err_t lpf_rc = LPF_SUCCESS;
+				lpf_memslot_t memslot = LPF_INVALID_MEMSLOT;
+
+
```
anyzelman: superfluous newline

include/graphblas/bsp/rdma.hpp
```diff
@@ -0,0 +1,309 @@
...
+				}
+
+				// check trivial dispatch
+				if( size == 0 ) {
+					return grb::SUCCESS;
+				}
+
+				// data.ensureMemslotAvailable( 1 ); // this function calls lpf_sync
+				{
+					const auto it = data.registered_slots.find( reinterpret_cast< const void* >( dst ) );
```
anyzelman: it may be more robust to introduce a grb::Scalar type ...but maybe let's worry about that later

include/graphblas/reference/rdma.hpp
```diff
@@ -0,0 +1,94 @@
...
+		}
+
+		template<
+			grb::Descriptor descr = descriptors::no_operation,
+			grb::Backend backend = grb::reference,
+			typename T,
+			typename Coords
+			>
+		static grb::RC put( const grb::Vector< T, backend, Coords > &src, const size_t dst_pid, grb::Vector< T, backend, Coords > &dst) {
+			return grb::SUCCESS;
```
anyzelman: I think we should use grb::set( dst, src ); ? (also elsewhere)

anyzelman (changes requested):
Intermediary review request to double-check code structure -- and all looks good from that perspective to me=)) Also some minor comments I saw while going through it quickly just now

include/graphblas/reference/rdma.hpp
GiovaGa: Yes, definitely (I didn't bother actually implementing these functions yet :sweat_smile: )

include/graphblas/bsp/rdma.hpp
GiovaGa: I agree for after line 62, but I am not sure about the SUCCESS check. There are many kinds of possible errors that could happen, so I think it's safer to take a memslot in any case

include/graphblas/bsp/rdma.hpp
GiovaGa: I'll leave this one open for the day I will actually implement this

GiovaGa:
Currently I register locally objects needed for put/get and then forget about them.
In theory we could deregister them after a subsequent sync, so @anyzelman do you think it is better to:
- just **keep them**, as they may be needed (pro: no need to modify sync function, con: more memory footprint)
- **deregister after every sync** (pro: less memory footprint, con: we may have many registration deregistration on the same memory, so it might be suboptimal)
- We might also implement some kind of caching strategy for this, but that seems definitely overkill

In both cases I will have to implement something to keep track of them.

It might also happen that we a locally registered address is registered globally, and in that case I guess we should deregister locally before registering globally

